PR #27, opened 2021-03-16
Add global names to Dallas sensors to avoid ugly <unknown> and other …

FauthD:
…issues in HA

Hi Proddy, 
you have asked to resend the PR from a fresh fork (cloned new ESP32 repo).

Many regards, DIeter Fauth

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ pio_local.ini
 # OS specific
 .DS_Store
 *Thumbs.db
+.directory
 
 # project specfic
 /scripts/stackdmp.txt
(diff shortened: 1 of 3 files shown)


FauthD:
Hi Proddy, 
looks like github collects further commits/pushes after I send the PR.

The "Add stuff to gitignore, add vscode workspace" was planned as a separate PR.

vscode workspace makes it easier to start fresh.

.gitignore now has two more files that are created by npm
+ .directory which is used by KDE for colored folders.

What I see is that "/interface/package-lock.json" is modified by npm. So after running it, the local copy is dirty (changed).
That kills automated version string creation (which I did not see in your project, but I am used to that since +20 years).
Not sure what can be done against it. Just wanted to mention it.

Dieter

proddy:
I've updated the gitignore. The package stuff has to stay in and there's a GitHub bot that will keep that up to date.

PR #40, opened 2021-03-30
Move texts and some other changes

MichaelDvP:
I see that you are son commits ahead, but i think i shoud merge.
topics are now all lower case, that's breaking change. Possible needs some fine tuning, but all texts are in local now.


proddy:
nice. I'll review and manually merge. I just did a 3.0.1 build to keep a snapshot since this is quite a major change

proddy (approved):

proddy (approved):

PR #44, opened 2021-04-10
mqtt-HA-config dynamically

MichaelDvP:
Please check if this works with HA.

### DIFF
--- a/src/emsdevice.cpp
+++ b/src/emsdevice.cpp
@@ -455,7 +455,7 @@ void EMSdevice::register_device_value(uint8_t                             tag,
         };
     }
 
-    devicevalues_.emplace_back(device_type_, tag, value_p, type, options, options_size, short_name, full_name, uom, has_cmd);
+    devicevalues_.emplace_back(device_type_, tag, value_p, type, options, options_size, short_name, full_name, uom, 0, has_cmd);
 }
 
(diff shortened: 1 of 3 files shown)


src/emsdevice.cpp
```diff
@@ -753,20 +754,24 @@ bool EMSdevice::generate_values_json(JsonObject & root, const uint8_t tag_filter
...
+    for (auto & dv : devicevalues_) {
+        if (dv.ha == DeviceValueHA::HA_VALUE) {
+            Mqtt::publish_mqtt_ha_sensor(dv.type, dv.tag, dv.full_name, device_type_, dv.short_name, dv.uom);
+            dv.ha |= DeviceValueHA::HA_DONE;
+        }
     }
-
-    bool ok = publish_ha_config();
-    ha_config_done(ok); // see if it worked
+    // bool ok = publish_ha_config();
```
proddy: this command is still needed for Home Assistant to build the main device (then all the sensors attach to it). When uncommented it works great!

Also maybe look at ha_config_done() as it really doesn't do anything. The idea was if you change the Home Assistant switch on the Web UI it would re-send all the Discovery HA's again.

Why don't you install Home Assistant yourself? It's really nice software. B-)

proddy (changes requested):

src/emsdevice.cpp
MichaelDvP: Yes i've seen it and commited a fix a few minutes ago.

PR #49, opened 2021-04-24
add min/max to values, boiler flags, ww-prefix, ha-prefix to mqtt

MichaelDvP:
When checking the missing commands i found some things:
- there are a lot of "homeassistent/.." topics, i moved this as prefix to mqtt and add/strip it here
- we have a lot of "warm water" also in other devices like mixer, thermostat, solar, I renamed `BOILER_DATA_WW` to `DEVICE_DATA_WW` and used in in all devices
- I added flags to boiler, but only for a few that i know
- add min/max optional to value registration, it's around 2-3k heap, i think the esp32 can manage it. For now it's only thermostat temp to 5-29°C, but we can add one by one if we know the range. If not set the info gives the range of storagefield.

It's working for me, check if you like it. It's a lot of changes for one commit, but one gives another...

### DIFF
--- a/src/dallassensor.cpp
+++ b/src/dallassensor.cpp
@@ -391,9 +391,9 @@ void DallasSensor::publish_values(const bool force) {
                     // use '_' as HA doesn't like '-' in the topic name
                     std::string topicname = sensor.to_string();
                     std::replace(topicname.begin(), topicname.end(), '-', '_');
-                    snprintf_P(topic, sizeof(topic), PSTR("homeassistant/sensor/%s/dallas_sensor%s/config"), Mqtt::base().c_str(), topicname.c_str());
+                    snprintf_P(topic, sizeof(topic), PSTR("sensor/%s/dallas_sensor%s/config"), Mqtt::base().c_str(), topicname.c_str());
                 } else {
-                    snprintf_P(topic, sizeof(topic), PSTR("homeassistant/sensor/%s/dallas_sensor%d/config"), Mqtt::base().c_str(), sensor_no);
+                    snprintf_P(topic, sizeof(topic), PSTR("sensor/%s/dallas_sensor%d/config"), Mqtt::base().c_str(), sensor_no);
                 }
                 Mqtt::publish_ha(topic, config.as<JsonObject>());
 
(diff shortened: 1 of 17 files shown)


proddy:
These are wonderful additions. I did a quick test on my system, a few comments:
 - with my RC20 I'm seeing "temporary set temperature" and the homeassistant values shown both in console and web. They should be hidden. I think I did this before by keeping the name (long name) empty

Screenshot 2021-04-24 144129

- Should we rename the "temporary set temperature" to "temporary" instead of "temp", as not to confuse it with the general word temperature?
`MAKE_PSTR_LIST(temp, F("temp"), F("temporary set temperature"))`

MichaelDvP:
> I think I did this before by keeping the name (long name) empty

Yes it's the same now, only longname is the second in LIST and has to be empty. We have to change in locale_EN.h: 
```
// extra commands, no long name, does not show on web/console
MAKE_PSTR_LIST(switchtime, F("switchtime"))
MAKE_PSTR_LIST(temp, F("temp"))
MAKE_PSTR_LIST(hatemp, F("hatemp"))
MAKE_PSTR_LIST(hamode, F("hamode"))
```
I was a bit in hurry and typed something in without deeper thinking.

> Should we rename the "temporary set temperature" 

I think we should also skip this longname and do not publish. I like the shortname `temp` because the name seems unspecific and can mean temperature or temporary, and it **is** unspecific, because it sets the temporary value in automode, but the daytemp in daymode (which is not temporary), but always have immediate effect changing the actual temperature.

proddy:
reviewed, test and it works fantastic. thanks.

proddy:
> I think we should also skip this longname and do not publish. I like the shortname temp because the name seems unspecific and can mean temperature or temporary, and it is unspecific, because it sets the temporary value in automode, but the daytemp in daymode (which is not temporary), but always have immediate effect changing the actual temperature.

@MichaelDvP  I still think having seltemp and temp is confusing. This is what I'm seeing now

in Console:
```sh
ems-esp:# call
Available commands per device: 
 system: pin send publish fetch info settings test 
 boiler: wwtapactivated reset selflowtemp heatingactivated heatingtemp pumpmodmax pumpmodmin pumpdelay burnminperiod burnminpower burnmaxpower boilhyston boilhystoff maintenance wwsettemp wwcomfort wwflowtempoffset wwmaxpower wwcircpump wwdisinfectiontemp wwcircmode wwcirc wwactivated wwonetime info 
 thermostat: datetime info seltemp temp mode heattemp ecotemp nofrosttemp 
ems-esp:#
```
and in MQTT:
```json
thermostat_data = 
{
  "id": 77,
  "errorcode": "(0)",
  "datetime": "13:30:47 26/04/2021",
  "hc1": {
    "seltemp": 15,
    "temp": 15,
    "currtemp": 19,
    "hatemp": 19,
    "hamode": "auto",
    "mode": "auto"
  }
}
```

Both `seltemp` and `temp` which have the same values and do the same thing. It may be confusing to some people.

MichaelDvP:
Oh yes, we need to suppress the command only values which have a command but no full_name. I'll fix it.

proddy:
thanks, looks good now.

PR #102, opened 2021-08-28
Added pool data to telegrams 0x494 & 0x495

Sunbuzz:
I have an IVT Geo 5xx, it's a Bosch Compress 7001i LW branched heatpump.
I have a MP100 connected on the EMS bus, which is a pool mixer, and when connected adds some pool options to the boiler. The mixer has a device_id of 0x11, and sends out telegrams of type 0x25C with information about pool temperature and status of the mixer. I am in progress of adding these as well as be able to set the desired pool temperature. But as a start I added the pool data that are coming from the boiler.


proddy:
nice addition, thanks

PR #109, opened 2021-09-12
add RC300 second summermode telegram

MichaelDvP:
I think this should work if for other users the telegrams are empty.


proddy:
looks ok. merge?

MichaelDvP:
Yes, please merge, and be aware of possible complains if the RC310 with summerSetMode in 0x2AF also have a non-empty 0x471.
The thermostats from tp1de and other known users have v18.xx, the thermostat with 0x471 is v13.xx. If the `fetch`does not work we can try this.

Issue #127, opened 2021-09-23
Add option for auxiliary port


Sunbuzz:
The main reason I got into this project, was to be able to control my pool circulation pump. I needed a signal from the heatpump when the pool was actually heating, so that I could turn the circulation pump off, the pool circulation pump is about 1kW and I thought it was unnecessary to keep it on 24h a day when the heatpump only heats about 2h a day.

I have this system all set up on a custom forked branch, but I was thinking on making it available to everybody.
The Gateway E32 (which I use) have an auxiliary GPIO port (32) just above the ESP32 chip, and the Gateway S32 has one too (GPIO 22 I think).

I have made it so that you can edit the aux GPIO in the interface, and choose what function it should cater to, thinking others may use this port for something else in the future. I have attached a "pool circulation pump" option with functionality. 

If you think this is a good Idea -
I have made a clean branch based on "dev" and I am wondering if you want me to create a pull request.

<img width="1268" alt="Skärmavbild 2021-09-23 kl  11 28 13" src="https://user-images.githubusercontent.com/29543733/134484489-541b8c40-d925-4f76-a3e7-df563ebcc8bd.png">
<img width="1271" alt="Skärmavbild 2021-09-23 kl  11 28 29" src="https://user-images.githubusercontent.com/29543733/134484615-18711507-7bba-43ae-bd0d-de4a29487fd8.png">

proddy:
EMS-ESP is open-source so we welcome any new additions. 

The idea of exposing one of the GPIOs in the WebUI is a good idea (we need to vote on what to call it though!) and also having a list of configurable actions like your 'pool circulation pump'.  There is already a `pin` command in EMS-ESP to set any of the ESP's GPIO's to HIGH or LOW. I think Michael uses this to switch external pumps on&off so that could be an additional action to add.

The pump stop delay - is that just a time for switching between on and off? 

I would like to see some of the code - is it at https://github.com/Sunbuzz/EMS-ESP32/tree/PumpSwitch ?

@MichaelDvP what do you think?

Sunbuzz:
The pump delay is how long the circulation should stay on after the heatpump stopped heating the pool.

No, PumpSwitch was a first draft, I'll push the branch now, it's called "ft_aux"

MichaelDvP:
>  I think Michael uses this to switch external pumps on&off so that could be an additional action to add.

[...]

PR #131, opened 2021-09-26
Add option for auxiliary port (#127)

Sunbuzz:
Adding support for the auxiliary GPIO found in Gateway E32 and Gateway S32, and an option to choose the GPIO on a custom board.

It includes a custom function "Pool circulation pump" for controlling a circulation pump when the pool is actually heating, with a value how long the pump should stay on after the heating stopped (in seconds).


proddy:
please add to the ft_aux branch so we can work collectively on this

PR #181, opened 2021-11-06
Show mode (off / heat) in Home Assistant for Moduline 200

kpschaper:
Shows mode in Home Assistant based on manually selected mode on thermostat (day or off)
Unfortunately, it does not allow you to set the mode remotely through RC10Set
(RC10Set shows up in log for temp change but not for mode change) 

Below some tests using read:

**read thermostat b1**

day (temp: 20deg, night temp 19deg, off temp 7deg, 7h)
04 28 00 C2 00 00 00 C2

**set** night (temp: 20deg, night temp 19deg, off temp 7deg, 7h)
02 26 00 C2 01 A3 00 C2

night (temp: 20deg, **set** night temp 15deg, off temp 7deg, 7h)
02 1E 00 C5 01 A3 00 C6

**set** off (temp: 20deg, night temp 19deg, off temp 7deg, 7h)
01 0E 00 C2 00 00 00 C2

**set** on (curr temp: 16deg, set temp 19deg)
04 23 00 BA 00 00 00 BA

**read thermostat b0**

off (temp: 20deg, night temp 19deg, return day 7h)
00 FF 00 26 28 07 01

on (**set** temp: 16deg, night temp 19deg, return day 7h)
00 FF 00 26 20 07 01

night (temp: 16deg, **set** night temp 14deg, return day 7h)
00 FF 00 1C 20 07 01

night (temp: 16deg, night temp 14deg, **set** return day 8h)
00 FF 00 1C 20 08 01

### DIFF
--- a/src/devices/thermostat.cpp
+++ b/src/devices/thermostat.cpp
@@ -493,7 +493,15 @@ uint8_t Thermostat::HeatingCircuit::get_mode() const {
         return HeatingCircuit::Mode::UNKNOWN;
     }
 
-    if (model == EMSdevice::EMS_DEVICE_FLAG_RC20) {
+    if (model == EMSdevice::EMS_DEVICE_FLAG_RC10) {
+        if (mode == 1) {
+            return HeatingCircuit::Mode::OFF;
+        } else if (mode == 2) {
+            return HeatingCircuit::Mode::NIGHT;
+        } else if (mode == 4) {
+            return HeatingCircuit::Mode::ON;
+        }
+    } else if (model == EMSdevice::EMS_DEVICE_FLAG_RC20) {
         if (mode == 0) {
             return HeatingCircuit::Mode::OFF;
         } else if (mode == 1) {
@@ -703,7 +711,8 @@ void Thermostat::process_RC10Monitor(std::shared_ptr<const Telegram> telegram) {
     if (hc == nullptr) {
         return;
     }
-
+    has_update(telegram->read_value(hc->mode, 0));
+    hc->hamode = hc->mode == 4 ? 1 : 0;                            // set special HA mode: off, on, auto
     has_update(telegram->read_value(hc->setpoint_roomTemp, 1, 1)); // is * 2, force as single byte
     has_update(telegram->read_value(hc->curr_roomTemp, 2));        // is * 10
 }


MichaelDvP:
Nice to see the whole B0 telegram and different states of B1. Maybe we can extract a bit more.
B1:
- first byte 0 (mode), maybe better to store it in modetype as it is the reported state, but since there is no mode-setting it does not matter. hamode can be set independent of the naming. As there is a off-temperature of 7 degrees i suggest to use mode(types) : "nofrost, night, day" for emsesp publish (hamode has fixed names)
- 1,2 is the actual temperature, it is repeated in pos 6/7, i think there is a calibration stored in B0 (pos 0 or 2) and B1 shows raw and calibrated temp.
- 4/5 shows a value in night mode, that could be the remaining time in minutes for return to day function. Does the value decrease with time?
B0:
- you have clearly identified 3-nighttemp, 4-daytemp, 5-return time. We should add publish and setting functions for these values. (Now we have seltemp writing to daytemp, we can change that seltemp writes to the actual mode).
- The FF surely means something is activated, maybe warm water?
- When changing the mode by key, is the B0 (maybe partly) published by the thermostat? Maybe the modechange can be triggert by writing something?

kpschaper:
> When changing the mode by key, is the B0 (maybe partly) published by the thermostat? Maybe the modechange can be triggert by writing something?

No, the thermostat does not publish the B0 when the mode is changed.
(In contrary to setting temperature where the thermostat does publish a shorter B0)

I think most of the above would only make sense if the mode could be set remotely.
Is there any commanline functionality, similar to read, to test writing to different offsets in B0?

proddy:
shall we make an enhancement issue for this?

MichaelDvP:
> I think most of the above would only make sense if the mode could be set remotely.

Yes, it would be better, but changing nighttemp, night(reduce)-hours and read when the thermmostat switches to daymode could be usefull. 
> Is there any commanline functionality, similar to read, to test writing to different offsets in B0?

for writing we only have the `call system send "0B 17 B0 xx yy"` command, (xx-offset, yy-data), for writing we always need offset and data, so an extra write command does not shorten the command much. 

@proddy 
> shall we make an enhancement issue for this?

Yes, imo merge this PR and we add other parameters later, if there is somthing usefull.

proddy:
Ok, I'll merge and then if @kpschaper you would be so kind to create a new GH issue where we can continue the best way fwd.

PR #185, opened 2021-11-07
Fix shower timer; doing_cold_shot is never being reset

kpschaper:
@proddy
I know that the shower timer/alert functionality is still experimental.
Would it be an idea to restructure the codebase to allow (native)unit testing?
[PlatformIO Unit Testing Engine does not build source code from src_dir folder by default](https://docs.platformio.org/en/latest/plus/unit-testing.html#test-types)

It would not be environmently friendly to take a 7min+ shower every time we want to test this :sweat_smile:

### DIFF
--- a/src/shower.cpp
+++ b/src/shower.cpp
@@ -60,7 +60,7 @@ void Shower::loop() {
                     LOG_DEBUG(F("[Shower] hot water still running, starting shower timer"));
                 }
                 // check if the shower has been on too long
-                else if ((((time_now - timer_start_) > SHOWER_MAX_DURATION) && !doing_cold_shot_) && shower_alert_) {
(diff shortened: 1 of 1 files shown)


proddy:
thanks, not sure why I forget to complete that piece of code. I've been using the shower timer and alert and it works fine - only switching the boiler to the Factory Test is a bit risky but it seems to work on my boiler.

proddy:
and on Unit Testing, I never had much luck with pio's way of doing it so built my own test suite which I call standalone mode (see https://emsesp.github.io/docs/#/Coding). It's quite hard to build Unit Tests due to the real-time nature of the application and the Web. Now I do 99% of the development offline using the standalone version with dependency injection (the `test` command from the console). So `make run` for the EMS-ESP core and `npm run standalone` for the WebUI. But happy to explore other methods if you have any...!

Issue #183, opened 2021-11-07
Improve moduline 200 functionality


kpschaper:
New issue to improve moduline 200. Ref: #181 

**B1:** 

- first byte 0 (mode)
> Maybe better to store it in modetype as it is the reported state, but since there is no mode-setting it does not matter. hamode can be set independent of the naming. As there is a off-temperature of 7 degrees i suggest to use mode(types) : "nofrost, night, day" for emsesp publish (hamode has fixed names)
- 1,2 is the actual temperature
- 4,5 remaining time in minutes for return to day function -> value decreases with time

> For example return to day time set to 8 hours:
day 04 27 00 C9 00 00 00 C9
night 02 1C 00 C9 01 E0 00 C9
000+00:05:01.454 N 9: [emsesp] Thermostat(0x17) -> Me(0x0B), RC10Monitor(0xB1), data: 02 1C 00 C9 01 E0 00 C9
000+00:07:01.526 N 12: [emsesp] Thermostat(0x17) -> Me(0x0B), RC10Monitor(0xB1), data: 02 1C 00 C9 01 DE 00 C9

**B0:**
- call system send "0B 17 B0 00 01" -> calibrate room temperature 0.1 = 01 , -0.1 = FF and -0,2 = FE 
(steps 0,1 with range -5 - 5 deg according to manual)
- call system send "0B 17 B0 01 00" -> set key backlight FF = on, 00 = off
- call system send "0B 17 B0 02 00" -> set hot water preparation 0 = on, 1 = off, 2 = auto 
- call system send "0B 17 B0 03 00" -> set night temperature
- call system send "0B 17 B0 04 00" -> set day temperature
- call system send "0B 17 B0 05 08" -> set night timer start to 8 hours (value between 1 and 99 according to manual)
- call system send "0B 17 B0 06 00" -> set building type valid options = 0,1,2 
(valid options +1 map to options 1,2,3 according to manual)

**Unfortunately, it does not seem to be possible to set the mode remotely through RC10Set**

MichaelDvP:
Oh, you have been busy, great  👍  Some ideas to test:
Setting the timer only changes the value, or does it start night-mode on writing?
Is B1 writable, the modes looks like a bit field, maybe the high nibble is a mask (like 0x35 telegram on other thermostats). 
Try to write `call system send "0B 17 B1 00 22"` for nightmode, 44 for day.

[...]

Issue #182, opened 2021-11-06
RC35 and controlling two heat circuits with custom time programs.  


greg-de:
Hi all,

I have a Buderus Logano oil boiler with RC35 connected to it. It controls two heat circuits. 
According to the documentation this thermostat provides two user specific programs for every of each four heat circuits called Eigen 1 and Eigen 2 (German names). In my current configuration I'm using:
- for heat circuit 1 Eigen 1 program  and 
- for heat circuit 2 Eigen 2 program. 

Eigen 1 program shall be controlled by 0x3F telegram according to https://emswiki.thefischer.net/doku.php?id=wiki:ems:telegramme, Eigen 2 shall be controlled by 0x42. 

Based on the code of thermostat.cpp (line 66) for RC35 only the first telegram (0x3F) is supported. As these both telegrams are used for example for switching on/off party or pause function I think that the current implementation of EMS-ESP app can only control the Eigen 1 program.  

Could you please check the implementation and if necessary extend it to support both programs. 

BR,
Greg

proddy:
sure, @MichaelDvP this is your expertise...

MichaelDvP:
The "Eigen 1" contains the used program, holiday, pause, party functions, "Eigen 2" only the switchprogram, Thats why we read only the first telegram. But for[ setting the switchtime](https://github.com/emsesp/EMS-ESP32/blob/91b7fd59d11b9d5a117c356dca4787c59d8dccb2/src/devices/thermostat.cpp#L1981) we should add an optional program-no and support writing a entry of second telegram. 

But i think each hc has its own 2 programs, You can use hc2 ownprog 1 with different values as hc1 ownprog 1.
Can you do in terminal `read 10 3f`, `read 10 42`, `read 10 49`, `read 10 4c` and post the result.

greg-de:
thanks for the info, I will send you the results on Tuesday/Wednesday as I'm currently on a business trip.

greg-de:
Hi MichaelDvP,

here are the results:

┌──────────────────────────────────────┐
│ EMS-ESP version 3.3.0b3              │
│ https://github.com/emsesp/EMS-ESP32  │
│                                      │
│ type help to show available commands │
└──────────────────────────────────────┘

ems-esp:$ read 10 3f
[...]

PR #188, opened 2021-11-11
Mode update for #183

MichaelDvP:
This is not the thermostat change for RC35, only the update for #183. I wanted to push, but is seems i'm blocked out.
There is also a small check for double log messages in weblog, i've tested yesterday.
The thermostat additions for #182 i will first check with my own build.


src/devices/thermostat.cpp
```diff
@@ -1326,16 +1343,28 @@ bool Thermostat::set_building(const char * value, const int8_t id) {
         return false;
     }
     LOG_INFO(F("Setting building to %s"), value);
-    if (model() == EMS_DEVICE_FLAG_RC10) {
-        write_command(0xB0, 6, bd, 0xB0);
-    } else if ((model() == EMS_DEVICE_FLAG_RC300) || (model() == EMS_DEVICE_FLAG_RC100)) {
+    if ((model() == EMS_DEVICE_FLAG_RC300) || (model() == EMS_DEVICE_FLAG_RC100)) {
         write_command(0x240, 9, bd + 1, 0x240);
```
proddy, line 1347: would be nice to have this 0x240 defined in thermostat.h as constexpr

interface/src/system/LogEventController.tsx
```diff
@@ -157,7 +159,10 @@ class LogEventController extends Component<
     const rawData = event.data;
     if (typeof rawData === 'string' || rawData instanceof String) {
       const event = JSON.parse(rawData as string) as LogEvent;
-      this.setState((state) => ({ events: [...state.events, event] }));
+      if (event.i > this.state.last_id) {
```
proddy, line 162: is this to prevent duplicates? I think you mentioned seeing that occur earlier?

proddy:
Merged. From now it's just PRs and we get to review and comment on each other's code. Please also update the changelog with these additions.

PR #193, opened 2021-11-14
Mode update for #183

kpschaper:
Allow mode "off" to be set from home assistant + remove non available "auto" mode from home assistant

Changed nofrost to off because:
- Other thermostats that use nofrost allow the temprature to be set, the moduline 200 does not (fixed at 7 deg).
- Home assistant does not have a nofrost mode e.g. off is used
- Thermostat manual refers to the button as: "off-function"

image

### DIFF
--- a/CHANGELOG_LATEST.md
+++ b/CHANGELOG_LATEST.md
@@ -43,6 +43,7 @@
 - Unit of Measure 'times' added to MQTT Fails, Rx fails, Rx received, Tx fails, Tx reads & Tx writes
 - Improved API. Restful HTTP API works in the same way as MQTT calls
 - Removed settings for MQTT subscribe format [#173](https://github.com/emsesp/EMS-ESP32/issues/173)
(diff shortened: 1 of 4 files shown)


proddy (commented):
Looks good. Could you mention the change in CHANGELOG_LATEST,md ?

proddy:
So the RC10 doesn't have an Auto mode? Does it not have a schedule with heating points? My RC20 does and in HA it shows as

image

kpschaper:
> Looks good. Could you mention the change in CHANGELOG_LATEST,md ?
Yes, I will add a commit for that.

> So the RC10 doesn't have an Auto mode? Does it not have a schedule with heating points? My RC20 does and in HA it shows as

No it does not, it has three modes: day, night and off
Where night allows you to change to a lower temprature for x amount of time.
But still you need to push the night button manually (without a Gateway S32).

This was the main reason to get a Gateway S32; schedule temperature trough Home Assistant

proddy (approved):

PR #194, opened 2021-11-14
bugfix revert auto mode -> the mode is hardcoded for the thermostat_ha_cmd #183

kpschaper:
@proddy 

Not sure if this is the best solution, always pass AUTO mode and determine inside the function if the mode is actual set to AUTO?

https://github.com/emsesp/EMS-ESP32/blob/5893487d4aed8264ec1a3fbd3da5c9b52b746fc0/src/devices/thermostat.cpp#L483

### DIFF
--- a/src/devices/thermostat.cpp
+++ b/src/devices/thermostat.cpp
@@ -2228,8 +2228,14 @@ bool Thermostat::set_temperature(const float temperature, const uint8_t mode, co
         case HeatingCircuit::Mode::DAY:
             offset = 4;
             break;
+        case HeatingCircuit::Mode::AUTO:
+            if (hc->get_mode() == HeatingCircuit::Mode::NIGHT) {
+                offset = 3;
+            } else {
+                offset = 4;
+            }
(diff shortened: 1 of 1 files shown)


proddy:
that is what it does now. The thermostat component in HA is very simple and stupid. It doesn't handle all the modes you can have in thermostats (like temp, summer, comfort, eco, holiday, design etc etc) so I just default it to AUTO when a temperature is changed in HA

kpschaper:
Ah I get it now, it's just the most convenient way to map from HA modes to thermostat specific modes

proddy:
exactly. Probably the best method is to use whatever the current Thermostat mode is and adjust that setpoint temperature. There's probably a good reason why we don't do it that way. I can't remember. No one has complained so I guess it just works !

What do you want to do with this PR - merge it?

kpschaper:
I think it has to do with the fact that some thermostats have a mode and mode_type.

Yes, please merge this PR and leave it like this!

Issue #228, opened 2021-11-30
lastcode broke MQTT JSON structure 


firegnom:
**Isssue with the lastCode json format**

Welcome, all I am very fresh in this community so sorry if this is already reported. 
I am running version :
```
EMS-ESP Version: v3.3.0
Device (Platform / SDK): ESP32 / v3.3.5-1-g85c43024c

```
My new boiler is Buderus GB212  and i just installed EMS-ESP , after the first configuration all worked with no problem, but then I have started to see that in MQTT Explorer the ems-esp/boiler_data was not parsed correctly by the MQTT server .

it turned out that the issue was:
```json
"lastcode":" 4/7760 start: 2060.149.11 11:1, end: none"
```

when I use lint on that message i get the error : 
```
Invalid characters in string. Control characters must be escaped.json(262)
```

I have attached sample MQTT messages - [sample.json.txt](https://github.com/emsesp/EMS-ESP32/files/7629055/sample.json.txt)
System log, that I have gathered during restart - [ALL.log](https://github.com/emsesp/EMS-ESP32/files/7629085/ALL.log)

I would really appreciate the help here.

firegnom:
Can confirm the same issue on : 
```
EMS-ESP Version
v3.3.1b0
```

```json
"lastcode":"�\f/7760 start: 2021.11.19 12:24, end: none"
```

[sample.json.2.txt](https://github.com/emsesp/EMS-ESP32/files/7629228/sample.json.2.txt)

firegnom:
Switched to : 

```
EMS-ESP Version
v3.2.1
```

All works correctly but of course last code is not in json

proddy:
It’s a simple fix. I’ll look into it later this week and make a patch.
Thanks for reporting.

pswid:
it's because in commit #220 following condition has been removed. 

if (telegram->offset > 0 || telegram->message_length < 14) {
     return;
}

For decoding last error code, we need 20 bytes starting from offset 0. So, now condition should look e.g. like this. Am I right?
if (telegram->offset > 0 || telegram->message_length < 20) {
    return;
 }

proddy:
yes, that's it. We need those extra bytes to accommodate the extra data. Do you want to fix with a PR?

proddy:
@firegnom can you try the latest dev build so we can close this, if it works

firegnom:
Hi, I can confirm It works now. I was running the latest version over the weekend all works now. 

Thank you!

Issue #204, opened 2021-11-18
"last error code" is not retrieved


pswid:
I'm currently using the latest development build v3.3.0b7 with Buderus WSW196i heat pump. It work's fine (although it shows a lot of parameters not used by heat pumps, so it would be worth giving the possibility to block/hide irrelevant parameters in the future versions, don't you think?). 

My heat pump currently (almost) all the time reports errors A01-5503 and H01-5504. However, EMS-ESP hardly ever shows "last error code". Restarting the EMS-ESP also deosn't help. I wonder, why? Maybe EMS-ESP should (at least at startup) "ask" the boiler for last error info? Maybe even more than only one, last error (on the heat pump display I can see the last 20 errors)?

Besides, in case the error is displayed on dashboard, there is also "uptime", like this:
**last error code          H01(5504) uptime:724749 min**
What is this uptime? Shouldn't it rather be decoded as the date and time the error occurred as it KM200 gateway does? In my case, response from KM200 looks e.g. like this:
**/notifications/Error-01          Time: 2021-11-16T14:57:34 -ErrorCode: A01 -AddCode: 5503 -Category: N
/notifications/Error-02          Time: 2021-11-16T14:57:34 -ErrorCode: H01 -AddCode: 5504 -Category: N**

If it helps I could make some tests if I only knew which EMS telegrams are related to error codes and what to look at...

tp1de:
A01/5503 indicates a problem with CAN-bus cable (disconnected or loose connection or to narrow to power cable)
H01/5504 indicates a problem with heat pump ventilation (filter to be cleaned or ventilation defect)

Most probably these are not permanent errors and overwritten afterwards. Within RC310 the last errors are stored locally while ems-esp polls the bus and shows the last status. I had a similar issue and RC310 error only disappeared on new start of heating system.

Within RC310 the date/time of first startup of the system is stored. This date + uptime should indicate the date-time of the error. But first startup is not available within km200 and as as I know neither in ems-esp.

Maybe Michael has another idea ...

pswid:
Errors are permanent and shows up after first compressor start since applying power to the HP.
[...]

Issue #124, opened 2021-09-22
introduce React Hooks for handling state in WebUI


proddy:
One thing I wanted to try out is replacing the class-based functions (HOCs) that handle local component states with React Hooks. Essentially using functional components to handle the setting of, listening to and updating state. It would make the code slightly easier to read, reduce the boilerplate size (e.g. the onComponent* calls can be replaced with useEffect()) and the state won't be passed around the map.

proddy:
First implementation is a hook for loading and saving settings data instead of extending RestControllerProps and using the HOC. The hook loads the data on mount and exposes the data, error message (if applicable) and save/load functions and a saving flag (boolean). This saving flag can be used so the form inputs may be disabled instead of the loading indicator which we currently use which causes the annoying "Loading..." flicker on form submits.

Doing this work in the `ft_reacthooks` branch

proddy:
Most of the work is done, a lot in the backend and a very cosmetic changes. This will be the basis for v3.4. The UI looks a little different now. It's in https://github.com/emsesp/EMS-ESP32/tree/ft_reacthooks if anyone wants to test drive it (@bbqkees, @MichaelDvP)

![image](https://user-images.githubusercontent.com/1230712/144507353-37e6d5bd-f976-4bc6-bef0-2dfb4226dfe5.png)

bbqkees:
Look great I'll try it out this week.

MichaelDvP:
First test results: 
Looks nice and code is better readable (as promised).

- The first commits causes some git issues, it could be fetched but checkout or merge was not possible. Later i see that it was a local path (C:\users\paul) in the fonts path (already removed). Because of this i loaded the zip and merged manual to my build. It's more work, but i can read and understand the code better. 
- You've updated most packages and changed nearly every file, is there a reason why you stay with react-router 5.3?
- mqtt path crashes for guest users, mqtt-status is set as [Admin-route](https://github.com/emsesp/EMS-ESP32/blob/8d67f2dd23641e41b0460e0a3b7b7d3675f1d2eb/interface/src/framework/mqtt/Mqtt.tsx#L29) and redirect is to status, this causes a loop and esp crashes. Remove admin from status and add to settings fixes this.
[...]

Issue #121, opened 2021-05-03
Bosch Condens GC2300W/Logamax Plus GB122 - missing error message


partikus:
**Bug description**

Once the boiler stopped working due to an issue, the error sent is not being handled. 

```
?(0xBF), data: 08 EA 00 06 11 20 2D 2D 0B 87 00 00 00 00 00 00 00 00 00 00 00 00 00 00
```

Based on [the error message type](https://emsesp.github.io/docs/#/MC110-controller?id=type-0xbf-deviceserrormessage) I've decoded the following values:

```
# pattern
[byte:{number}] | [value] - [notes]

[b:0] | 08 - [8] device ID
[b:1] | EA - [234] - model ID
[b:2] | 00 - [0] - blocking error: always 0x0A (fault class, see service manual) 
[b:3] | 06 - [6] blocking error: 0x00 or 0x10 = interlocking error, 0x11 = blocking error, 0x12 = system error, 0x13 = reset system error
[b:4] | 11 - [DC1] - device control ASCII | Fault code 0. character (possibly 0x0a = space)
[b:5] | 20 - [ ] ASCII | Fault code (operation code) 1st character
[b:6] | 2D - [E] Fault code (operation code) 2nd character
[b:7] | 2D - [E] additional code (status code)
[b:8] | 0B - [...
[b:9] | 87 - ...] - 2951 - data 
```

```
Boiler: Bosch Logamax Plus GB122 (DeviceID:0x08 ProductID:234, Version:05.03)
 This Boiler will respond to telegram type IDs: 0x10 0x11 0x14 0x15 0x16 0x18 0x19 0x1A 0x1C 0x2A 0x33 0x34 0x35 0xD1 0xE3 0xE4 0xE5 0xE6 0xE9 0xEA
 ```

**Device information**
```json
{
  "WIFI": {
    "ssid": "XXXX",
    "hostname": "ems-esp",
    "static_ip_config": "off"
  },
  "AP": {
    "provision_mode": 1,
    "ssid": "ems-esp",
    "local_ip": "192.168.4.1",
    "gateway_ip": "192.168.4.1",
    "subnet_mask": "255.255.255.0"
  },
  "MQTT": {
    "enabled": "on",
    "host": "192.168.98.134",
    "port": 1883,
    "username": "mqtt",
    "client_id": "ems-esp",
    "keep_alive": 120,
    "clean_session": "on",
    "max_topic_length": 128,
    "publish_time_boiler": 0,
[...]

PR #235, opened 2021-12-03
fix overlapping 0xC2 and overflow of offset

pswid:
Don't know why but while reading 0xC2 offset should be increased  by 27 not 25 (is it only valid for Buderus WSW196i heat pump?). Furthermore the patch breaks reading of next telegrams in case of offset overflow. Attached files show how it was and how it is now. Was over 200(!) unnecessary telegrams and first two bytes of each telegram are identical to the last two bytes from previous telegram (except the lines with offset overflow).
[before.txt](https://github.com/emsesp/EMS-ESP32/files/7648926/before.txt)
[after.txt](https://github.com/emsesp/EMS-ESP32/files/7648927/after.txt)

### DIFF
--- a/src/telegram.cpp
+++ b/src/telegram.cpp
@@ -613,19 +613,10 @@ uint16_t TxService::read_next_tx(uint8_t offset) {
     if (telegram_last_->offset != offset) {
         return 0;
     }
-<<<<<<< Updated upstream
-
-    uint8_t add_offset;
-    if (telegram_last_->dest == 0x08 && telegram_last_->type_id == 0xC2) {  //fix for overlapping 0xC2 telegrams from Boiler. Are other telegrams also affected?
-        add_offset = 27;
-    } else {
-        add_offset = 25;
-=======
     
     uint8_t add_offset = 25;                //for EMS+ telegram increase offset by 25
     if (telegram_last_->type_id < 0x100) {  //but for EMS1.0 by 27
         add_offset = 27;
->>>>>>> Stashed changes
     }
 
     if (UINT8_MAX - telegram_last_->offset < add_offset) {   //stop if new offset would overflow
--- a/src/telegram.cpp
+++ b/src/telegram.cpp
@@ -613,7 +613,19 @@ uint16_t TxService::read_next_tx(uint8_t offset) {
     if (telegram_last_->offset != offset) {
         return 0;
     }
-    add(Telegram::Operation::TX_READ, telegram_last_->dest, telegram_last_->type_id, telegram_last_->offset + 25, message_data, 1, 0, true);
+
+    uint8_t add_offset;
+    if (telegram_last_->dest == 0x08 && telegram_last_->type_id == 0xC2) {  //fix for overlapping 0xC2 telegrams from Boiler. Are other telegrams also affected?
+        add_offset = 27;
+    } else {
+        add_offset = 25;
+    }
+
+    if (UINT8_MAX - telegram_last_->offset < add_offset) {   //stop if new offset would overflow
+        return 0;
+    }
+
+    add(Telegram::Operation::TX_READ, telegram_last_->dest, telegram_last_->type_id, telegram_last_->offset + add_offset, message_data, 1, 0, true);
     return telegram_last_->type_id;
 }
 
--- a/src/telegram.cpp
+++ b/src/telegram.cpp
@@ -613,12 +613,19 @@ uint16_t TxService::read_next_tx(uint8_t offset) {
     if (telegram_last_->offset != offset) {
         return 0;
     }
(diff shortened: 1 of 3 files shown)


src/telegram.cpp
```diff
@@ -613,7 +613,19 @@ uint16_t TxService::read_next_tx(uint8_t offset) {
     if (telegram_last_->offset != offset) {
         return 0;
     }
-    add(Telegram::Operation::TX_READ, telegram_last_->dest, telegram_last_->type_id, telegram_last_->offset + 25, message_data, 1, 0, true);
+
+    uint8_t add_offset;
+    if (telegram_last_->dest == 0x08 && telegram_last_->type_id == 0xC2) {  //fix for overlapping 0xC2 telegrams from Boiler. Are other telegrams also affected?
```
MichaelDvP, line 618: > Are other telegrams also affected

@pswid  You're asking the right question. The answer is yes, all ems1.0 telegrams have 5 byte header and 27 data, ems+ telegrams 7 byte header and 25 byte data. Emsesp adds 0x100 to all ems+ telegrams, so you can check for type_id > 0xFF to decide what offset is needed. The overflow seems special to the C2 telegram, i have not seen this before, but my boiler sends only 26 bytes for 0x10 telegram with offset 0, but the telegram is much longer. You can check my implementation [here](https://github.com/MichaelDvP/EMS-ESP32/blob/dev2/src/telegram.cpp#L610-L624)

pswid: My solution was just a simple fix, yours is comprehensive and more professional. However, in your implementation, the condition checking if the date is valid (in boiler.cpp):

    // check for valid date, https://github.com/emsesp/EMS-ESP32/issues/204
    if (telegram->message_data[11] & 0x80) {

is wrong, because in my case it causes the error to be displayed in the form:
image
instead of:
image
BTW, some units are displayed incorrectly in your version (e.g. uA instead of minutes) see above.

There are also other telegrams, the reading of which caused the offset to overflow, e.g. 0xC6 & 0xC7 and probably some others.

MichaelDvP: Thanks, it should be `message_data[10]` to alllow #121 and #204. The units should work after page reload (browser cache).
But my personal build is a permanent alpha, no releases, and dont care about breaking changes. I only wanted to show another solution and clearify the telegram data length.  
I first see the overflow in #228 for the 0xC2 (my boiler does not have it), but i agree that if there is one telegram with this behaviour, probably there are others, and  overflow should always be checked.

proddy:
ok to merge?

pswid:
it works, but as @MichaelDvP noted, offset should be increased by 27 for every EMS1.0 telegram, not only 0xC2. @proddy will you  modify my code as bellow or should I create new PR?

change:
`uint8_t add_offset;`
`    if (telegram_last_->dest == 0x08 && telegram_last_->type_id == 0xC2) {  //fix for overlapping 0xC2 telegrams from Boiler. Are other telegrams also affected?`
`        add_offset = 27;`
`    } else {`
`        add_offset = 25;`
`    }`

to:

`   uint8_t add_offset = 25;                //for EMS+ telegram increase offset by 25`
`    if (telegram_last_->type_id < 0x100) {  //but for EMS1.0 by 27`
`        add_offset = 27;`
`    }`

src/telegram.cpp
pswid: You're right. After page reload units are OK. 
BTW, your personal build visually looks very good. I would even say it's nicer than the official version.

proddy:
create a new PR please, so we can track the changes

src/telegram.cpp
proddy: > You're right. After page reload units are OK. BTW, your personal build visually looks very good. I would even say it's nicer than the official version.

Looks like Michael is using my latest work from the ft_reacthooks branch. sneaky! It's not public yet, still adding some features but I plan to push it to dev in the next few days.

MichaelDvP: > Looks like Michael is using my latest work from the ft_reacthooks branch.

That's right, I like the new look and as proposed the code is better readable and easier to enhance.  
> sneaky! It's not public yet,

That's a lie. The branch (https://github.com/emsesp/EMS-ESP32/tree/ft_reacthooks) is public and you ask [here](https://github.com/emsesp/EMS-ESP32/issues/124#issuecomment-985022031) for testing it. Now you call it sneaky if someone do a test? Strange!

proddy: @MichaelDvP no need to be rude. What ever is bothering you let me know so we can continue to work together. You're are admin on this project and have made some great contributions, and I enjoy working together, but your recent attitude is concerning. Great you merged ft_reacthooks into your dev branch - any feedback on the new UI so far?

PR #279, opened 2022-01-04
V3.4 analog

MichaelDvP:
I've tested a bit witch adc and counter. The  arduino adc  is not really bug free, the attenuation works for mV reading only if set to all ([__analogAttenuation](https://github.com/espressif/arduino-esp32/blob/caef4006af491130136b219c1205bdcf8f08bf2b/cores/esp32/esp32-hal-adc.c#L197)). I've changed the terminal output for dallas and analog, please check if you like it.

### DIFF
--- a/src/emsdevice.cpp
+++ b/src/emsdevice.cpp
@@ -703,8 +703,8 @@ const std::string EMSdevice::get_value_uom(const char * key) {
 // except additional data is stored in the JSON document needed for the Web UI like the UOM and command
 // v = value, u=uom, n=name, c=cmd
 void EMSdevice::generate_values_web(JsonObject & output) {
-    output["type"] = to_string_short(); // device_type_name();
-    JsonArray data = output.createNestedArray("data");
+    output["label"] = to_string_short(); // device_type_name();
+    JsonArray data  = output.createNestedArray("data");
 
     for (auto & dv : devicevalues_) {
         uint8_t fahrenheit = 0;
--- a/src/system.cpp
+++ b/src/system.cpp
@@ -1047,6 +1047,23 @@ bool System::command_customizations(const char * value, const int8_t id, JsonObj
             sensorJson["offset"]  = sensor.offset; // o
         }
 
+        JsonArray analogJson = node.createNestedArray("analogs");
+        for (const AnalogCustomization & sensor : settings.analogCustomizations) {
+            JsonObject sensorJson = analogJson.createNestedObject();
+            sensorJson["gpio"]    = sensor.id;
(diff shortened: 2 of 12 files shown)


src/system.cpp
```diff
@@ -1047,6 +1047,23 @@ bool System::command_customizations(const char * value, const int8_t id, JsonObj
             sensorJson["offset"]  = sensor.offset; // o
         }
 
+        JsonArray analogJson = node.createNestedArray("analogs");
+        for (const AnalogCustomization & sensor : settings.analogCustomizations) {
+            JsonObject sensorJson = analogJson.createNestedObject();
+            sensorJson["gpio"]    = sensor.id;
+            sensorJson["name"]    = sensor.name;
+            if (sensor.type == AnalogSensor::AnalogType::ADC) {
+                sensorJson["type"]   = "adc";
```
proddy, line 1056: why switch from the enum to strings?

src/emsdevice.cpp
```diff
@@ -703,8 +703,8 @@ const std::string EMSdevice::get_value_uom(const char * key) {
 // except additional data is stored in the JSON document needed for the Web UI like the UOM and command
 // v = value, u=uom, n=name, c=cmd
 void EMSdevice::generate_values_web(JsonObject & output) {
-    output["type"] = to_string_short(); // device_type_name();
-    JsonArray data = output.createNestedArray("data");
+    output["label"] = to_string_short(); // device_type_name();
```
proddy, line 706: good catch! forgot about that one

proddy:
thanks for that Michael. I took a quick peek at the code and it looks great. I'll build and try it out later. One thing I did notice in the first version is the MQTT publishing would send the single payload followed by all the nested analog sensors so maybe you've addressed that bug too.

I'm about to push another large update that tidies up a few of the HA MQTT things that have been annoying me. I'll check in those first and then try out this PR. And after that I'll go back to doing PR's myself (it was getting boring reviewing and approving all my own changes!)

src/system.cpp
MichaelDvP: You're right, better to check the enum-format and put the strings to a flash-list.

MichaelDvP:
The enum is IOCOUNTER, but in web it is `Counter` I was not sure about the names and think it need adaptation, but both are clear, `Read` is very unspecific, `gpio` also not clear, `digital input` or `dig_in` should be better.

proddy:
agree, the sensor type for reading GPIO high/low state should be called 'digital input'. Shall we change it now and do it later?

PR #302, opened 2022-01-18
adapt limits

MichaelDvP:


src/emsdevice.cpp
```diff
@@ -1070,9 +1070,10 @@ bool EMSdevice::get_value_info(JsonObject & output, const char * cmd, const int8
 
             json["writeable"] = dv.has_cmd;
             // if we have individual limits, overwrite the common limits
+            // limits are not scaled with divider and temperatures are °C
             if (dv.min != 0 || dv.max != 0) {
```
proddy, line 1074: as this overrides the previous json[min] & json[max] assignments is there not a more efficient way of doing this? It's a costly operation

MichaelDvP: For now we have only 2 or 3 values with stored limits. It's a lot of work to examine all the limits. We can use a bool and check this for all limit settings.

proddy: ok. ArduinoJSON can be a pig sometimes

PR #304, opened 2022-01-19
mixer wwc and test for RC20

MichaelDvP:
Please check if the RC20 fix isworking, or if need to clear offset 28 onmode change.
For mixer i think there are some more values coming.


proddy:
nice, i'll check tonight when I'm home. thanks Michael

proddy:
did a quick test. set RC20 to Manual using the thermostat button, changed temp using the thermostat button -> not seeing this in  EMS-ESP. I'll watch the traffic on 0x17 and figure out what is going on

Issue #300, opened 2022-01-17
Moduline 300 'crashes' when in Manual mode


joepvanboekel:
**Bug description**
Moduline 300, does not show set temperature from EMS-ESP or vice versa under some circumstances

**Steps to reproduce**
step 1: set hc1 mode from auto to manual
step 2: thermostat can set 'hc1 selected room temperature' and is synced to ems-esp
<repeat step 2 as many times as you want, it stays ok>
step 3: ems-esp can set 'hc1 selected room temperature' and is NOT synced to thermostat
step 4: thermostat can set 'hc1 selected room temperature' and is NOT synced to ems-esp, also Boiler does not respond to set temperature from thermostat anymore

you can reset this behaviour by settings thermostat physically to AUTO and to MANUAL back, until you perform step3 again.

**Expected behavior**
Thermostat should always reflect / and or set the correct 'hc1 selected room temperature'

**Device information**
```json
{
  "System": {
    "version": "3.3.1b0",
    "uptime": "004+12:58:13.002",
    "uptime (seconds)": 392293,
    "freemem": 164,
    "reset reason": "Software reset CPU / Software reset CPU",
    "Dallas sensors": 0
  },
  "Network": {
    "connection": "Ethernet",
    "hostname": "ems-esp",
    "MAC": "11:22:33:44:55:66",
    "IPv4 address": "192.168.39.184/255.255.255.0",
    "IPv4 gateway": "192.168.39.1",
    "IPv4 nameserver": "192.168.39.1"
  },
  "Status": {
    "bus status": "connected",
    "bus protocol": "Buderus",
    "telegrams received": 187482,
    "read requests sent": 24,
    "write requests sent": 23,
    "incomplete telegrams": 1,
    "tx fails": 2,
    "rx line quality": 100,
    "tx line quality": 92,
    "MQTT": "connected",
    "MQTT publishes": 59258,
    "MQTT publish fails": 0,
    "Dallas reads": 0,
    "Dallas fails": 0
  },
  "Devices": [
    {
      "type": "Boiler",
      "name": "Nefit GBx72/Trendline/Cerapur/Greenstar Si/27i (DeviceID:0x08 ProductID:123, Version:06.08)",
      "handlers": "0x10 0x11 0xC2 0x14 0x15 0x1C 0x18 0x19 0x1A 0x35 0x16 0x33 0x34 0x26 0x2A 0xD1 0xE3 0xE4 0xE5 0xE6 0xE9 0xEA"
[...]

PR #310, opened 2022-01-21
tx read/write counters and devices handlers

MichaelDvP:
I've also added a warning for mqtt queue overflow
Texts are not the same in Web / system info / console and heartbeat. I don't wat to change too much.
I've tried with colored device handlers (commented out), but i think few lines more are clearer.
Check the publish_value. I think the visibilty should  affect this too.
Check generate_values, the bool-HA format as strings: afaik HA uses only mqtt, or does it affect api too?
in #300 there was a weblog buffer overflow. I've changed the memory allocation to buffered messages and a bit larger in maximum.


proddy:
in generate_values() booleans will be strings if HA is enabled, only for MQTT (not API). It uses `output_target == OUTPUT_TARGET::MQTT`

PR #320, opened 2022-01-24
helptext for string commands

MichaelDvP:
This is a small addition for a few  commands that needs a special formats, like thermostat datetime, switchtime, and holidays. 
Looks like this: 
Screenshot 2022-01-24 at 17-55-28 EMS-ESP
Screenshot 2022-01-24 at 17-56-33 EMS-ESP

Check if you ike it.

### DIFF
--- a/interface/src/project/DashboardData.tsx
+++ b/interface/src/project/DashboardData.tsx
@@ -14,6 +14,7 @@ import {
   DialogActions,
   MenuItem,
   InputAdornment,
+  FormHelperText,
   IconButton,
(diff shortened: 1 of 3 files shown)


proddy:
That is really clever. I saw this in an earlier build of yours and was hoping you'd port it over. very nice.

I was also thinking whether the shortname should be displayed anywhere in that dialog box. Sometimes I forget the shortname is and use the Customization page to look it up, which is not the purpose. What do you think about adding it somewhere?

Issue #321, opened 2022-01-24
tempautotemp disappeared


tp1de:
tempautotemp suddenly disappeared in actual dev version.

MichaelDvP:
No, only if not set, if it have a legal value it is shown. We discussed that before.

tp1de:
But if not seen within api it can't be set by api (iobroker) as well .....

proddy:
> tempautotemp suddenly disappeared in actual dev version.

That's because it does actually disappear. There was a long discussion about this (you really need to start watching this repo so you're up to date). And a result I had to code some logic into actually test and remove 'zombie' entities from the MQTT Discovery payload.

tp1de:
> And a result I had to code some logic into actually test and remove 'zombie' entities from the MQTT Discovery payload.

Why do you call it "zombie". On km200 it is called temporaryRoomSetpoint per hc is for me most used entity to change temporarily the hc room setpoint. 
On a RC310 you have to use it or change from automatic to manual mode. But then you have the risk to vhange it back again. We discussed this now a couple of times (sure I still remember).

Shall I provide the telegrams by changing out of km200?

proddy:
zombie because it would suddenly appear with a valid number, and hence included in the MQTT payload, and then later get a value of unset in the EMS telegram and then excluded. For example, in your case the number of Thermostat entities would bounce between 58 and 59. This confused Home Assistant as its expects static data. People were complaining of errors in the HA log because it couldn't find and render that sensor type. So now it's just removed and in the log you'll see the debug line saying 'lost device entity tempautotemp'

tp1de:
hc1: 01B9 offset 8:
 0: 48 10 FF 08 01 B9 00 D1
-1: 48 10 FF 08 01 B9 FF 2E
10: 48 10 FF 08 01 B9 14 C5
20: 48 10 FF 08 01 B9 28 F9

hc2: type: 01BA same offset

tp1de:
Everybody who has a RC310/RC300 ems+ Thermostat will need this entity. 
It shall always be present and should have valid range from 5 to 30 and two enum states 0:off and -1: automatic.
(in ioBroker it's a number with range from -1 to 30);

proddy:
it is present. I think we talking on different frequencies. I'm closing this issue.

tp1de:
?????

tp1de:
> it is present. where ???

[...]

PR #323, opened 2022-01-25
always show RC300 tempautotemp with minimum to -1°C #321

MichaelDvP:
to make an end to this stupid discussion, i add the RC300 tempautotemp with the value manipulation as requested from Thomas.
The limits are set to -1 to 30, otherwise we'll get complains for value -1 with limits 5-30.


proddy:
still undecided about this. I'll give it some thought.

PR #332, opened 2022-01-26
fix #329, numberformat in Edge, Chrome

### LINKED ISSUE #329: unable to set values in 0.5-Steps (temperature) in ems-esp-webinterface
In ems-esp 3.3 I used a thermostat-card in Home-Assistant to change temperatures. 

image

After upgrading to ems v3.4.0b1 I realized, that this card was no longer available and so I decided to use the "slider" in HA instead (eg. on Thermostat "hc manual temperature"), but then I realized, that it was somehow difficult to change the values (temperature range for that value was unfortunaly from 0 to 127 degrees C).

image

So I thought, it might be easier to change the temperature values in ems-ESP. But then I discovered the next surprise: in ems-esp you can only change the values in 1 degree steps..... and all entities that have a value that ends with ".5" were not displayed in the "change-Window". That field's were completly empty. And setting values in 0.5-Steps was impossible. 

image

**Request:** to make it possible to enter xx.5-Values in ems-esp

_Note: originaly I mentioned this issue in #325, but was asked to open a new issue for that._

MichaelDvP:
Seems Chrome and Edge dont like Intl.NumerFormat() in this place. Now it works here with Edge/win10, Chrome/Android.

### DIFF
--- a/interface/src/project/DashboardData.tsx
+++ b/interface/src/project/DashboardData.tsx
@@ -239,7 +239,7 @@ const DashboardData: FC = () => {
               <ValidatedTextField
                 name="v"
                 label={deviceValue.n}
-                value={deviceValue.u ? Intl.NumberFormat().format(deviceValue.v) : deviceValue.v}
+                value={deviceValue.v}
                 autoFocus
                 sx={{ width: '30ch' }}
                 type={deviceValue.u ? 'number' : 'text'}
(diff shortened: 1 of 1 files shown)


proddy:
That's what I thought and tried the same thing as you did, didn't make any difference. Then noticed that it worked as expected (tested on Edge 97.0.1072.69 and Chrome 97.0.4692.99).

proddy:
I'll look into this next week (travelling for 10days). We should try and keep to the system locale and fix the root cause of the problem and not mix up . with ,

PR #369, opened 2022-02-20
optimized create/remove mqtt discovery topics

proddy:
remove the config topics before publishing the payloads so HA doesn't throw warning


MichaelDvP:
At first look it's all ok, but i've never looked accurate how the config is generated and removed. Since i can only check with mqtt-explorer what is created, i dont know if HA complains because of timing. 

One thing i realized is that you now again check both, seltemp and roomtemp, to generate climate. I thought it's also working now with only seltemp?

proddy:
well spotted! I first thought HA wouldn't complain about a missing currtemp in the Discovery climate component but actually, it does. If I exclude it then HA throws the warning 
`Template variable warning: 'dict object' has no attribute 'currtemp' when rendering '{{value_json.hc1.currtemp}}`

which kinda makes sense. So I added it back.

If you're ok with the PR please approve and merge it. thanks!

src/devices/thermostat.cpp
```diff
@@ -679,12 +671,21 @@ void Thermostat::process_RC10Monitor(std::shared_ptr<const Telegram> telegram) {
...
+    if (hc->ha_climate_created() && !has_values) {
+        // remove the whole climate component
+        char topic[Mqtt::MQTT_TOPIC_MAX_SIZE];
+        snprintf(topic, sizeof(topic), "climate/%s/thermostat_hc%d/config", Mqtt::base().c_str(), hc->hc_num());
+        Mqtt::publish_ha(topic);
+    } else {
+        // create
         publish_ha_config_hc(hc);
         hc->ha_climate_created(true); // only create it once
     }
```
MichaelDvP, line 691: Is it intended that this publishes/removes in every call, every 10 sec on normal monitor-broadcast?

proddy: I'll fix it

MichaelDvP: Hmm, now it is created once, removed if seltemp is missing, but never created again. What about roomttemp now?
I think better is to remove ond re-add:
```
    if (hc->ha_climate_created()) {
        if (!has_values) {
            // remove the whole climate component
            char topic[Mqtt::MQTT_TOPIC_MAX_SIZE];
            snprintf(topic, sizeof(topic), "climate/%s/thermostat_hc%d/config", Mqtt::base().c_str(), hc->hc_num());
            Mqtt::publish_ha(topic);
            hc->ha_climate_created(false); // remove flag, can be added again
        }
    } else if (has_values) {
        // create
        publish_ha_config_hc(hc);
        hc->ha_climate_created(true);
    }

```
Is it a solution for missing roomtemp (outdoor controlled) and HA enabled to set `hc->roomTemp = hc->selTemp;`

proddy: if seltemp goes, it will be created and it should be created again with the check ha_climated_created()

the roomtemp is optional and handled in the publish_ha_config() function

MichaelDvP: The config is created regardless of seltemp and afterwards removed if seltemp is not set. If seltemp is coming back, it is NOT recreated, because ha_climate_created() stays true, it is only set to false on hc creation. 

But seltemp is the major value of thermostat, it will never be absent, without seltemp a thermostat is no thermostat. So this code will never be executed and works as before. 

I think you can merge this PR to add the other changes. We come back later to the climate creation in #371, which need some more adaptations.

Issue #368, opened 2022-02-19
Excluding thermostat entities should also remove the climate discovery topic


proddy:
As reported by eltoro on Discord, the MQTT Discovery climate component needs the selected temperature, room temperature and mode to render correctly. If any of these entities are disabled/excluded in EMS-ESP via the WebUI customization page, HA will report warnings constantly.

to reproduce: use the Web Customization and exclude `hc1/seltemp` or `hc1/currtemp` and you'll see warnings in HA like
```
Logger: homeassistant.components.mqtt.climate
Source: components/mqtt/climate.py:470
Integration: MQTT ([documentation](https://www.home-assistant.io/integrations/mqtt), [issues](https://github.com/home-assistant/home-assistant/issues?q=is%3Aissue+is%3Aopen+label%3A%22integration%3A+mqtt%22))
First occurred: 15:26:36 (12 occurrences)
Last logged: 15:28:26

Could not parse temperature from
```

The fix is to remove the MQTT /climate/config topic for a heating circuit if the required entities are not active (i.e have no value) or marked as excluded.

MichaelDvP:
I'm testing to move the HA climate from thermostat to mqtt. Depending on roomTemp active the climate is created with or without currtemp and disable roomtemp in customizations also disables climate. 
It's a suggestion,i'll make a PR, check first if you find it usefull.

proddy:
thanks, it would be easier when moving to the mqtt class as it exposes the DeviceValue class (dv) and you can check the flags like ACTIVE and VISIBLE. I'm thinking of extending the flags to support that other enhancement so you can choose per device entitity whether to exclude it from API, MQTT & Web/Console

proddy:
@MichaelDvP implemented this, I've tested and it works flawlessly., Nice work Michael! I think we can close this, bump to the version to b8

PR #375, opened 2022-02-23
move HA climate to mqtt

MichaelDvP:
as mentioned in #368, please check if you like it and if it works with HA. 
It's not perfect, disable seltemp or mode is not checked.

### DIFF
--- a/src/devices/thermostat.cpp
+++ b/src/devices/thermostat.cpp
@@ -333,8 +333,6 @@ std::shared_ptr<Thermostat::HeatingCircuit> Thermostat::heating_circuit(std::sha
     auto new_hc = std::make_shared<Thermostat::HeatingCircuit>(hc_num, model());
     heating_circuits_.push_back(new_hc);
 
-    new_hc->ha_climate_created(false); // set flag saying we're ready to create the MQTT Discovery Climate topic for this hc
-
     // sort based on hc number so there's a nice order when displaying
     // TODO temporarily commented out the HC sorting until I'm 100% sure the return object still references the newly created object
     // not sure if new_hc and heating_circuits_.back() will still reference the new HC after its sorted - to check!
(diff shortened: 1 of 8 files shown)


proddy:
It works, but the MQTT Discovery climate topic is created before the actual data is in the thermostat_data payload so HA will show the error:

 ```
Logger: homeassistant.helpers.template
Source: helpers/template.py:1834
First occurred: 10:58:22 (2 occurrences)
Last logged: 11:07:00

Template variable error: 'value_json' is undefined when rendering '{{ value_json.hc1.seltemp }}'
```

I fixed this in the earlier version by only creating the climate topic when selemp and currtemp are ACTIVE

**Comment by MichaelDvP** [comment]
Oh, you're right, i have not checked this. The climate is builded with seltemp if roomtemp is missing, so it's always created on first publish. 
I will go back to my first approach, create a `hc->climate` which is set to
```
if (hasValue(hc->roomTemp)) {
   hc-climate = 1;
} else if (hasValue(hc->selTemp)) {
   hc->climate = 0;
} else {
   hc->climate = EMS_VALUE_UINT_NOTSET;
}
```
I had this first as hidden without long_name, but there is a bug that resets all flags of hidden value on every publish and i did not find this flag-reset. But as i read that domotics have mqtt-discovery, but no climate, maybe we can publish the hc->climate as enum, showing the seltemp/roomTemp status and make it possible to disable climate in customizations.
I'll rework this PR to check.

MichaelDvP:
Ok, PR updated, please check now.

BTW: there were calls of `add_ha_climate(hc)` in settings telegrams where the mode is read, but the mode was never checked for climate creations, should there be a check?

proddy:
that fixed the error in HA now. So works well.

Yes, we really only need to call add_ha_climate for those functions that populate the seltemp as like you said, without it the climate component is pretty useless.

Also I'm still thinking about #368 and how to detect the deviceValue instead of the actual value. When you exclude seltemp from the Customationzation web page it will still go ahead and create the climate topic since it has a value (it's only not visible or included in the thermostat_data payload)

PR #378, opened 2022-02-26
check selTemp, roomTemp visibility for climate creation

MichaelDvP:
This could be a solution for #368.


proddy:
clever stuff! I forgot about value_p. testing now...

Issue #381, opened 2022-03-05
update static code analysis from codacy to sonar


proddy:
Moving to sonarcloud as it has better coverage and is good at finding any potential memory leaks.

https://sonarcloud.io/summary/new_code?id=emsesp_EMS-ESP32

proddy:
https://github.com/emsesp/EMS-ESP32/pull/382

proddy:
Sonar is invoked on every PR via a new GitHub action. The results are published for now only on the above web site. A secret is created called SONAR_TOKEN which binds the two. You can also invoke the check manually using the CLI (see scripts/run_sonar.sh). There were hundreds of recommendations and I went quickly through and applied some. There should be no breaking changes but the code is a lot cleaner. @MichaelDvP please have a look at the PR and if you're ok with it approve it.

PR #382, opened 2022-03-05
implement sonar checks and apply fixes - #381

proddy:


proddy:
I've integrated in sonarqube into the build process. It can be manually invoked via the CLI as in script/run_sonar.sh or on every pull in a new GitHub action.

src/helpers.cpp
```diff
@@ -192,7 +192,7 @@ char * Helpers::render_boolean(char * result, bool value) {
 
 // render for native char strings
 char * Helpers::render_value(char * result, const char * value, const int8_t format __attribute__((unused))) {
-    strcpy(result, value);
+    strlcpy(result, value, sizeof(result));
```
MichaelDvP, line 195: the sizeof will always be a pointersize (4), not the stringlength.

proddy: oops, that's bad. I'll fix.

src/web/WebLogService.cpp
```diff
@@ -158,7 +158,7 @@ void WebLogService::loop() {
 // convert time to real offset
 char * WebLogService::messagetime(char * out, const uint64_t t) {
     if (!time_offset_) {
-        strcpy(out, uuid::log::format_timestamp_ms(t, 3).c_str());
+        strlcpy(out, uuid::log::format_timestamp_ms(t, 3).c_str(), sizeof(out));
```
MichaelDvP, line 161: Same sizeof issue here, to check you'll need to pass the size to the messagetime-function

proddy: or use strlen(out) ?

MichaelDvP: No, declare as `char * WebLogService::messagetime(char * out, const uint64_t t, size_t  bufsize) {`, than you can use 
`strlcpy`. 
You have to determine the size where the busser is declared. `char buffer[20]` have sizeof(buffer) as 20, but`char * pbuffer=buffer` gives sizeof(pbuffer) with result 4. strlen is undefined if the buffer is not initialized.

proddy: fixed

MichaelDvP:
looks good now, you also can declare the busffsize as const and replace the `25` in line 164/165  by `buffsize` to have all checks set to the right sizeof(time_string).

proddy:
thanks for checking.

PR #397, opened 2022-03-12
fix: RC30_N sets the wrong temperature with home assistant

jerome992:


proddy:
can you commit the PR to the dev branch?

Issue #398, opened 2022-03-13
Temperature threshold for outdoor reduction (RC35)


brezel1965:
Can you please make an entry for the threshold for canceling the reduce at the outdoor temperature. I call it hc1_noreduceoutdoor. 

thx

![TempschwelleAbschalt](https://user-images.githubusercontent.com/32033996/158048732-4ffeb8a2-0ad6-41af-b5c7-71b558fab3be.jpg)

https://emswiki.thefischer.net/doku.php?id=wiki:ems:telegramme#hk1betriebsart

MichaelDvP:
> Can you please make an entry for the threshold for canceling the reduce at the outdoor temperature.

Canceling the the reduce mode is one parameter above and already implemented as `noreducetemp` (defaults to -20°C). 
The marked parameter is for switching to reduce (below this temp,  default 5°C) when reducemode is set to outdoorhold.  Above this outdoor temperature the heating switches off in night. 
I can add the last three parameters with settings, Suggestion for names (internal value /  mqtt-name / display name):
```
MAKE_PSTR_LIST(reducetemp, F("reducetemp"), F("off/reduce switch temperature"))
MAKE_PSTR_LIST(vacreducetemp, F("vacreducetemp"), F("vacations off/reduce switch temperature"))
MAKE_PSTR_LIST(vacreducemode, F("vacreducemode"), F("vacations reduce mode"))
```
Is this ok?

brezel1965:
I had already found the noreducetemp.  It is as you described with the marked parameter, I use exactly this parameter in connection with a heat pump.  It's a good idea to add all three parameters, the other two I don't currently use.

Your suggestions would be good for me. ;-)

PR #401, opened 2022-03-14
add RC35 values #398

MichaelDvP:
This will break nothing, but maybe we have discussion about namings. Please decide if merge now or after release to v3.4.1b0.


src/devices/thermostat.cpp
```diff
@@ -3321,6 +3377,10 @@ void Thermostat::register_device_values_hc(std::shared_ptr<Thermostat::HeatingCi
         register_device_value(tag, &hc->party, DeviceValueType::UINT, nullptr, FL_(party), DeviceValueUOM::HOURS, MAKE_CF_CB(set_party));
         register_device_value(tag, &hc->tempautotemp, DeviceValueType::UINT, FL_(div2), FL_(tempautotemp), DeviceValueUOM::DEGREES, MAKE_CF_CB(set_tempautotemp));
         register_device_value(tag, &hc->noreducetemp, DeviceValueType::INT, nullptr, FL_(noreducetemp), DeviceValueUOM::DEGREES, MAKE_CF_CB(set_noreducetemp));
+        register_device_value(tag, &hc->reducetemp, DeviceValueType::INT, nullptr, FL_(reducetemp), DeviceValueUOM::DEGREES, MAKE_CF_CB(set_reducetemp));
```
proddy, line 3380: best to move the registering of new device values to the end so it doesn't break the sequencing used in the Customization service

PR #420, opened 2022-03-28
remove sorting deviceValues

MichaelDvP:
This is my suggestion, i've also removed the deviceValue.id, it not used anymore.
Looks like much change, but it's only the extra loop formatting.


proddy:
I just saw this after I merged my PR. Will look manually

proddy:
I have merged manually

Issue #412, opened 2022-03-27
Wrong UTC time


vmonkey:
**Bug description**
After installing the 3.4.0b9 version (updating from 3.3.1 today), ems-esp shows wrong UTC time. It is shifted by 1 hour, regardless of manual input or NTP. I am in the Prague/Europe time zone. I suspect this to be related to the change of winter-to-summer time.

**Steps to reproduce**
Change time manually or turn on NTP.

**Expected behavior**
Correct time to be shown. Currently, the time is shifted by 1 hour.

**Screenshots**
Correct time is shown in the statusbar. Correct UTC is that time minus 2 hours.
![Screenshot_2022-03-27-11-33-14-695_com android chrome](https://user-images.githubusercontent.com/770524/160281453-b3fc33bd-21a1-4da7-8c23-7bc055d89861.jpg)

**Device information**
{
  "System": {
    "version": "3.4.0b9",
    "uptime": "000+06:14:35.078",
    "uptime (seconds)": 22475,
    "freemem": 152,
    "reset reason": "Software reset CPU / Software reset CPU",
    "temperature sensors": 0
  },
  "Network": {
    "connection": "WiFi",
    "hostname": "ems-esp",
    "SSID": "hodnekrabic",
    "BSSID": "D8:07:B6:C8:96:B6",
    "RSSI": -53,
    "MAC": "7C:87:CE:49:A3:14",
    "IPv4 address": "192.168.68.113/255.255.255.0",
    "IPv4 gateway": "192.168.68.1",
    "IPv4 nameserver": "0.0.0.0"
  },
  "Status": {
    "bus status": "connected",
    "bus protocol": "HT3",
    "bus telegrams received (rx)": 25378,
    "bus reads (tx)": 6422,
    "bus writes (tx)": 19,
    "bus incomplete telegrams": 0,
    "bus reads failed": 0,
    "bus writes failed": 0,
    "bus rx line quality": 100,
    "bus tx line quality": 100,
    "MQTT status": "connected",
    "MQTT publishes": 10066,
    "MQTT publish fails": 0,
    "temperature sensors": 0,
    "temperature sensor reads": 0,
    "temperature sensor fails": 0,
    "analog sensors": 0,
    "API calls": 2,
    "API fails": 0
  },
"Devices": [
    {
      "type": "Boiler",
      "name": "Condens 2500/Logamax/Logomatic/Cerapur Top/Greenstar/Generic HT3",
      "device id": "0x08",
[...]

PR #424, opened 2022-03-29
try to fix #408 and #412

### LINKED ISSUE #408: NTP does not work
In the web interface of the gateway (version v3.4.0b8) I set the NTP server to time.google.com and click on save. The UTC-time does not change and stays jan 11 1970 ( zero?) even after a day or so, Another ntp-server like ntp.time.nl  makes no difference. What am I doing wrong?

The system has a Bosch Fr100 thermostat. This thermostat has also a time This time is correct, but I can't see it in the web interface.
How does the NTP time works with the thermostat time. 
Is the thermostat time ignored?

MichaelDvP:
As mentioned i can not reproduce the issues, but the ntp status shows only if there is ntp service configured, not if ntp time is received. So a typo in server name shows active, but time is not set. 
With this commit i check if ntp time is realy received within the last 2 hours (ntp updates every hour). So the status can be checked on web page (network time -> status) and in system/info from api. 
Should we also add it to mqtt (heartbeat) and/or log changes in ntp state?

### DIFF
--- a/lib/framework/NTPSettingsService.cpp
+++ b/lib/framework/NTPSettingsService.cpp
@@ -46,8 +46,10 @@ void NTPSettingsService::WiFiEvent(WiFiEvent_t event) {
 
 // https://werner.rothschopf.net/microcontroller/202103_arduino_esp32_ntp_en.htm
 void NTPSettingsService::configureNTP() {
+    emsesp::EMSESP::system_.ntp_connected(false);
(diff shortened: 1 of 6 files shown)


proddy:
very nice. I was looking before on how to validate the time is received correctly, but since it's asynchronous and low-end code in espressif the call functions don't return anything. Checking after 7201 seconds is a clever way.

PR #427, opened 2022-03-29
add NTP status to mqtt/ha (heartbeat) and log

MichaelDvP:
I think it's better to have it logged and also in mqtt. Here is a log if wifi is reconnected:
```
2022-03-29 18:32:19.809 I 19: [mqtt] MQTT disconnected: TCP
2022-03-29 18:32:22.949 I 20: [system] NTP disconnected
2022-03-29 18:32:22.949 I 21: [emsesp] Starting NTP
2022-03-29 18:32:22.949 I 22: [emsesp] WiFi connected with IP=192.168.0.116, hostname=ems-esp
2022-03-29 18:32:22.991 I 23: [system] NTP connected
2022-03-29 18:32:23.009 I 24: [mqtt] MQTT connected
```


proddy:
@MichaelDvP, are there any objections if I remove "ntp_status" from the MQTT `heartbeat` topic payload? I don't think we need to publish service on/off statuses every minute.

MichaelDvP:
You can remove it. 
I think it was because the timeout has changed and we got a lot of  dis-/reconnects. But it's stable now, no need to monitor in heartbeat.

PR #430, opened 2022-03-30
Errormessage

MichaelDvP:
There is also a commit with your suggested HA update for changing readonly. Please check.


proddy:
nice!!

Issue #444, opened 2022-04-04
3.4b11 problems


harvey637:
Hi,
me again :-)

working parts look nice, e.g. the marking of "no api" in the dashboard enties, saving customisation of thermostat

NOT WORKING:
saving ANY customisation for "boiler" gives error 413.
saving in "thermostat" and "heatpump" give no error and work. even saving NOTHING changed in "boiler" returns 413. 

infos: not using MQTT, old customisation from 3.4b10 (some later prerelease) not recognised

this is the boiler:
```
 "Devices": [
    {
      "type": "Boiler",
      "name": "Enviline/Compress 6000AW/Hybrid 7000iAW/SupraEco/Geo 5xx",
      "device id": "0x08",
      "product id": 172,
      "version": "01.16",
      "entities": 100,
      "handlers received": "0xBF 0xC2 0xD1 0xE3 0xE4 0xE5 0xE9 0x0494 0x0495 0x048F",
      "handlers fetched": "0x14 0xE6 0xEA 0x048D 0x048A",
      "handlers pending": "0x10 0x11 0x15 0x1C 0x18 0x19 0x1A 0x35 0x16 0x33 0x34 0x26 0x2A"
    },

```

cu
Harvey

proddy:
ok, its probably due to a recent change I made where I send all the entities back to the server, and you have 100 of them. I'll need to redesign a few things

harvey637:
Hi,
thank you for looking into this!

You are right, I have 100 entites only for boiler (around 60 of them are for warmwater, which I dont use and want to hide).
cu
Harvey

MichaelDvP:
This is again a to small buffer. But proddys last change was only to send always all entities, before it has worked with a few exclusions, but will give the same error if you exclude all. Proddy metioned that he plans to pack the buffer, which is some work to do. As an intermediate solution you can check with a larger buffer. I've made [a test](https://github.com/MichaelDvP/EMS-ESP32/releases) with extra debug message to check. If you use this and set a log-debug before sending the exclusions you will see a message like `[emsesp] Masked entities json size: 1307`. (This size is for my boiler with 67 entities). Yours will be more than 2048, the old limit for http:413.

harvey637:
Hi Michael,
tested your special version - WORKING!

This is the log entry: ` [emsesp] Masked entities json size: 2153`

In the first run I removed (hided from api and dashboard)  around 57 entities, making life for iobroker/ems-esp easier,

cu
Harvey

[...]

PR #447, opened 2022-04-04
changes per #444

proddy:
only send back entities that have changed, by saving the old state of the mask flag. This fixes the problem when the mask is reset in the customization UI (to 0) and it's not sent back to the server


proddy:
limit is 50, but I think with the latest memory changes it can be higher now

PR #449, opened 2022-04-05
added moduline 400 installation parameters

kwertie01:
Hi Proddy,

This is my first PR so be gentile with me if I made any mistakes :-)
I added 11 parameters of the 0xA7 telegram from the Moduline 400 (RC30) thermostat.

- intoffset           offset int. temperature sensor, by * 0.1 Kelvin
- language         language on Thermostat: 0 german, 1 dutch
- clockoffset       offset (in sec) to clock, 0xff = -1 s, 0x02 = 2 s
- building           building type: 0 = light, 1 = medium, 2 = heavy
- heatingpid       PID setting 00=1 01=2 02=3
- backlight          Keyboard lighting: (0x00 = off, 0xFF = on)
- brightness        Screen brightness 0F=dark F1=light
- autodst            Automatic change Daylight Saving time: (0x00 = off, 0xFF = on)
- preheating       Preheating in the clock program: (0x00 = off, 0xFF = on)
- offtemp            Set Temperature when mode is Off / 10 (e.g.: 0x0F = 7.5 degrees Celsius)
- mixingvalves    Number of Mixing Valves: (0x00=0, 0x01=1, 0x02=2)

Probably the parameter names and/or translations are not the best.


proddy:
Impressive for your first PR! The framework has morphed over the years into a bit of a Frankenstein so well done navigation through the code logic. I couldn't find any faults so happy to merge. I've assigned Michael to also review

kwertie01:
Calling your code "Frankenstein" sounds way too negative for all the work you (and Michael) put into this very nice project.
I made over time similar updates for the Moduline 400 in the now archived EMS-ESP version of your project but never gave it to the community by requesting a PR.
Now I upgraded to the ESP32 and needed to redo my updates and indeed the code morphed quite a bit since I started to use it in 2019. But I think that's a good thing, it evolves, and getting better every time. I must say that I prefer the current code compared to the old one btw. 😊
I still have some more functionality to add for the Moduline 400 like wwsettings (0x3A) and clock program (0x3F and 0x40) so I hope to make more PR's in the future if you don't mind.

MichaelDvP (approved):
👍

proddy:
only thing left to do is update the changelog and also fix the package-lock

kwertie01:
I see that I have a modified package-lock.json file which I didn't intentionally change myself. Do you know how this could have  happened? (so I can avoid it for the next time)

proddy:
It can happen during the build (npm install) probably because you're using an older version of node. Just don't check it in (discard action)

Gert05:
Impressive stuff. Thanks for sharing.

Issue #450, opened 2022-04-05
3.4 b11 - very small observations / clarifications


harvey637:
Hi,
saving customization works great - no problems with saving in chunks!

some very small "beautifying" of things:
thermostat:
- "date/time" -> "time/date" (as this is the order of infos: "10:13:31 05.04.2022", or change the value to the more often used term date/time)
- "hc1/summermode" = "on|off" (index is shown, value is more intuitive)

boiler:
- some entities are shown as index, where a value list "On|Off" is known to the api and send correctly, (HP heating, HP Compressor, ...)
- some entities have (short and long) names, but no value (e.g. mixing switch temperature (switchtemp)): could they be marked "no_api + hidden" by default?
- "wwstarts2" -> "dhw control starts2"

The "api/system/customization" does only contain some of the customisation, and I have a lot of it in product_id=172 (boiler) :
```
{
  "label": "customizations",
  "Customizations": {
    "sensors": [],
    "analogs": [],
    "masked_entities": [
      {
        "product_id": 200,
        "device_id": 56,
        "entities": [
          "08dewtemperature"
        ]
      },
      {
        "product_id": 158,
        "device_id": 16,
        "entities": [
          "02errorcode",
          "02lastcode",
          "03floordry",
          "03floordrytemp",
          "03building",
          "03minexttemp",
          "03wwsettemp",
          "03wwmode",
          "03wwsettemplow",
          "03wwcircmode",
          "03wwchargeduration",
          "03wwcharge",
          "03wwextra1",
          "03wwextra2",
          "03wwdisinfecting",
          "03wwdisinfectday"
        ]
      },
      {
        "product_id": 172,
        "device_id": 8,
        "entities": []
      }
    ]
  }
} 
```

please notice that this is NOT complains, just "customer" oberservations in MY real life installation. I want to help to improve even small things, hopefully also only small code changes :-)

Keep on your good work please!
cu
Harvey

harvey637:
Hi,
1: mine looks like this: "date/time 16:42:07 05.04.2022", changing one or the other maks it right :-)
2: summer_mode is "0", but e.g. control_mode is "optimized" (also see date/time ... )
![summer_mode](https://user-images.githubusercontent.com/10156426/161780769-428fe2a3-53d0-4e6f-9271-8b5af5ae7357.png)
in the api I get the correct valuelist:
```
{
  "name": "summersetmode",
  "fullname": "hc1 set summer mode",
  "circuit": "hc1",
[...]

PR #457, opened 2022-04-08
Changes #450

MichaelDvP:
Please check if you like the additions: 
- I've disabled the banner while uploading new firmware, giving more optical feedback that upload is started.
- Still i don't see why the bool hpheatingon is shown as number 255, Changed auto-declaration to bool to exclude a compiler issue. And send strings to web instead of native true/false.


proddy:
nice additions. thanks

PR #458, opened 2022-04-08
mqtt HA uses only json, disable single

MichaelDvP:


src/system.cpp
```diff
@@ -961,6 +961,7 @@ bool System::command_settings(const char * value, const int8_t id, JsonObject &
         node["publish_time_other"]      = settings.publish_time_other;
         node["publish_time_sensor"]     = settings.publish_time_sensor;
         node["publish_single"]          = settings.publish_single;
+        node["publish_2_command"]       = settings.publish_single2cmd;
```
proddy, line 964: what is this?

MichaelDvP: The option to to publish to command topic as set in web, and was missing in api/system/settings:
grafik

e.g. publish datetime to `/emsesp/thermostat/datetime` (command topic) or `/emsesp/thermostat_data/datetime` (data topic)
In ioBroker-mqtt the option creates a single read/writeable entitiy.  In mosquitto it creates a lot of commands, because mosquitto (in default confiig) send back to subscription.

proddy: got it, thanks

MichaelDvP: Maybe settings is better readable with only possible settings:. if HA enabled there is no need to show `publish_single`, 
or combine settings to: 
`node["publish_single"] = setting.publish_single ? settings.publish_single2cmd ? "command" : "data" : "off;`
But there is a milestone to upload settingsfile, and for this it's easier to use the native options.

Issue #469, opened 2022-04-14
Invalid value for number.thermostat_hc1_selected_room_temperature: 0 (range 5.0 - 29.0)


bavo:
**Bug description**
Home assistant spams the following message on every thermostat mqtt message when my selected_room_temperature is 0:

`[homeassistant.components.mqtt.number] Invalid value for number.thermostat_hc1_selected_room_temperature: 0 (range 5.0 - 29.0)`

My thermostat sets the selected room temperature to 0 when it is not heating. 

ems-esp32: v3.4.0b11
home assistant: 2022.3.7

proddy:
The range is hardcoded in https://github.com/emsesp/EMS-ESP32/blob/fdefd028124f203484704684ee58fff8cfa1fcda/src/devices/thermostat.cpp#L3639 

we can change it.

PR #472, opened 2022-04-19
Increase range for the selected_room_temperature variable

bavo:
Fixes issue #469


bavo:
@proddy, i just changed value of the variable but i dont know if this breaks any tests? 
Let me know if this breaks anything

proddy:
no its fine, nice. thanks for doing the PR. I forgot about this one.

Issue #470, opened 2022-04-14
Updating table component in the WebUI to support sorting & filtering


proddy:
Replacing MUI React Table component with https://github.com/table-library/react-table-library

https://github.com/proddy/EMS-ESP32/tree/newtable

harvey637:
Hi,
this sounds exciting!

By the way, will this also work with customization? That is asking for a "all selected" line to activate/deactivate the "favourite, read_only, hide_from_api, hide_from_dashboard" to work only on the selected entities. Mostly for peaple like me, who always have to deactivate around 60 DHW-entities, making 120+ clicks :-) And sad to say, that I have some rare reboots on rtc-watchdog, where all cusomization is lost, dont worry, I am used to live with that.

just thinking, not demanding - thank you
cu
Harvey

proddy:
Hi Harvey, yes, I had you mind when developing this. I'll add filters to the table so you can quickly group the customizations. I'll be sharing early versions so you can play around and give feedback.

proddy:
I made the first build. As usual, I got a bit carried away and touched a few other things. The latest customization code is not compatible with earlier 3.4 releases so any changes to the Analog sensors will be lost. 

The table columns are resizeable and you can sort on most of the columns. I've also added filters and the "set all" type buttons as requested by @harvey637 . Please take a look and let me know what you think.

Build is here:

[EMS-ESP-3_4_0b12-ESP32.bin](https://github.com/proddy/EMS-ESP32/releases/download/latest/EMS-ESP-3_4_0b12-ESP32.bin)

harvey637:
Hi,
first impressions with my personel ideas:
very nice, with some fine tuning to go .....
- the position to resize a column is not easy visable (a little more highlighting the selected columns header)
- initial column size abriviates values (eg. those that are enums like "optimized" for thermostat) in customizations, 
- filtering favourites with no favourites?
- filtering for content like "dhw" in the entity name? (makes is easy to dasable everything with warm water ...)
- sorting "up|down|default"
- favourites not visibly marked (no star) and not handled by sort, (they are somewhere due to sorting, only on|of for favourites

So my very short observations:
- sorting should separate favourites and non_favourites, and sort them individuelly
- filter by content (part of the entity lng name, like "all with dhw")
- unmark_all does not unmark read_only, which makes sense and is ok, only a bit irritating at first sight, maybe it should be greyed out too?

happy easter monday
Harvey

proddy:
[...]

PR #475, opened 2022-04-24
Changes to Table (sorting, format, search)

proddy:


proddy:
https://github.com/emsesp/EMS-ESP32/issues/470

PR #479, opened 2022-04-25
Cell height for multiline values

MichaelDvP:
some long text strings are shown multiline and need a larger cell:
grafik


proddy:
or keep the row height fixed (it looks nicer) and use ... when it's cut off, which is default behaviour built into the library so just need to enable it

proddy:
I've merged this into my next PR

PR #493, opened 2022-05-06
api entitiy attributes with hcs

MichaelDvP:
I had to change the `parse_command_string` to allow `api/device/hcx` without command. Check if i have missed something.

While testing i see that analog and dallas also needs this addition.

### DIFF
--- a/src/analogsensor.cpp
+++ b/src/analogsensor.cpp
@@ -435,15 +435,43 @@ void AnalogSensor::publish_values(const bool force) {
 // called from emsesp.cpp, similar to the emsdevice->get_value_info
 // searches by name
 bool AnalogSensor::get_value_info(JsonObject & output, const char * cmd, const int8_t id) const {
+    // make a copy of the string command for parsing
+    char command_s[30];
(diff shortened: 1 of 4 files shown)


proddy:
I just checked out the PR and ran run unit test (`make run; test api_values`), and it failed

```C
   request.url("/api/boiler/wwcirc");
   EMSESP::webAPIService.webAPIService_get(&request);
```

comes back with:

```
000+00:00:00.000 D 133: [command] [DEBUG] Calling boiler command 'rc' to retrieve attributes
000+00:00:00.000 E 134: [emsesp] Call failed with error: cannot find values for entity 'rc' (Error)
```

I haven't looked but suspect the string stripping in `Command::parse_command_string()` is not working?

MichaelDvP:
Yes, i forgot that there are commands starting with wwc. Stick to slashes does not work for `/api/thermostat/hc2` (show hc2/values), i think we should check the numbers, and/or nullbyte.

proddy:
works fine

Issue #500, opened 2022-05-14
support "Hybrid Manager HM200" Buderus HM200 heatpump module


scat703:
**Is your feature request related to a problem? Please describe.**
*A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]*
Buderus Hybrid Manager HM200 not known so far by ems-esp

**Describe the solution you'd like**
*A clear and concise description of what you want to happen.*
In my Buderus Logasys SL213 environment I see the boiler, RC310 and KM200. DeviceID 0x53 is unknown to ems-esp so far. From my perspective this should be the Buderus Hybrid Manager HM200, as this is the only additional device connected to the ems bus. The heat pump is connected to HM200 by canbus.

**Describe alternatives you've considered**
*A clear and concise description of any alternative solutions or features you've considered.*
Checked KM200 for HM200 values, but KM200 does not discover/recognize HM200 so far.

**Additional context**
*Add any other context or screenshots about the feature request here.*

MichaelDvP:
It is a first step to add the HM200, i think best to heatpump-class, [i've done it here](https://github.com/MichaelDvP/EMS-ESP32/releases/tag/latest). But you also want to have the measurements and settings from this module.
To determine the measurements, go to terminal and set `watch 53` to see all telegrams from this module. Write down names and values of all measured values you can see on the thermostat-display and we can search these values in the broadcasted telegrams. (`heatpump(0x53) -> All (0x00), unkown(0x..), data:...`), if a value changes while logging, not both values, we can better identify it.
For the settings also set the `watch 53` and change each HM200 setting on the thermostat, not only show up, really change the setting, confirm,, and then change it back. Note name and setting value. This should result in short broadcast telegrams with single value and offset.  (`heatpump(0x53) ->All(0x00) unknown(0x..), data: xx (offset: xy)`

MichaelDvP:
Correction: for the settings we expect a message from thermostat to H-module `Thermostat(0x10) -> heatpump(0x53) unknown(0x..), data: xx (offset: xy)`. If these messages are missing, it's also possible that the values are stored in thermostat, than we have a broadcast from thermostat and need to `watch 10` to fetch them.

scat703:
As the upcoming two weeks will be quite busy, I plan to gather all the information by the end of May or early June. I will upload them asap!

Joern-W:
Hi, 
[...]

PR #506, opened 2022-05-17
summermode for heatpump and HM200 hybridmodule

MichaelDvP:


proddy:
@MichaelDvP I'm back from travel again and catching up - are all these PRs good to go?

MichaelDvP:
Yes, HM200 is only first step to register the device, there will be some telegram adds in the future, see #500, so it's not a real enhancement now.
I'm not sure if one of the other small changes worth mentioned in changelog?

PR #515, opened 2022-05-21
emergencyops and emergencytemp for RC310

tp1de:


MichaelDvP (approved):

proddy:
@tp1de please also update the `CHANGELOG_LATEST.md` if you're adding any major new changes or anything that needs mentioning that would be a breaking change. I'm not sure this PR needs to be added so it's more of a general comment.

PR #526, opened 2022-05-26
disable web-exclude-button for hidden entities

MichaelDvP:


proddy:
I'm looking into another way for these special hidden entities. The problem is in the new React table the attribute `id` needs to be unique. So by chance it works now since it's empty with

https://github.com/emsesp/EMS-ESP32/blob/a47e0e82666dc8ac31a28a0742fd6992bd2f1913/src/emsdevice.cpp#L849

but it will crash the front-end if there are more than 1 special entities in the list.

proddy:
I have an alternative fix for this, will do a PR soon

PR #527, opened 2022-05-26
various minor changes, including managing hidden entities in customization screen

proddy:


interface/src/project/SettingsCustomization.tsx
```diff
@@ -464,7 +471,7 @@ const SettingsCustomization: FC = () => {
                           setMasks(['']);
                         }}
                       >
-                        <ToggleButton value="8" disabled={(de.m & 1) !== 0 || de.id === ''}>
+                        <ToggleButton value="8" disabled={(de.m & 1) !== 0 || de.id === '' || de.s === de.id}>
```
MichaelDvP, line 474: i think the check for `=== ''` is not needed anymore.

MichaelDvP (approved):

interface/src/project/SettingsCustomization.tsx
proddy: yes, good spot

MichaelDvP:
Another point, the reset have  short- and fullname "reset", we have to rename to different names now to allow to show in web ui.

proddy:
> Another point, the reset have short- and fullname "reset", we have to rename to different names now to allow to show in web ui.
> 
> And same for some other values with identical texts: heatmetering, pumpkick, doublematchflow, activated, status. Isn't it easier to leave fullname blank and use a combined string`(de.id + de.s)` as unique key?

that's an option. but how would we detect in the web UI if it's a hidden entity or not?

MichaelDvP:
I have not read the table documentation, so maybe i'm wrong. As you mentioned the unique id i thought this is this key: [https://github.com/emsesp/EMS-ESP32/blob/9d25f8049c92b024e33c0bceefc66306a008018d/interface/src/project/SettingsCustomization.tsx#L453](https://github.com/emsesp/EMS-ESP32/blob/9d25f8049c92b024e33c0bceefc66306a008018d/interface/src/project/SettingsCustomization.tsx#L453l) 
And changing to `key={de.id + de.s}` should make the key unique also with empty id.

proddy:
> I have not read the table documentation, so maybe i'm wrong. As you mentioned the unique id i thought this is this key: [https://github.com/emsesp/EMS-ESP32/blob/9d25f8049c92b024e33c0bceefc66306a008018d/interface/src/project/SettingsCustomization.tsx#L453](https://github.com/emsesp/EMS-ESP32/blob/9d25f8049c92b024e33c0bceefc66306a008018d/interface/src/project/SettingsCustomization.tsx#L453l) And changing to `key={de.id + de.s}` should make the key unique also with empty id.

the data object takes an array of nodes which must contain the id property. It's internally and not the `key` attribute of the map function. The solution I'll use for hidden entities is using a special char in the name (like a *) and handle the check in the typescript code

src/emsdevice.cpp
```diff
@@ -831,29 +829,27 @@ void EMSdevice::generate_values_web_all(JsonArray & output) {
...
         }
 
-        // shortname
-        if (dv.tag >= DeviceValueTAG::TAG_HC1) {
-            obj["s"] = tag_to_string(dv.tag) + "/" + read_flash_string(dv.short_name);
-        } else {
-            obj["s"] = dv.short_name;
+        // n is the fullname, and can be optional
+        // don't add the fullname if its a command
+        if (dv.type != DeviceValueType::CMD) {
```
MichaelDvP, line 843: Why do you skip the fullname for commands? I think show them as normal and delete line 842/843/853

proddy: there is only one command, in the boiler called reset and it shows up in the Customizations as `reset (reset)` which is silly. The shortname is also not relevant as it's not part of the API and MQTT. If you think we still need to show the shortname then I would suggest changing the fullname to something more descriptive like 'reset command' ?

MichaelDvP: But you skip the fullname for command and so it's handled as hidden, but it is not hidden, we want to have all options to show in web. Since you are using now the shortname as id, you can not skip the shortname.

MichaelDvP: You can modify formatName() to 
```
    if (de.n === undefined || de.n === de.id) {
      return de.id;
```
to avoid "reset (reset)" and "pumpkick (pumpkick)", "activated (activated)", etc. This have no effect to the customization logic.

proddy: Ok that's a much better option. Thanks

proddy: I'm still not sending the entity fullname if its a command as this is the only way the Customizations web page knows whether it's a command or not.

MichaelDvP:
You are mixing commands and hidden values with the result that now both are not working correctly. 
Why not add a `de.c` for commands, like in dashboard, or use de.n undefined for hidden and de.n empty for commands, but not the same for both.

proddy:
> You are mixing commands and hidden values with the result that now both are not working correctly. Why not add a `de.c` for commands, like in dashboard, or use de.n undefined for hidden and de.n empty for commands, but not the same for both. Correction: the de.c in dashboard is different aand not commandonly. don't use this to avoid misunderstanding.,

not sure I follow. It works for me. What is a 'hidden' entity?

MichaelDvP:
> not sure I follow. It works for me. What is a 'hidden' entity?

hidden iswhat you have introduced for id and ha-mode, a entitiy not shown in the web with missing fullname. Now it's only the climate generation, not shown in web and mqtt, no command but customization button for mqtt/api to disable the climate generation. Shoul look like this:
grafik

The command `reset` does not show in mqtt (but is subscribed via wildcard), because it have no value, but shows in web to send the right command. It is and has a command, so we could disable write, but than it should also disable the web. Should look like this:
grafik

But you handle both the same way without fullname.

I'm using now de.n undefined for hidden and de.n empty for command: 
```
        if (dv.type == DeviceValueType::CMD) {
            obj["n"] = "";
```
```
  function formatName(de: DeviceEntity) {
    if (de.n === undefined || de.n === de.id) {
      return de.id;
    } else if (de.n === '') {
      return 'Command: ' + de.id;
    }
    return de.n + ' (' + de.id + ')';
  }
```
```
                          if (de.n === '' && (de.m & DeviceEntityMask.DV_READONLY)) {
                            de.m = de.m | DeviceEntityMask.DV_WEB_EXCLUDE;
                          }
                          if (de.m & DeviceEntityMask.DV_WEB_EXCLUDE) {
                            de.m = de.m & ~DeviceEntityMask.DV_FAVORITE;
                          }
                          setMasks(['']);
                        }}
                      >
                        <ToggleButton value="8" disabled={(de.m & 1) !== 0 || de.n === undefined}>
                          <StarIcon sx={{ fontSize: 14 }} />
                        </ToggleButton>
                        <ToggleButton value="4" disabled={!de.w || (de.m & 3) === 3}>
                          <EditOffOutlinedIcon sx={{ fontSize: 14 }} />
                        </ToggleButton>
                        <ToggleButton value="2" disabled={de.n === ''}>
                          <CommentsDisabledOutlinedIcon sx={{ fontSize: 14 }} />
                        </ToggleButton>
                        <ToggleButton value="1" disabled={de.n === undefined}>
                          <VisibilityOffOutlinedIcon sx={{ fontSize: 14 }} />
                        </ToggleButton>
                      </ToggleButtonGroup>
```

proddy:
wow, you know the code better than I do. I'll incorporate your changes. One last thing, should we show the value for ha/climate creation. Now it looks like:

image

MichaelDvP:
Yes, i think leave the value to climate, it's a good feedback that it's active and shows the right value. (as you see above in my screenshot for ha disabled the value is empty).

proddy:
> Yes, i think leave the value to climate, it's a good feedback that it's active and shows the right value. (as you see above in my screenshot for ha disabled the value is empty).

ok, so if you're ok with PR can you approve and merge it? (or I'll do it, either way)

MichaelDvP (approved):

PR #532, opened 2022-05-29
Update to platformio2.3.0/IDF4/Arduino2.0, remove master themostat 

MichaelDvP:
I've updated the branches with new platform. I think we should also remove the master-thermostat.
To collect the mqtt the tags has to be resorted.
TAG_NONE only for collecting the tags in nested mqtt (wildcard), notused for tagging registered values
TAG_DEVICE_DATA for all data without leading tag
TAG_BOILER_DATA_WW for publishing as extra topic
TAG_DEVCIE_DATA_WW for publishing with tag dhw withing _data topic.


proddy:
woah, lots of change

PR #585, opened 2022-07-23
AM200, EM10 changes

MichaelDvP:
I've reduced to the changes that are confirmed working. 
The EM10 is now mapped to gateway, I don't see why interface fits better than gateway.
Valves and other entities from AM200 can be added later.
wwFlowTempOffset max is set to 45.


src/web/WebSettingsService.cpp
```diff
@@ -274,6 +274,7 @@ void WebSettingsService::onUpdate() {
 
 void WebSettingsService::begin() {
     _fsPersistence.readFromFS();
+    WebSettings::reset_flags();
```
proddy, line 277: good catch!

proddy:
shall we add a line to the changelog and bump up the version number?

MichaelDvP:
I think we should wait with version bump and changelog until the AM200 is completed. Or, maybe  add " Start integrating AM200" now?

proddy:
let's wait then

PR #586, opened 2022-07-24
link to device entity details

proddy:
i made a small change in the customizations page that opens the details of a device entity when clicking on the name in the customization list. I found sometimes I had to look up in the code to find which type an entity was so this was handy to have


proddy:
image

proddy:
one thing I discovered is that if you have multiple thermostats or mixers then the API becomes /api/{device type}{n}/....

not sure if this works though

MichaelDvP:
> one thing I discovered is that if you have multiple thermostats or mixers then the API becomes /api/{device type}{n}/....

No, not working (api) and not necessary. If we have multiple devices of one class, they have different entities or different tags. Also mqtt uses only one nested topic for device class, not for single devices. 

Links for single devices work well.

proddy:
> > one thing I discovered is that if you have multiple thermostats or mixers then the API becomes /api/{device type}{n}/....
> 
> No, not working (api) and not necessary. If we have multiple devices of one class, they have different entities or different tags. Also mqtt uses only one nested topic for device class, not for single devices.
> 
> Links for single devices work well.

ok then I can remove the {n}. 

So if you have multiple thermostats there can be only one hc1 for example?

MichaelDvP:
> So if you have multiple thermostats there can be only one hc1 for example?

Yes, with master 0x10 and 0x18, the 0x18 is remote and only sends current room temp, if there is no 0x10, 0x18 manages hc1, 0x19 hc2 and so on. dhw is only in the first thermostat, the others don't have dhw. 
Because thermostats can be remotes or manage hcs, we have some registered entities without values, the link will than get the value from master.  
This is my remote from hc2 (not a real remote thermostat, only the emulated roomcontrol.cpp from ems-esp), 
grafik
grafik

proddy:
ok, I removed the /{n}

is this good to go?

MichaelDvP:
> is this good to go?

Yes, looks fine.

MichaelDvP (approved):

Issue #591, opened 2022-08-04
Add a tip/warning on the screen when the bus stays disconnected


bbqkees:
If you have an E32 and you reset it or upload new firmware (via USB), then the flash is erased and all settings are gone.
Most users will pick the default binary from the repository, which has a default board profile 'custom'.

So the firmware update goes well, and the E32 reboots. But now it cannot find the EMS bus anymore.

Now both EMS-ESP and the Gateway hardware development have matured over the years I hardly get emails with support requests.
If so, 70% is with this issue (the other 30% either MQTT or HA config issues).

Can we add a note on the Status page pinpointing people to possible solutions?
Like 'Bus disconnected. Check wiring and board profile'.

MichaelDvP:
If the bus isn't connected the dashboard permanently shows a warning banner "Scanning for EMS devices...". 
Should we change this (in case of bus disconnect) to an error-banner like this:
![grafik](https://user-images.githubusercontent.com/59284019/182906760-7762336d-7bb2-4762-b2cb-536d5f75ef1f.png)
Or do you think of something different?

bbqkees:
Yes like that but maybe say something like 'check settings and board profile'.
For non-technical users it is not trivial at all that the board profile may be the cause of the issue. Better point them directly to it.

Issue #590, opened 2022-08-01
Add Buderus RC100(H) support


MichScha:
Hi, I have a combination of
- Enviline/Compress 6000AW/Hybrid 7000iAW/SupraEco/Geo 5xx
- Thermostat: RC300/RC310/Moduline 3000/1010H/CW400/Sense II
- KM200
- RC100(H)

The device RC100(H) seems to be recognized as HP module and has a humidity sensor integrated in contrast to the RC100 device
```
Heatpump: HP Module (DeviceID:0x38)
 Received telegram type IDs:
 Fetched telegram type IDs:
 Pending telegram type IDs: 0x42B 0x47B
 Ignored telegram type IDs: 0xF7 0x01 0xBF
```

Captures telegrams
```
000+05:50:21.619 TRACE 16583: [emsesp] Heatpump(0x38) -> Thermostat(0x10), HP1(0x042B), data: 00 EA 09 22 01
000+05:50:21.619 DEBUG 16584: [emsesp] No telegram type handler found for ID 0x42B (src 0x38)
000+05:50:21.656 TRACE 16585: [emsesp] Heatpump(0x38) -> Thermostat(0x10), HP2(0x047B), data: 10 38 00 9B
000+05:50:21.656 DEBUG 16586: [emsesp] No telegram type handler found for ID 0x47B (src 0x38)

From another day
003+07:44:12.385 N 1610: [emsesp] Heatpump(0x38) -> All(0x00), ErrorMessage(0xBF), data: 38 C8 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
003+07:44:16.774 N 1611: [emsesp] Heatpump(0x38) -> Thermostat(0x10), HP1(0x042B), data: 00 E9 09 1E 01
003+07:44:16.812 N 1612: [emsesp] Heatpump(0x38) -> Thermostat(0x10), HP2(0x047B), data: 0F 37 00 96
003+07:51:16.241  NOTICE 1163927: [emsesp]     Heatpump(0x38) -> Thermostat(0x10), HP1(0x042B), data: 00 E9 09 1D 01
003+07:51:16.241   DEBUG 1163928: [emsesp]     No telegram type handler found for ID 0x42B (src 0x38)
003+07:51:16.279  NOTICE 1163929: [emsesp]     Heatpump(0x38) -> Thermostat(0x10), HP2(0x047B), data: 0F 37 00 96
[...]

PR #597, opened 2022-08-08
AM200, update version

MichaelDvP:
I think the AM200 integration is now completed. I've count up version and added changelog.
For EM10 and RC100H there is no further feedback, i think it's ok for now.
I've added also the banner #591, check if you are ok with it.


MichaelDvP:
Seems there is some more to do for RC2100H support, #597. Closing this PR for first test in my repo and make a new PR if it works.

proddy:
Ok. Looked good though! I'm working on the language translations and touched a lot of code and made some fixes here and there so it's going to be interesting when these two merge. I think we should get AM200 out as Stable and then follow up with the 3.5 with the translations.

MichaelDvP:
Sorry typo, i've meant RC100H in #590 needs some more checks and additions.
With the new parameters i'm not really happy, it's a bit mixed with hc-> and device mapping because remote-device is always only for one hc and all need to be tagged with hc to show up in thermostat mqtt.

PR #613, opened 2022-09-09
fix #42

### LINKED ISSUE #42: Mqtt client does not reconnect if mqtt server is temporary down
I did some software updates on my home automization system. Therefore the mqtt server was temporary down (15 minutes).
After restarting my other mqtt clients reconnected automatically but ems-esp32 does not. I had to reconnect manually by webui.
I tested it twice, but automatic reconnection does not work if mqtt server was down.

I did some more testing. If I stop the mqtt broker for a couple of minutes then reconnection works.
When I disconnect the LAN-cable from the server where the mqtt broker is running on, then ems-esp does not recognize a "disconnected" status - It stays "connected" (I tested for 3-4 minutes - IP-Address not reachable). When I just stop the service but IP stays reachable then the ems-esp recognize the disconnect status immediatly.

MichaelDvP:
mainly it's call `onConfigUpdated()` in `MqttSettingsService

### DIFF
--- a/interface/package-lock.json
+++ b/interface/package-lock.json
@@ -8,15 +8,15 @@
       "name": "EMS-ESP",
       "version": "3.4.0",
       "dependencies": {
-        "@emotion/react": "^11.10.0",
-        "@emotion/styled": "^11.10.0",
-        "@msgpack/msgpack": "^2.7.2",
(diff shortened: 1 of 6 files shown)


proddy:
could we add to 3.5.0 too?

PR #620, opened 2022-09-14
Add support for Lolin C3 mini

CheeseE:
Adding support for Lolin C3 mini: [link](https://www.wemos.cc/en/latest/c3/c3_mini.html)
This board is pin to pin compatible with the old D1 mini. No need to cut the enclosure when upgrading from old gateways, and because of the low power RISC-V CPU it can be still powered from EMS bus.


src/emsdevice.cpp
```diff
@@ -768,7 +768,7 @@ void EMSdevice::generate_values_web(JsonObject & output) {
                     if (divider > 0) {
                         obj["s"] = Helpers::render_value(s, (float)1 / divider, 1);
                     } else if (divider < 0) {
-                        obj["s"] = Helpers::render_value(s, (-1) * divider, 0);
+                        obj["s"] = Helpers::render_value(s, (float)(-1) * divider, 0);
```
proddy, line 771: what is the reason behind adding this?

src/mqtt.cpp
```diff
@@ -1311,8 +1311,8 @@ void Mqtt::publish_ha_climate_config(uint8_t tag, bool has_roomtemp, bool remove
         doc["curr_temp_tpl"] = currtemp_s;
     }
 
-    doc["min_temp"]  = Helpers::render_value(min_s, 5, 0, EMSESP::system_.fahrenheit() ? 2 : 0);
-    doc["max_temp"]  = Helpers::render_value(max_s, 30, 0, EMSESP::system_.fahrenheit() ? 2 : 0);
+    doc["min_temp"]  = Helpers::render_value(min_s, (uint32_t)5, 0, EMSESP::system_.fahrenheit() ? 2 : 0);
+    doc["max_temp"]  = Helpers::render_value(max_s, (uint32_t)30, 0, EMSESP::system_.fahrenheit() ? 2 : 0);
```
proddy, line 1315: and also what is the reason behind adding these two casts?

proddy:
looks ok, I've added a comment. @MichaelDvP do you want to review too?

lib/framework/NetworkSettingsService.cpp
```diff
@@ -75,6 +75,9 @@ void NetworkSettingsService::manageSTA() {
         });
 
         WiFi.begin(_state.ssid.c_str(), _state.password.c_str()); // attempt to connect to the network
+        #ifdef ARDUINO_LOLIN_C3_MINI
+            WiFi.setTxPower(WIFI_POWER_8_5dBm); //https://www.wemos.cc/en/latest/c3/c3_mini.html#about-wifi
```
MichaelDvP, line 79: A few lines above we set ' esp_wifi_set_max_tx_power'  to a selectable value, here it's overwritten by fixed value? Why? The link with` about-wifi` does not work.

MichaelDvP (commented):
Nice that C3 works with little changes.
The C3 is set on compile-time, but the board-profile can changed in settings to a non-C3, Is this usefull? Also on ESP32 the C3 can be selected.
I think it's better to fix board in cpp code to C3 mini and in webcode disable selection box if C3Mini is set in program, otherwise dont add C3Mini to selection box.

src/emsdevice.cpp
CheeseE: It's interesting because this was causing compile issues for me even without any modification. (src/emsdevice.cpp:771:78: error: call of overloaded 'render_value(char [10], int, int)' is ambiguous) I'm using the 5.1.1 version of the espriff 32 lib.

src/mqtt.cpp
CheeseE: same as above

lib/framework/NetworkSettingsService.cpp
CheeseE: Haha, they have changed this wiki overnight. They updated the board version now to v2.1, I still have the v1, which has a bug that only works with this setting. It was on the wiki. Seems like they have fixed this with the new version.
https://web.archive.org/web/20220825213136/https://www.wemos.cc/en/latest/tutorials/c3/get_started_with_arduino_c3.html

MichaelDvP: Thanks, now it's clear.

CheeseE:
> Nice that C3 works with little changes. The C3 is set on compile-time, but the board-profile can changed in settings to a non-C3, Is this usefull? Also on ESP32 the C3 can be selected. I think it's better to fix board in cpp code to C3 mini and in webcode disable selection box if C3Mini is set in program, otherwise dont add C3Mini to selection box.

Hi Michael,
Thanks for the comment.
Do you know how to use compile time env vars in TS? So that I can export the board profiles list based on the board type/env type.

src/uart/emsuart_esp32.h
```diff
@@ -26,7 +26,11 @@
 
 #define EMS_MAXBUFFERSIZE 33 // max size of the buffer. EMS packets are max 32 bytes, plus extra for BRK
 
-#define EMSUART_NUM UART_NUM_2 // on the ESP32 we're using UART2
+#ifdef ARDUINO_LOLIN_C3_MINI
+    #define EMSUART_NUM UART_NUM_1 // on C3 mini we're using UART1
```
MichaelDvP, line 30: I think you have already tested the code. Don't you have conflicts with the serial console on uart1?
Disabling the serial console is [in emsesp.cpp](https://github.com/emsesp/EMS-ESP32/blob/70efa4f4e65331bba8dc9fa88d2bc6707bf9b7eb/src/emsesp.cpp#L1282).

proddy:
Also @CheeseE we'll be pushing the dev branch to main soon so suggest this PR goes into the next dev build. just a FYI

src/uart/emsuart_esp32.h
CheeseE: Isn't the serial console on uart0 by default? c3 has 2 ports, uart0 and uart1, where uart0 is for uploading firmware and serial log: https://docs.espressif.com/projects/esp-at/en/latest/esp32c3/Get_Started/Hardware_connection.html

MichaelDvP: You're right, serial is uart0, sorry for confusing.

MichaelDvP (approved):

proddy:
@CheeseE can you merge your change into the latest dev and ammend the PR?

lib/framework/NetworkSettingsService.cpp
CheeseE: Hi @MichaelDvP , I just checked and it does not work with other values, seems to be a hardware issue with this board version.

MichaelDvP: Ok,thanks, will keep an extra pio-env for board v1.0.

PR #627, opened 2022-09-19
Fallback to en on empty entity string

MichaelDvP:
We have some missing entities in NL like building. No entry shits SE to that index, so we need an empty string that should fal back to EN.


proddy:
good catch

Issue #657, opened 2022-10-02
Actual room temperature should be condisder the offset


tefracky:
**Is your feature request related to a problem? Please describe.**
It is possible to set a possible offset to the measured room temperature of the thermostat. The actual temperature (hc1/currtemp) does not consider this offset and therefore, also the resulting MQTT climate entity shows the wrong room temperature.

**Describe the solution you'd like**
Either the current room temperature should directly consider this offset or an additional offset-corrected temperature should be provided.

**Describe alternatives you've considered**
In Home Assistant, custom climates can be defined. Hence, the room temperature can be calculaded through a template. It works, but a direct implementation in the EMS-system would be better.

MichaelDvP:
This is a function of the thermostat. There is an setting to correct the internal sensor and the resut on temperature is used by the thermostat for control and is send as roomtemp to ems-esp. If this setting is not implenmented for your thermostat, please determine it by using settings on the thermostat and log the change, then we can implement it.

tefracky:
Ahhh, you are right! I used "Raumtemperatur Offset" (room temperature offset?), but I have to use "Fühlerabgleich" (sensor calibration?)... But unfortunaly, I did not found "sensor calibration" in the EMS-ESP. Maybe I'm just incompetent, but does the function exist somewhere or is it not implemented?
I use a RC300 thermnostat from Bosch/Buderus.
Thanks a lot!

MichaelDvP:
We have it for RC30, RC35 and RC100H, for RC300 it is unknown. Maybe the same as RC100H, check telegram 0x273.

tefracky:
I made some tests and it seems to be 0x0240.

This is the output:

Offset = 0.0       Thermostat (0x10) -B-> All(0x00), RC300Settings(0x0240), data: 00 (offset 7)
Offset = 0.1       Thermostat (0x10) -B-> All(0x00), RC300Settings(0x0240), data: 01 (offset 7)
Offset = 0.2       Thermostat (0x10) -B-> All(0x00), RC300Settings(0x0240), data: 02 (offset 7)
Offset = 0.3       Thermostat (0x10) -B-> All(0x00), RC300Settings(0x0240), data: 03 (offset 7)
Offset = -0.1      Thermostat (0x10) -B-> All(0x00), RC300Settings(0x0240), data: FF (offset 7)
Offset = -0.2      Thermostat (0x10) -B-> All(0x00), RC300Settings(0x0240), data: FE (offset 7)
[...]

Issue #647, opened 2022-09-25
Invalid value for number.thermostat_hc1_selected_room_temperature: 30 (range 0.0 - 29.0)


nagyrobi:
**Bug description**
Home assistant spams the following message on every thermostat mqtt message when `thermostat_hc1_selected_room_temperature` goes to 30:

`[homeassistant.components.mqtt.number] Invalid value for number.thermostat_hc1_selected_room_temperature: 30 (range 0.0 - 29.0)`

This is on a Generic HT3, with a controller detected as thermostat FW200 (not having an actual EMS thermostat in the system), in a Junkers Ceraclass Excellence ZSC 28-3 MFA E 23.

Observed that this value is tied to climate.thermostat_hc1's `temperature` value. The thermostat entry has `max_temp: 30`, and when it's set to that, `number.thermostat_hc1_selected_room_temperature` tries to go there by itself to the same value.

Similar to https://github.com/emsesp/EMS-ESP32/issues/469, only that the top value is capped the code. Perhaps it would suffice not to hardcode the `max` to `30`, but instead make it equal with thermostat's `max_temp`.

ems-esp32: v3.4.2
home assistant: 2022.9.6

MichaelDvP:
@proddy Seems that bosch changes the limits with every device. Should we adapt the limit dynamically for ems-reported values out of range. e.g. add to `emsdevice::generate values`
```
                if ((output_target == OUTPUT_TARGET::MQTT) && (dv.min != 0 || dv.max != 0)) {
                    if (json[name].is<float>() || json[name].is<int>()) {
                        int v = json[name];
                        if (fahrenheit) {
                            v = (v - (32 * (fahrenheit - 1))) / 1.8; // reset to °C
                        }
                        if (v < dv.min) {
                            dv.min = v;
                            dv.remove_state(DeviceValueState::DV_HA_CONFIG_CREATED);
                        } else if (v > dv.max) {
                            dv.max = v;
                            dv.remove_state(DeviceValueState::DV_HA_CONFIG_CREATED);
                        }
                    }
                }
``` 
This should hopefully end all HA complains about min/max values.

MichaelDvP:
I thought about user defined min/max limits without wasting a lot of RAM.  I think i've found a solution storing user defined values in the customization in a single string only if set e.g. `00hc2/daytemp|Tagestemp>5<30` if Name and both limits set . The input looks like this:
[...]

PR #660, opened 2022-10-04
Climate min/max #647, RC300 CalIntTemp #657

MichaelDvP:
Also added counting tx fails for skipped (queue overflow) telegrams.


proddy:
I've also manually merged this into the flash_optimization branch

Issue #671, opened 2022-10-09
REST-API - strange characters in http://<hostname>/api/<device>/commands output


evroom:
Hi,

REST-API calls for `http://<hostname>/api/<device>/commands` have strange characters in the output.
Seen after updating to EMS-ESP Version: v3.5.0b5 (from v.3.4.2b5).

Example: `http://ems-esp.local/api/Boiler/commands`

```
{
  "boilhystoff": "•",
  "boilhyston": "•",
  "burnmaxpower": "¥¦§¨©ª«¬xV­ºL¢ü?$±ý?",
  "burnminperiod": "¥¦§¨©ª«¬xV­ºL¢ü?$±ý?",
  "burnminpower": "¥¦§¨©ª«¬xV­ºL¢ü?$±ý?",
  "commands": "lists all commands",
```

Same for the devices Thermostat and Mixer.

Is this a bug or a local issue ?

Greetings,

E.J.

evroom:
Here my system info:

[emsesp_info_v.3.5.0b5.txt](https://github.com/emsesp/EMS-ESP32/files/9741307/emsesp_info_v.3.5.0b5.txt)

proddy:
Definitely a bug. Thanks for reporting

evroom:
Solved with EMS-ESP Version v3.5.0b6
(EMS-ESP-CI-3_5_0b6.bin)

`http://ems-esp.local/api/Boiler/commands`
```
{
  "boilhystoff": "hysteresis off temperature",
  "boilhyston": "hysteresis on temperature",
  "burnmaxpower": "burner max power",
  "burnminperiod": "burner min period",
  "burnminpower": "burner min power",
  "commands": "lists all commands",

```
Thanks for the quick fix !

evroom:
Solved with EMS-ESP Version v3.5.0b6
(EMS-ESP-CI-3_5_0b6.bin)

PR #672, opened 2022-10-09
fix REST-API - strange characters in description

proddy:
…ommands output #671


MichaelDvP (approved):
Yes i think this works, but i don't like to register only the translated fullname to command. 
If you later change language the translation of registration stays.
It's a bit more work and need some more translated words for system and dallas commands, but it's cleaner to to have Command::add with the complete fullname-list and translate in Command::show() etc..

proddy:
> Yes i think this works, but i don't like to register only the translated fullname to command. If you later change language the translation of registration stays. It's a bit more work and need some more translated words for system and dallas commands, but it's cleaner to to have Command::add with the complete fullname-list and translate in Command::show() etc..

yes I agree, the translations should be dynamic like the rest of the code. It was a design principle I started with when writing the translation code so just forgot this. I'll see if I can implement it tomorrow

PR #676, opened 2022-10-09
NO and Commands translations

proddy:


MichaelDvP (approved):
Looks good, i'll add the german translations with my next PR

proddy:
did some more cleaning, saved some ram. Think we need to push this PR first, and then your S2/C3 one as not to have too many merge conflicts

PR #677, opened 2022-10-10
C3/S2 support, tags translated

MichaelDvP:
I've already merged your PR (with actual NO translation). So it looks like more changes as it is from me.


src/devices/thermostat.cpp
```diff
@@ -593,12 +593,12 @@ void Thermostat::process_RC20Timer(std::shared_ptr<const Telegram> telegram) {
         uint8_t time = telegram->message_data[1];
 
         // we use EN settings for the day abbreviation
-        std::string sday = (FL_(enum_dayOfWeek)[day][0]);
```
proddy, line 596: I've been doing the same, everytime there's a FL_ try and use const char *. It really releases a lot of heap memory. My free available went from 140kb to 153kb with a few changes I made over the weekend.

proddy (approved):

proddy:
@MichaelDvP I see you're still making changes. Let me know when it's ok to merge

MichaelDvP:
It's now only very small things i realize when testing. You can merge if you are ok with the changes. S2 works good, i'm now on C3, seems stable too, Last change was playing around with color led.

PR #679, opened 2022-10-11
Updated German translation

tefracky:
Fixed some typos and language clearificaiton


src/locale_translations.h
```diff
@@ -350,7 +350,7 @@ MAKE_PSTR_LIST(releaseWait, "releasewait", "boiler release wait time", "Wartezei
 // the following are dhw for the boiler and automatically tagged with 'dhw'
 MAKE_PSTR_LIST(wwSelTempLow, "wwseltemplow", "selected lower temperature", "untere Solltemperatur", "Onderste streeftemperatuur", "Vald lägstatemperatur")
 MAKE_PSTR_LIST(wwSelTempOff, "wwseltempoff", "selected temperature for off", "Solltemperatur bei AUS", "Streeftemperatuur bij UIT", "Vald tempereatur för AV")
-MAKE_PSTR_LIST(wwSelTempSingle, "wwseltempsingle", "single charge temperature", "Solltemperatur Einmalladung", "Streeftemperatuur enkele lading", "Temperatur Engångsladdning")
+MAKE_PSTR_LIST(wwSelTempSingle, "wwseltempsingle", "single charge temperature", "Ladetemperatur", "Streeftemperatuur enkele lading", "Temperatur Engångsladdning")
```
MichaelDvP, line 353: As far as i understood this setting is not for normal charging, only for a single charge triggered by a button. It's RC300 and i don't have one, so i'm not sure about that.

MichaelDvP (approved):

src/locale_translations.h
tefracky: What does you mean with "normal charging"? The reason for this change was, that is also named "Desinfektionstemperatur" and not "Solltemperatur Desinfektion". It is shorter and more comparable to the other strings, but the meaning is the same. But as I already said, for me it is not clearly, what is the difference between "normal charging" and "single charge", so maybe I am wrong with this translation.

MichaelDvP: As far as i understand: wwSelTemp is the charging temperature used for charging in comfort mode, wwSelTempLow in eco/night mode, wwSelTempOff is something like frost protection if dhw is switched off, wwSelTempSingle is only used if someone presses the charge button on the thermostat.

tefracky: Okay, than my translation/change is missunderstanding and should not be used.

proddy:
merged

PR #683, opened 2022-10-13
Added Polish names of the entities

pswid:
not 100% complete


pswid:
now it should be OK

proddy:
thanks!

pswid:
Not at all. I am happy to help (as much as I can) because EMS-ESP is great and very usefull.
Unfortunately, Polish, as they say, is one of the most complicated languages ​​in the world and it is practically impossible to translate into Polish correctly without knowing the context. Most of the parameters that I have translated so far appear in my Buderus heat pump and I tried to translate them as best as possible based on the messages displayed by the pump controller or from the Polish manual of the pump. I will try to translate something more into Polish in the future.
However, there are two global problems with getting a correct Polish translation.
1. In the translation of the web interface you are breaking English sentences into words, the words are then translated into a specific language and already translated are "put together" back. It's probably ok for German but for Polish it will almost never give you a correct sentence! The order and "variant" of words usually also changes. Example: "Network" -> "Sieć", "Settings" -> "Ustawienia", but "Network Settings" != "Sieć Ustawienia" but "Ustawienia sieci" (word order changed, no capital letter in middle and one word changed)! The only 100% solution is to translate whole sentences (like for entity names). 
2. The translations of the values' names ​​also differ depending on the name of the entity. In English, for example, there is a simple OFF, but in Polish, one of the following may be correct: WŁĄCZ, WŁĄCZONO, WŁĄCZONE, WŁĄCZONA, WŁĄCZONY and many others. It would be great if few alternative translations (even only for Polish and only nesessary ones) could be added to the code. Do you think it's possible?
For cases with up to three variants of the word, PLURAL syntax: {{zero | one | other}} seems to be enough. Do you want to see my attempt? I can prepare test PR...

Issue #654, opened 2022-10-01
Temperature artifacts when controlling current flow temperature via MQTT


Lemonn:
I use an external program to control the current flow temperature of the boiler via MQTT. EMS-ESP32 [v3.4.2] and my program both use MQTT QoS 0. Sometimes when the temperature changes from a period of x°C to y°C, I get artifacts where the temp jumps back for a short period to the previous value. Resulting in unwanted burner starts/stops. I've verified that my program is not sending these values thru monitoring the corresponding MQTT topic. I've also ruled out warmwater or warmwater refreshs as cause.

<details>
  <summary>Log Data</summary>

  ```
 map[client:192.168.178.178:60575 content:1 2022-10-01T19:29:24.843140+2:00 ems-esp telegram - - - 001+00:00:27.827 T 110145: Rx: 88 00 34 00 28 01 B5 80 00 A1 00 00 01 00 01 DE 04 02 59 41 00 FB facility:14 hostname:192.168.178.178 priority:119 severity:7 tag: timestamp:2022-10-01 19:29:25 +0200 CEST tls_peer:]
map[client:192.168.178.178:60575 content:1 2022-10-01T19:29:24.861726+2:00 ems-esp mqtt - - - 001+00:00:27.842 D 110146: Publishing topic ems-esp/boiler_data_ww (#62562, retain=0, retry=1, size=484, pid=1) facility:3 hostname:192.168.178.178 priority:31 severity:7 tag: timestamp:2022-10-01 19:29:25 +0200 CEST tls_peer:]
map[client:192.168.178.178:60575 content:1 2022-10-01T19:29:24.870833+2:00 ems-esp command - - - 001+00:00:27.853 D 110147: Calling command 'boiler/selflowtemp' (selected flow temperature) with value 0 facility:3 hostname:192.168.178.178 priority:31 severity:7 tag: timestamp:2022-10-01 19:29:25 +0200 CEST tls_peer:]
map[client:192.168.178.178:60575 content:1 2022-10-01T19:29:24.960940+2:00 ems-esp mqtt - - - 001+00:00:27.943 D 110148: Publishing topic ems-esp/boiler_data (#62563, retain=0, retry=1, size=613, pid=1) facility:3 hostname:192.168.178.178 priority:31 severity:7 tag: timestamp:2022-10-01 19:29:25 +0200 CEST tls_peer:]
[...]

PR #688, opened 2022-10-16
fix: Bugs occurring whit fast sending clients

Lemonn:
Fixes https://github.com/emsesp/EMS-ESP32/issues/654

### DIFF
--- a/src/emsesp.cpp
+++ b/src/emsesp.cpp
@@ -1137,7 +1137,7 @@ void EMSESP::send_write_request(const uint16_t type_id,
                                 uint8_t *      message_data,
                                 const uint8_t  message_length,
                                 const uint16_t validate_typeid) {
-    txservice_.add(Telegram::Operation::TX_WRITE, dest, type_id, offset, message_data, message_length, validate_typeid, true);
+    txservice_.add(Telegram::Operation::TX_WRITE, dest, type_id, offset, message_data, message_length, validate_typeid, false);
 }
 
 void EMSESP::send_write_request(const uint16_t type_id, const uint8_t dest, const uint8_t offset, const uint8_t value) {
--- a/src/telegram.cpp
+++ b/src/telegram.cpp
@@ -431,58 +431,60 @@ void TxService::add(const Telegram::Operation  operation,
                     uint8_t *      message_data,
                     const uint8_t  message_length,
                     const uint16_t            validate_id,
-                    const bool     prioritize_message) {
-    auto telegram = std::make_shared<Telegram>(operation, ems_bus_id(), dest, type_id, offset, message_data, message_length);
-
-#ifdef EMSESP_DEBUG
-    LOG_DEBUG(F("[DEBUG] New Tx [#%d] telegram, length %d"), tx_telegram_id_, message_length);
-#endif
-    //If the queue is empty, just append the telegram
+                    bool     system_message) {
+    //TODO set the system_message on each call accordingly and then remove this!
+    //Right now every read operation is a system generated message
+    if (operation == Telegram::Operation::TX_READ) {
+        system_message = true;
+    }
+    auto new_telegram = std::make_shared<Telegram>(operation, ems_bus_id(), dest, type_id, offset, message_data, message_length);
+    tx_telegram_id_++;
+    //If the queue is empty, just append the telegram.
     if (tx_telegrams_.empty()) {
-        tx_telegrams_.emplace_back(tx_telegram_id_, std::move(telegram), false, validate_id);
+        tx_telegrams_.emplace_back(tx_telegram_id_, std::move(new_telegram), false, validate_id, system_message);
         return ;
     }
-    auto latest_telegram = tx_telegrams_.back().telegram_;
-    if (validate_id != 0) {
-        EMSESP::wait_validate(validate_id);
-#ifdef EMSESP_DEBUG
-        LOG_DEBUG(F("[DEBUG] Old type_id %d : New type_id %d "), latest_telegram->type_id, telegram->type_id);
-        LOG_DEBUG(F("[DEBUG] Old latest_telegram %d : New latest_telegram %d "), latest_telegram->offset, telegram->offset);
-#endif
-        //Replace previous write if the newly created telegram wants to write the same value
-        if (latest_telegram->type_id == telegram->type_id && latest_telegram->offset == telegram->offset && !prioritize_message) {
-            LOG_DEBUG(F("[DEBUG] Replaced telegram [#%d] whit updated version, because it sets the same value"), tx_telegrams_.front().id_);
-            tx_telegrams_.pop_back();
-        }
-        else if (tx_telegrams_.size() >= MAX_TX_TELEGRAMS) {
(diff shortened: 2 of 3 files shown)


MichaelDvP (commented):
I think i know what you want to do, but this way is wrong and your code does not do this.
With this code the queue is not really used, only the first fech-telegram is sent, all others ignored. 
The fetched telegrams are redundant, if we skip one, it will be send a minute later, no hurt. But writes are only received once from api/mqtt and have to send out with high priority.

Your issue is a wrong configured mqtt server-connection, We should fight this issue at it's root. 
If we want to do this in ems-esp my suggestion is: remember the last published mqtt topic/payload and if publish_topic2cmd is configured and the identical message is received as echo, skip it.

Lemonn:
I see. The mistake is that if a read operation follows a read operation, an unwanted overwrite happens. I will fix this so that reads get prioritized and never overwrite each other. 

The main incentive was to fix the reordering of write requests whit a fast-sending sender. I also contained queue pollution, by replacing duplicate writes of the same value whit the updated version.

This fix is not perfect. Because when someone sends a=1, a=2, a=3, everything is replaced as intended. But when someone sends a, b, a, b ... the queue is still filled whit repeating commands. I would like to fix this as well. But the queue and its elements are static, so the options would be to allow changes in the queue or to rebuild it on each change. What do you think is the better option?

I never had an issue whit my MQTT-Server. The messages originate from my application, as stated in the linked Issue.

Lemonn:
Is there a reason why reading operations should be prioritized over previous writing operations? I would suggest prioritizing system-generated messages over user-generated messages.

For user-generated messages, I would not prioritize between reading and writing operations. Instead, I would hint a full queue and the resulting message deletion to the user in the log.

MichaelDvP:
> I never had an issue whit my MQTT-Server. The messages originate from my application, as stated in the linked Issue.

Oh, yes i remember. Why don't you change your application?
> Is there a reason why reading operations should be prioritized over previous writing operations? 

? You misunderstood the code. Writing and verifying is priorized to give user-settings a fast response. And again: a one time sent writing message should always be executed, a redundant read can be droped or delayed.

Lemonn:
> You misunderstood the code. Writing and verifying is priorized to give user-settings a fast response. And again: a one time sent writing message should always be executed, a redundant read can be droped or delayed.

The provided change never drops writing operations, except when the queue is full. The Code does exactly what you describe as intended behavior. In case of a write operation and a corresponding automatically generated read, the read is prioritized. The only mistake I could see is that if two read requests are generated the second one overwrites the first one. I run this fix for 24H on my system and it works just fine.

> Oh, yes i remember. Why don't you change your application?

I already explained why it's better to fix the reordering issue on the side of EMS-ESP32. Let me try it again. The reason is if someone writes for example a PID controller that reacts to updates in values. It's confusing and complex to limit the number of messages sent by the client to prevent message reordering. It's better if EMS-ESP32 drops messages that it couldn't send. EMS-ESP32 does this but without preserving the order. This was fixed in this pull request. In addition to this, the pull request also limits the number of messages that need to be sent. It does this by replacing messages in the queue that set the same value, whit the updated version.

Lemonn:
Don't look into this version anymore. I will update it whit an improved version soon.

MichaelDvP:
Again you are confsing the prioritoies in a unwanted way. I'll try to explain again:
- Ems-esp fetches every minute telegrams that are not broadcasted, This is done [here](https://github.com/emsesp/EMS-ESP32/blob/8b10970e03123df80e688e80a6096c6636757141/src/emsesp.cpp#L1327). For a RC300 with 4 heatingcircuits this queues 42 telegram-reads. The bus is slow, it takes ~25 sec to send all these messages out. This timing is done by busmaster, we can not change it. 
- If a user changes a value in web-ui or api/mqtt he wants to see that it changes and don't want for wait 30 sec for feedback. So these writes need priority and have to send next. Also the verify, i.e. the read to verify the change is set, needs priority to give the user feedback.

BTW: In discord users worry about sending 5 writes a day to the bus, and you try a hundred per minute. 

With the above logic your usecase leeds to a pingpong. Sending a write via mqtt generates the ems-write, next the verify read and a mqtt publish of the result. You take this publish as trigger to send again a message, that is immediate written and give next response, etc. At the time you send the message, the old write isn't anymore in the queue, there is nothing to replace. 
You should check before sending the mqtt write if a change is nessesary.

A small change to do in emsesp what should be done on your side is to add [here](https://github.com/emsesp/EMS-ESP32/blob/8b10970e03123df80e688e80a6096c6636757141/src/devices/boiler.cpp#L1760):
```
    if (v == selFlowTemp_) {
        return true; // nothing to change
    }
```
This will skip the unnessesary ems-writes and the verifys and the mqtt responses, and your logic is triggert only on change of boiler values. But boiler_data is send if any of the values changes, this is also very often and will still generate much useless messages on your side (but most of them with no change in selfowtemp and not generating a ems-write).

Lemonn:
Thanks for the explanation. It wasn't clear to me that certain setups produce so many read operations. Now I see the tradeoff. If we prioritize writes, the regularly performed reads could suffer from starvation. On the other hand, if we prioritize reads. Users have to wait for their changes to be performed. Up to the point where reading operations starve the writing operations. 

For sure we could fix all these problems and create a solution that works for both cases. For example, the user writes could be separated from automatic writes. Additionally, the automatically performed thermostat read could be disabled and only performed on request. But as you said most users won't run into this problem. Therofre I think the effort needed is not worth it. For my use case, this fix is perfect, so I think I will keep it separate. 


> With the above logic your usecase leeds to a pingpong. Sending a write via mqtt generates the ems-write, next the verify read and a mqtt publish of the result. You take this publish as trigger to send again a message, that is immediate written and give next response, etc. At the time you send the message, the old write isn't anymore in the queue, there is nothing to replace.
> You should check before sending the mqtt write if a change is nessesary.

I understand that this creates a ping-pong effect. Have a look at the following picture. The whole graph spans 15min. The steep rising happens in under 45 seconds. Even if I filter values and only respond to changes, the number of changes still will be huge.
image

This happens if you need heating of 1-2,5Kw/h and the installed boiler has 24Kw/h whit a max modulation down to 45%. Combined whit no Room temperature sensors attached, which only allows setting the flow temp whiteout knowing the demand.

This, combined whit a boiler which does not allow setting the modulation value. Results in the boiler always start at 65+% of its max power, which then results in burner cycling, because of overshoot (600000 Heating burner starts during 300 days of burner runtime). This is PWM modulation at its finest. You could even hear the water cooking if the burner turns on without heating demand.

I tried to fix this whit EMS-ESP by using the outside temperature + some factors for the isolation and space to determine an estimated heating load. I then try to run the burner as long as possible, in a row, by forcing it to modulate down to 45%.

Therefore I tried to set the flow temp always a few degrees under the actual flow temp, which then results in a slow downwards modulation without cycling. The modulation still needs over 5 mins from 65% to 45%. I don't know why this is the case during heating, because if warm water is requested, the boiler modulates in seconds from 80% to 45%. Let me know If you have a better solution to force the burner into modulation.

PR #693, opened 2022-10-17
Polish translation improvements

pswid:
Strange, but if i use capital letters in Polish translation for "ON' and "OFF" (`local_translations.h`), I'm getting compile error. So instead of whole words in capital letters I used abbreviations in lower case (more useful, IMHO).
To hold whole word "wyłączono" (="off") I also increased everywhere(?) buffer size for boolean conversions to 12 bytes. Please adjust it if it's not ok.


proddy:
this looks ok to merge. @MichaelDvP ?

proddy:
> change e.g. line 99 to the: ` MAKE_PSTR_LIST(OFF, "OFF", "AUS", "UIT", "AV", "WYŁ.")` and you'll see. I'm getting several errors: `lib/framework/../../src/telegram.h:32:10: fatal error: uuid/log.h: No such file or directory` full console output: [log.txt](https://github.com/emsesp/EMS-ESP32/files/9803567/log.txt)

ah, it's probably because of the Macros. I haven't tried but try backslashing the dot, so `\.`? I'll do some experiments later this week too

proddy:
> `MAKE_PSTR_LIST(OFF, "OFF", "AUS", "UIT", "AV", "WYŁ.")`

I'm not seeing that error if I insert that line. I also noticed you're compiling with older libraries. Make sure you have the latest espressif arduino core by doing a `pio pkg update`. From VSCode you'll need to first use the PlatformIO extension and choose 'Miscellaneous/PlatformIO Core CLI' to open the terminal. I know, it's stupid.

interface/src/i18n/se/index.ts
```diff
@@ -174,6 +174,10 @@ const se: BaseTranslation = {
   SYSTEM_FACTORY_TEXT_DIALOG: 'Är du säker att du vill fabriksåterställa enheten?',
   VERSION_CHECK: 'Kontrollera version',
   THE_LATEST: 'Den senaste',
+  OFFICIAL: 'official',
+  DEVELOPMENT: 'development',
+  /**VERSION_IS: 'version is',*/
```
MichaelDvP, line 179: Afaik the language fallback of i18n is on file level, not on word level. We need all words defined or it will be empty. Better use the english text and add comment // translation needed! 
In SE there is also RESTART_CONFIRM from earlier missing. The different line nubers of the index files are irritating, maybe use the same formatting for each index.ts to see differences better.

MichaelDvP (approved):
Looks good, i've added a comment about index files, but we can do this later.

MichaelDvP:
With c6p this line `MAKE_PSTR_LIST(OFF, "OFF", "AUS", "UIT", "AV", "WYŁ.")` i can reproduce the issue. 
Using PLATFORM: Espressif 32 (5.2.0+sha.03254d4) > WEMOS LOLIN C3 Mini
and also for ESP32. 
Something in charset? 
Modifying the working lower case to `MAKE_PSTR_LIST(OFF, "OFF", "AUS", "UIT", "AV", "Wył.")` works.

pswid:
Same with me. I still have a compile error after upgrading platform to 5.2.0. It is evidently caused by a capital letter **Ł** (\u0081). Weird.

interface/src/i18n/se/index.ts
pswid: I noticed that there is no translation for VERSION_IS in SE. Not knowing if it was by mistake or on purpose, I put the entire line from EN as a comment

pswid: So why not to do something like this: https://github.com/ivanhofer/typesafe-i18n#i-have-two-similar-locales-only-a-few-translations-are-different-but-i-dont-want-to-duplicate-my-translations ?
Every missed/forgotten word missed would be automatically replaced by English one. We could even remove (or comment out) words that we doesn't want to be translated...

MichaelDvP:
> So why not to do something like this: https://github.com/ivanhofer/typesafe-i18n#i-have-two-similar-locales-only-a-few-translations-are-different-but-i-dont-want-to-duplicate-my-translations ?

Nice, i tried it, but it gave me for web: 
```
File sizes after gzip:
  278.78 kB (+34 B)   build\js\179.06b8.js
  8.04 kB (+3.05 kB)  build\js\99.a1be.js
  7.52 kB (+2.84 kB)  build\js\189.1b22.js
  7.16 kB (+2.77 kB)  build\js\981.77a6.js
  6.88 kB (+2.44 kB)  build\js\865.cdba.js
  4.3 kB (-56 B)      build\js\170.a0ee.js
  4.25 kB             build\js\528.69d9.js
```
not good for my C3, wihch is already without this at:
grafik

We should only use it if a lot of words missing.

pswid:
Pity. Though maybe there is any other way to achieve this without losing memory?

I also noticed that all language files (except DE) use `BaseTraslation` except `Translation` but not being sure if it was done on purpose or by mistake, I didn't change it.

MichaelDvP:
> I also noticed that all language files (except DE) use BaseTraslation except Translation

My fault, copying the en index to other languages, should be `Translation`. Please change.

proddy:
> My fault, copying the en index to other languages, should be `Translation`. Please change.

the doc in https://emsesp.github.io/docs/#/Coding?id=adding-a-new-language needs to be updated then (it says not copy the whole index.ts file)

proddy:
thanks all

PR #702, opened 2022-10-25
The translation of the web interface into Polish has been corrected and completed (e.g. the word order has been fixed).

pswid:
Newly added words should be translated from English into other languages in `index.ts` files


proddy:
I'll review, thanks. looks like a lot of work!

pswid:
I noticed that Polish chars look a bit different than others (they're bigger and thinner). It seems that font you're using doesn't contain Polish chars and they are being automatically substituted by chars taken from the different font (does the browser do this?). The difference is not big but noticable especially the capital chars (compare Ą to A and Ż to Z). Could you "fix" the font?

image

MichaelDvP:
[https://github.com/emsesp/EMS-ESP32/blob/c3cfed5ac36f728d65917b8ac146b93ecfe52739/interface/public/css/roboto.css#L2](https://github.com/emsesp/EMS-ESP32/blob/c3cfed5ac36f728d65917b8ac146b93ecfe52739/interface/public/css/roboto.css#L2). A full roboto is 64kB each, latin only 16kB, Maybe you can use a font editor and add only the few characters you need?

MichaelDvP (approved):
I'm not able to check pl translations, only checked that en/de is not broken.

pswid:
> https://github.com/emsesp/EMS-ESP32/blob/c3cfed5ac36f728d65917b8ac146b93ecfe52739/interface/public/css/roboto.css#L2
> 
> . A full roboto is 64kB each, latin only 16kB, Maybe you can use a font editor and add only the few characters you need?

I don't know how to do it. Sorry.
There are only 9 special Polish letters, so 18 chars in total are missing (small and capital letters). See the whole list e.g. [here](https://www.learnpolishtoday.com/mini-polish-tips/unicode-codehtml-code-for-polish-letters-and-symbols). Could you extract missing Polish characters from full roboto font and add them to the code, please?

PR #713, opened 2022-10-31
min/max, bus-ids, MD5

MichaelDvP:
I've resorted to thematic commits with only a few changes.


interface/.typesafe-i18n.json
```diff
@@ -1,5 +1,5 @@
 {
   "adapter": "react",
   "baseLocale": "pl",
```
proddy, line 3: the baseLocale should be 'en', but this will only work when the rest of the index.ts translations file use the same format as was done for Polish with the none|single|many syntax

MichaelDvP: This was already in dev, i did not change it.

PR #719, opened 2022-11-03
fix to the translation of the webUI in Polish.

pswid:
Everything visible should already be translated into Polish. However, I was not able to check how the translations of all errors are displayed (how to display them without real error?), so I'm not sure that they are 100% ok. 
BTW, there are still many error messages hardcoded (e.g. "Local IP is required"). I think we should translate them as well.


proddy:
I've made a note about the 'local IP is required' to go back and fix the missing translations

PR #722, opened 2022-11-05
Changes in discussion

MichaelDvP:
Don't know if the merged pl/no should go to vikingn first, there was no feedback.
This also contains my fix for baseTranslation (i know youre planing a different change, but for nw it works this way), the discussed board profile fallback to E32/S32, and the restart to factory partition (if exists).
Each change is in an extra commit,


proddy:
I just made a fix to the translations, I'll see if we had the same approach

PR #723, opened 2022-11-05
fix to base translations, auto generate i18n

proddy:


proddy:
@MichaelDvP I took a similar approach and made the default translation polish so the fallback works and you don't need to use {{ }} in the non-polish translations. Seems to work.

MichaelDvP (approved):
with baseLocal pl other languages don't need the cases.
Please merge, than i'll update my PR to avoid merge conflicts.

proddy:
ok, merged.

PR #733, opened 2022-11-07
fix: cosmetic webUI improvements

pswid:
mainly in custom board profile and analog sensors settings. Now it looks more consistent with other pages.


proddy:
I'll take a look when I'm back (traveling this week). One thing to keep in mind is that it needs to render and be usable on both a mobile phone in landscape and portrait.

interface/src/framework/mqtt/MqttSettingsForm.tsx
```diff
@@ -219,15 +219,15 @@ const MqttSettingsForm: FC = () => {
...
+              <Grid item xs={2}>
                 <ValidatedTextField
                   name="discovery_prefix"
                   label={LL.MQTT_PUBLISH_TEXT_4()}
                   fullWidth
                   variant="outlined"
                   value={data.discovery_prefix}
                   onChange={updateFormValue}
-                  margin="normal"
+                  size="small"
```
proddy, line 230: making it small is not consistent with the other sizes, it also renders weird and the text is cutoff in English

interface/src/project/SettingsApplication.tsx
```diff
@@ -404,7 +405,7 @@ const SettingsApplication: FC = () => {
           />
           {data.shower_alert && (
             <>
-              <Grid item xs={4}>
```
proddy, line 407: ok to have larger boxes if you prefer, but the spacing between the two is large and it looks funny on a desktop browser

proddy (changes requested):
I reviewed this PR. I like the dialog boxes and settings improvements. Two comments
 - the shower timer & duration spacing is off
 - the MQTT HA discovery name is too small and the text is cut off

I've made individual comments

interface/src/framework/mqtt/MqttSettingsForm.tsx
pswid: IMHO, a `size="small"` style rectangle looks better than a normal (large) one when is displayed in the line with the option name. Don't you think? Also, the same style was already used in the entity language option before.

interface/src/project/SettingsApplication.tsx
pswid: my new fixes should solve the text clipping issues

Issue #748, opened 2022-11-15
Compressor Activity is empty during defrost run


HansRemmerswaal:
The compressor activity normally has the values none, heating, hot water, (and I guess cooling).
But I start to see examples where the compressor activity is empty. When I check it looks like it's always at the same time that there is a defrost run. For 2 min my circulation pump speed is becoming 100% and I see the temperature dropping.

Is this value retrieved from the EMS bus, so a software issue in the firmware of my Nefit heat pump? Or is this a values determined in the EMS-ESP32 and needs to be reported as an issue?

![image](https://user-images.githubusercontent.com/116959990/201850901-173d4d91-3bee-4627-9a3d-2f52a3a606f5.png)

![Capture](https://user-images.githubusercontent.com/116959990/201850240-c82dc87a-f404-465a-a59b-7732f8931c64.JPG)

Devices I use:
BBQKees Gateway E32 with firmware 3.5
Nefit Eviline A/W Split 7.0 E-S

MichaelDvP:
I think this is a 6th option in hpactivity. If it happens again try in terminal `read 8 48D` or set in web system/log to all, wait until message 48D is received, save the log, post it here.
You can also test https://github.com/MichaelDvP/EMS-ESP32/releases/tag/latest, there i have added a option to hpactivity (5=noforst), but if the value goes to 6, 7, .. you still get empty text.

HansRemmerswaal:
I have set in web system/log to all. Will post the log as soon as I see the issue again.

HansRemmerswaal:
That system log is not very big, only +/- 200 lines. So I need to react within minutes when it happens...? I put the buffer size to max.

MichaelDvP:
As long as you leave the log page open the log is only limited by the memory the browser can allocate. Buffer is emsesp internal and only used if you switch to a different emsesp-page, this is limited to max. 100 lines.
But it's easier to use syslog or telnet with logging to file (like putty) and in telnet commands `set timeout 0` and `watch 8 48d`
And/or check with my lates commit metioned above if your compressor activity log shows empty or "defrost".

HansRemmerswaal:
Yesterday I have installed v3.5.0b9, today I had the issue again. See screenshots. So the latest development version didn't solve the problem.

[...]

PR #753, opened 2022-11-18
fixes #747, #748

### LINKED ISSUE #747: Weird things since upgrade to 3.4.4
[Sorry - don't get the issue template and can't assign any tags, for some reason. Hopefully someone can add for me please?]

I'm seeing two weird things since I upgraded to 3.4.4...

- NTP is disconnecting and reconnecting regularly, shown both in the system log and in HA
image
image
image
The gap in the bottom image on Saturday is when I did the upgrade, so the regular disconnects are only happening since then.

- Two new entities have appeared in HA...
image
image
This is curious, as I don't have a heat pump! There are no MQTT Explorer entries with the string 'heatpump' and no entities in HA with the string 'heatpump' either. So where the heck are these coming from? The logs include...
```
2022-11-14 07:51:37.758 WARNING (MainThread) [homeassistant.helpers.template] Template variable warning: 'dict object' has no attribute 'summermode' when rendering '{{value_json.hc1.summermode}}'
2022-11-14 08:00:19.147 WARNING (MainThread) [homeassistant.helpers.template] Template variable warning: 'dict object' has no attribute 'hpoperatingstate' when rendering '{{value_json.hc1.hpoperatingstate}}'
2022-11-14 08:05:48.332 WARNING (MainThread) [homeassistant.helpers.template] Template variable warning: 'dict object' has no attribute 'summermode' when rendering '{{value_json.hc1.summermode}}'
2022-11-14 08:16:11.731 WARNING (MainThread) [homeassistant.helpers.template] Template variable warning: 'dict object' has no attribute 'hpoperatingstate' when rendering '{{value_json.hc1.hpoperatingstate}}'
```

MichaelDvP:
- for #747 thetranslations are from google-translate, maybe needs rework from native speakers
- #748 is only the thermostat check for summer2_typeids, no feedback yet, but it should work
- i've forgot to add the "has_loader" in https://github.com/emsesp/EMS-ESP32/commit/1d4634a76cf0faa5a3cad0460b1a3311b0a45dd1
- changed the weblog to deque to free memory if buffersize is reduced. Now i check heap on every logmessage and size 20k. Have not tested if this is a good value. It's the size of the largest json buffer we use (EMSESP_JSON_SIZE_XXXLARGE_DYN)


proddy:
> * changed the weblog to deque to free memory if buffersize is reduced. Now i check heap on every logmessage and size 20k. Have not tested if this is a good value. It's the size of the largest json buffer we use (EMSESP_JSON_SIZE_XXXLARGE_DYN)

Another thing on my list, not sure if mentioned it before, was to use https://github.com/muwerk/ustd to replace map, queues and lists eventually. It's light-weight and I used it in ESP v2 on the ESP8266, also contributed to the library back then. I think it should save more memory if heap becomes a problem. I just need to run some experiments as I'm afraid of breaking something.

src/web/WebLogService.cpp
```diff
@@ -110,6 +110,10 @@ WebLogService::QueuedLogMessage::QueuedLogMessage(unsigned long id, std::shared_
 }
 
 void WebLogService::operator<<(std::shared_ptr<uuid::log::Message> message) {
+    if (ESP.getMaxAllocHeap() < 20480) {
```
proddy, line 113: i was was always worried these ESP.() functions are expensive and sometimes blocking, but if it works its good!

MichaelDvP:
Free heap is already a problem for a few people with E32 and lot of entities from heatpump and RC300 with multiple hcs.
Typical issue is a white page for these pages that uses the 20k json. (also systeminfo only reports "NULL").

proddy:
> Free heap is already a problem for a few people with E32 and lot of entities from heatpump and RC300 with multiple hcs. Typical issue is a white page for these pages that uses the 20k json. (also systeminfo only reports "NULL").

is it just the web UI that is impacted or does EMS-ESP crash and reboot? If it's just the web we can look at streaming the data on the Dashboard

MichaelDvP:
I think only blank pages on web ui. One was in #748 after setting logbuffer to max. Another was on discord [here](https://discord.com/channels/816637840644505620/816958087160266775/1040371629764657173).

Testing the S2 with high logbuffer and growing mqtt queue also does not crash, but ems-esp was not reachable over wifi anymore (web and mqtt). With S32/E32 and c3 i always have enough memory in my system.

Issue #759, opened 2022-11-20
Multiple EMS Devices share same id when using HA & MQTT


Rainson12:
*Before creating a new issue please check that you have:*

**Bug description**
We operate 2 EMS-ESP-32 systems for 2 houses conntected to the same MQTT broker. The heating systems connected to the ems-esps are the same model. I have configured unique network hostnames and unique client ids for both ems-esps. However when running HA & MQTT, home assistant only detects one of the ems-esps and it seems like all entities are overwritten by each other meaning if I add the detected entites to the overview, the displayed values keep switching between the first and seconds ems system (overwriting each others values).

**Steps to reproduce**
Connect two EMS-ESP-32 to the same heating system model, set different hostnames and clientids and run Home Assistant and enable auto discovery. 

**Expected behavior**
The detected devices and entities have unique ids.

**Screenshots**
EMS-ESP #1
![image](https://user-images.githubusercontent.com/13119203/202903535-ed685271-59db-4aec-932c-3d1b1c09794c.png)
![image](https://user-images.githubusercontent.com/13119203/202903557-a8668b6e-9a18-4208-acb9-07e3679428b7.png)

EMS-ESP #2 
![image](https://user-images.githubusercontent.com/13119203/202903573-4179c474-7a98-420f-b589-ed8e6cb69e88.png)
![image](https://user-images.githubusercontent.com/13119203/202903582-1039f7ea-beed-4ef7-95ed-d647bee2fc62.png)

HA logs:
![image](https://user-images.githubusercontent.com/13119203/202903606-4731a4f0-e2ab-47c9-9726-3017a3041cff.png)

HA MQTT detected devices:
![image](https://user-images.githubusercontent.com/13119203/202903636-f82bfff3-42b1-46e7-be16-a368df3d1768.png)

**Device information**
Device #1
```
{
  "System Status": {
    "version": "3.5.0b4",
    "uptime": "000+00:19:32.997",
    "freemem": 117,
    "free_app": 115,
    "reset reason": "Software reset CPU / Software reset CPU"
  },
  "Network Status": {
    "connection": "WiFi",
    "hostname": "hausa",
    "RSSI": -72,
    "IPv4 address": "192.168.100.20/255.255.255.0",
    "IPv4 gateway": "192.168.100.1",
    "IPv4 nameserver": "192.168.100.1",
    "static ip config": true,
    "enable IPv6": false,
    "low bandwidth": false,
    "disable sleep": true,
[...]

Issue #758, opened 2022-11-20
Entity id for sensors within HA changes when updating name in web-ui 


tp1de:
I recognized that while changing names for sensors (analog & dallas) within web-ui the entity id changes as well for HA.
I think only full name should change. Entity id should be kept.

proddy:
I did a quick test with Analog sensors, and you're correct, when you create one (say GPIO 31, name is Analog1) it will create the HA entity called `sensor.analog_sensor_analog1`. I think it's best to use the GPIO and not the name, so `sensor.analog_sensor_31`

tp1de:
I think it is important to keep the unique id in ha when updating names within ems-esp to keep history in statistics as well the dashboard configuration. I haven't recognized issues with dashes within unique id. I would keep it. For analog sensor I would recommend to add gpio to the name ... e.g.  sensor.analog_gpio_31  .... two times "sensor" ist not needed.

tp1de:
... and for dallas the "value" in api seems to be rounded to full integer.

proddy:
> ... and for dallas the "value" in api seems to be rounded to full integer.

can you create a new issue for this please. It's a new bug, and not related to HA

tp1de:
... the whole issue is related to ha as well to api. Api responses for dallas and analog sensors are different anyhow.

MichaelDvP:
Yes dallas and anlog api is not in line with devices. 
I suggest to use the info/values command only for a list
```
call dallassensor values
{
  "cylinder_top": 45.3,
  "solar_flow": 23.2,
  "cylinder_middle": 33.9,
  "cylinder_bottom": 21.8,
  "circ_backflow": 34.1,
  "solar_backflow": 22.6,
  "boiler_backflow": 40.6
}
```
and add the value info with name,
```
call dallassensor cylinder_top
{
  "id": "28-7284-131F-1901",
  "name": "cylinder_top",
  "value": 45.3,
  "type": "number",
  "uom": "°C",
  "writeable": false
}
```
For HA the info should change the object_id to from name to id also matching the topic.
( actual:  "object_id": "temperature_sensor_cylinder_top",)
```
homeassistent/sensor/ems-esp/dallassensor_28-7284-131F-1901:
{
  "dev_cla": "temperature",
  "stat_t": "ems/dallassensor_data",
  "unit_of_meas": "°C",
  "val_tpl": "{{value_json['cylinder_top']}}",
[...]

PR #764, opened 2022-11-21
Command `values` for analog/dallas sensors, writeable for analog-out,…

MichaelDvP:
These are the changes for analog/dallas i mentioned in #758. 
/api/dallassensor now defaults to "values" and reponse a list of values, i left the info command untouched.
same for analogsensor, the outputs generate a command to name. Because these names are not lowercase the namecheck has to compare incoming and stored commandname in lowercase.
value-info for analog/dallas works with name and id, writeable is included and min/max only for commands, as for ems-entities.
Changing analog sensors removes/regenerates the commands.
I've changed the HA entity for dallas/analog to ID. 

Adding the basename to entity for #759 is your turn, you can better check and it's a breaking change for HA users.

@tp1de, i've modified your adapter to test the changes, check https://github.com/MichaelDvP/ioBroker.ems-esp (I see you have also made some changes since i've pulled)

### DIFF
--- a/CHANGELOG_LATEST.md
+++ b/CHANGELOG_LATEST.md
@@ -18,6 +18,7 @@
 - Add MD5 check [#637](https://github.com/emsesp/EMS-ESP32/issues/637)
 - Add more bus-ids [#673](https://github.com/emsesp/EMS-ESP32/issues/673)
 - Use HA connectivity device class for Status [#751](https://github.com/emsesp/EMS-ESP32/issues/751)
+- Add commands for analog sensors outputs
 
 ## Fixed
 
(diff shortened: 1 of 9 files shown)


tp1de:
Michael, web-ui looks nice, but I am not happy about changes to api calls for .../api/dallassensor and .... /api/analogsensor
output structure changes. This is a breaking change for my ioBroker adapter where users use different ems-esp firmware versions. Why is this needed?

MichaelDvP:
@tp1de : I have not changed anything in web ui, only correct input field for DAC outputs.
The naming is discussed earlier. Yes: Counter and digital i/o are not analog and outputs are not really sensors. Correct name could be "gpio functions"  or "esp-perpheral-functions" but that is to technical for most people. Most requests for io functions are  ADC for pressure sensors and counter for gas/water flow meters, which are at least analog values.

proddy:
I'll check HA and will add the base when I'm back home this weekend

Issue #749, opened 2022-11-15
Add Command / Telegram to set current temperature of thermostat


JelleRoets-Oqton:
**Is your feature request related to a problem? Please describe.**
My setup consist of a Bosch Boiler Condens [7000iW](https://www.bosch-thermotechnology.com/be/nl/ocs/residencial/condens-7000iw-754481-p/), Bosch Thermostat [CW400](https://www.bosch-thermotechnology.com/be/nl/ocs/residencial/cw400--754545-p/) and a Bosch [MM200](https://www.bosch-thermotechnology.com/be/media/country_pool/professionals/our-projects/tender/nl/controller/firstspirit_1553669641165lastenboek_mcm_101_mm200_v2_nl.pdf) mixing module with 2 pumps. The main thermostat (CW400) is installed in the living and controlling the first circuit installed in the living room. 
However it is also controlling the 2nd circuit, which is connected to other different rooms in the house, but still using the measured living room temperature as current temperature, which does not make sense of course.

So to solve this, I installed extra temperature sensors in the other rooms and connected them to Home assistent.
With the E32 Gateway I can already overwrite the setpoint temperature for both the first and second circuit. However I cannot set the actual current room temperature of the 2nd circuit with the measured value of my temp sensors.
![image](https://user-images.githubusercontent.com/42942096/201931135-c3cb1e13-9366-4371-a042-8d6dd259011c.png)

**Describe the solution you'd like**
A command / telegram to set the current temperature of the 2nd circuit in the CW400 thermostat, if possible.

**Describe alternatives you've considered**
1. I could directly control the 2nd pump status from Home Assistent. However this way Home Assistent will act as a thermostat, while that is not the intention. Moreover, most likely the CW400 will still try control the 2nd circuit as well and probably overwrite my manual updates to the pump status. 
![image](https://user-images.githubusercontent.com/42942096/201931718-3e8bbf04-d10a-408a-bb1b-e4473023c800.png)
2. Buy an extra Bosch [CR10 thermostat](https://www.bosch-thermotechnology.com/be/nl/ocs/residencial/cr-10--754549-p/), that can act as a remote control for the main CW400. But actually this wouldn't work as there is no bus wire available in the other rooms and I would have to buy one for each extra room. So then I should buy the [EasyControl with CT200](https://www.bosch-thermotechnology.com/be/nl/ocs/residencial/easycontrol-set-ct-200-1136896-p/), which is way too expensive.

[...]

PR #780, opened 2022-11-29
allow raw telegram sending with other src-id

MichaelDvP:
Check if you think this is usefull. Maybe for testing a temperature telegram from external RF sensor. 
It allows for the `system send "<telegram>"` to use any src id. Works for me, but:
- My busmaster does not poll all ids, so i added a simple timeout to remove the telegram from queue if it is not polled within 500 master-polls. A timeout is not usefull, because some communication takes very long. 
- for reading/writing the communcation is not finalized with client poll, the master timeout ends the communication. (could be added, but i think it is not needed)
- I do not check for collision to recognized devices. For a few test commands this will not hurt, hopefully the (few) users flooding the bus with commands do not use this feature with wrong src-id.

### DIFF
--- a/.github/workflows/tagged_release.yml
+++ b/.github/workflows/tagged_release.yml
@@ -31,6 +31,7 @@ jobs:
           cd interface
           npm ci
           npx typesafe-i18n --no-watch
+          sed -i "s/= 'pl'/= 'en'/" ./src/i18n/i18n-util.ts
           npm run build
 
       - name: Build firmware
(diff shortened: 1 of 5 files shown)


proddy:
this would be a great addition. the poll timeout is useful. Did the comment from Discord about how to use MQTT to write values to EMS inspire this?

I'll check this weekend as I'm travelling the rest of the week.

MichaelDvP:
> Did the comment from Discord about how to use MQTT to write values to EMS inspire this?

No, the discussions how to use a external room sensor and send it to thermostat. #749. On discord there was a similar request for outdoor sensor, For room sensor we know that is it mounted on IVT, but seems RC300 compatible. There is a wireless base 0x50, with version, but not sending anything. The sensor (0x40) only publishes a single telegram and is not readable/writeable. 
Now the questions: Does RC300 accept the telegram from wireless sensor without a base recognized? The RC300 can't read the sensor, so it may work with a single telegram with src 0x40. People with RC300 can test without larger changes to software/commands.

proddy:
works well.

Issue #751, opened 2022-11-16
Change EMS-ESP device sensor types in Home Assistant


laszlojakab:
**Is your feature request related to a problem? Please describe.**
I'm using EMS-ESP32 with Home Assistant. My problem is that some of EMS-ESP device sensors have "wrong type". I know I could use template sensors to fix the problem but it would nice if EMS-ESP could register these sensors with the "correct type". The following sensors are problematic:

- `status`
- `ntp_status`
- `uptime`

**Describe the solution you'd like**
I think the following modifications needed:
- `status`: it should be registered as a `binary_sensor` with `connectivity` [device class](https://www.home-assistant.io/integrations/binary_sensor/#device-class) and not as a `sensor`. By doing this Home Assistant will localize the value to `connected` or `disconnected` or whatever based on the selected language. Also have to mention that Home Assistant dashboard can change color of binary sensor's icon natively based on it state. With the current string value `connected` it is only possible if I introduce a template binary sensor or if I'm using a third party ui component.
- `ntp_status`: same as `status` sensor.
- `uptime`: the natively supported [uptime sensor](https://www.home-assistant.io/integrations/uptime/) in Home Assistant returns the `timestamp` of booting in ISO 8601 format (eg: `2022-11-15 20:46:38.049117`). If NTP is enabled EMS-ESP could also provide this value. The [device class](https://www.home-assistant.io/integrations/sensor/#device-class) of the sensor should be `timestamp`. By this modification Home Assistant can natively convert the elapsed time to more easily readable text on the dashboards (for example: `2 days ago`). One more note: the devices in Home Assistant which reports uptime and has an RTC or NTP connection reports the booting time as uptime.

proddy:
Good points. Which version are you using?

laszlojakab:
v3.4.4

proddy:
regarding the 'uptime' ISO 8601 is not good for showing the elapsed time. It's better to add it to the MQTT info topic which looks like:

proddy:
and `status` is used as a last will message (https://www.home-assistant.io/integrations/mqtt/#birth-and-last-will-messages). Do you want to see if the EMS-ESP is online or just that it's connected to the EMS bus?

proddy:
I made the changes to status. For uptime, it needs more thinking and some help from your end

laszlojakab:
Thanks for the updates! What do you need from me about uptime?

proddy:
[...]

PR #781, opened 2022-11-29
#751 and #759

proddy:


MichaelDvP:
A lot of breaking changes for HA, but there are some thing to think about:
1. The boottime is now send on every ntp connect/disconnect, but on a wifi reconnect it is removed. This will give a HA complain. Better leave it with ntp_connected() and in NTPSettingService. Multiple sending does not hurt, the boottime is corrected by uptime and stays the same. If you want reduce to a single ntp-trigger you can use something like:
```
static bool firsttime= true;
if (firsttime) {
    emsesp::EMSESP::system_.send_info_mqtt("connected");
   firsttime = false;
}
```
2. The HA autoconfig is a bit mixed now. As far as i've seen you change a lot of things. The path for config does not contain the basename anymore, eg. `homeassistant/number/ems-esp/thermostat_hc2_daytemp/config` becomes `homeassistant/number/ems-esp_thermostat_hc2_daytemp/config`. The removal/cleanup by subscribing to `homeassistant/#/ems-esp/*` does not work anymore, You can't  use a wildcard subscription now without removing non-ems-esp configs.
3. The commands will not work, the basename is doubled in` "~/%s"` because "~" and uniq_id contains it. Also the `replace('_','/')` could make trouble if the shortname contains a '_' (not the case for now).
4. Now you have "uniq_id" and "object_id" the same with the shortname, good idea because the shortname always have to be unique for mqtt. The en-fullname is not used anymore and can be removed from this function.
5. In system-status, heartbeat, dallas, analog the "uniq_id" and "object_id" are not the same, this should be in line. Also the basename prefix is missing there, confusing, and give collision for multiple ems-esp in one system.

proddy:
Thanks for the review

> 1. The boottime is now send on every ntp connect/disconnect, but on a wifi reconnect it is removed. This will give a HA complain. Better leave it with ntp_connected() and in NTPSettingService. Multiple sending does not hurt, the boottime is corrected by uptime and stays the same. If you want reduce to a single ntp-trigger you can use something like:

I changed it, like you said no harm in sending every hour
 
> 2. The HA autoconfig is a bit mixed now. As far as i've seen you change a lot of things. The path for config does not contain the basename anymore, eg. `homeassistant/number/ems-esp/thermostat_hc2_daytemp/config` becomes `homeassistant/number/ems-esp_thermostat_hc2_daytemp/config`. The removal/cleanup by subscribing to `homeassistant/#/ems-esp/*` does not work anymore, You can't  use a wildcard subscription now without removing non-ems-esp configs.

I couldn't reproduce that. It worked fine in my setup.

> 3. The commands will not work, the basename is doubled in` "~/%s"` because "~" and uniq_id contains it. Also the `replace('_','/')` could make trouble if the shortname contains a '_' (not the case for now).

fixed that , thanks

> 4. Now you have "uniq_id" and "object_id" the same with the shortname, good idea because the shortname always have to be unique for mqtt. The en-fullname is not used anymore and can be removed from this function.

en-fullname removed, good catch.

> 5. In system-status, heartbeat, dallas, analog the "uniq_id" and "object_id" are not the same, this should be in line. Also the basename prefix is missing there, confusing, and give collision for multiple ems-esp in one system.

I've fixed that too.

Now just need some small tests and I'll check in the latest updates to this PR

proddy:
still not following you 100%

I made some changes, but haven't tested as I don't have HA right now (on a plane). From the output it looks ok and the folder structure is as you say (and has been since v3.0).

MichaelDvP:
> still not following you 100%

You're right, sorry, i dont know what i have seen, but now i've tested and it's ok.

A few last things:
- analogsensors: object_id and uniq_id should use gpio and `analogsensor`as in path (not `analog_sensor`)
- dallassensor: sync path/object_id/uniq_id to `dallassensor`  (not `temperature_sensor`)
- system_status: sync path/object_id/uniq_id to `system_status` (path is now only `system` and object_is is different to uniq_id)

And a question:
in dallassensor the sensorid in path is marked with `// use '_' as HA doesn't like '-' in the topic name`, but the basename is default `ems-esp` and seems to work in path? What is right? Has it changed with HA versions?
https://www.home-assistant.io/integrations/mqtt/#mqtt-discovery
says: 
```
<discovery_prefix>/<component>/[<node_id>/]<object_id>/config
<object_id>: The ID of the device. This is only to allow for separate topics for each device and is not used for the entity_id. The ID of the device must only consist of characters from the character class [a-zA-Z0-9_-] (alphanumerics, underscore and hyphen).
```

proddy:
@MichaelDvP I need your help. With this PR (version b11) the MQTT subscribe is not working anymore and I can't figure out why. It works in b10 and if I compare this PR against it I see nothing that would break the subscribe. This is without HA. For example, sending a publish to `ems-esp/system/test` is picked up in b10 but not in b11. I can see the MQTT broker receives the subscribe request so on the server everything looks solid. Also no library updates. I've spent 2hrs trying to figure this out, and I'm sure it's something stupid I've overlooked. very frustrating! If you have some time could you take a look please?

proddy:
> I'll take a look when i'm back home, tomorrow evening or mo morning.

Thanks, it's driving me nuts and just can't figure it out. To reproduce
1. Build this PR (EMS-ESP b11) (building with EMSESP_DEBUG in pio_local.ini gives more info as you know)
2. Turn off HA discovery in MQTT Settings so it's plain vanilla MQTT
3. Use MQTTExplorer to publish anything to one of the subscribed topics like `ems-esp/system` or `ems-esp/analogsensor`
4. you should see it being picked up on the log and console as a MQTT topic received with the contents of the package

only with b11 (this PR) step 4 doesn't comes up with nothing I don't know why since most of the changes are HA related, except for the NTP time in the info topic.

proddy:
> > I'll take a look when i'm back home, tomorrow evening or mo morning.
> 
> Thanks, it's driving me nuts and just can't figure it out. To reproduce
> 
> 1. Build this PR (EMS-ESP b11) (building with EMSESP_DEBUG in pio_local.ini gives more info as you know)
> 2. Turn off HA discovery in MQTT Settings so it's plain vanilla MQTT
> 3. Use MQTTExplorer to publish anything to one of the subscribed topics like `ems-esp/system` or `ems-esp/analogsensor`
> 4. you should see it being picked up on the log and console as a MQTT topic received with the contents of the package
> 
> only with b11 (this PR) step 4 doesn't comes up with nothing I don't know why since most of the changes are HA related, except for the NTP time in the info topic.

@MichaelDvP I found the error. I had removed `maxTopicLength` from MqttSettingsService, now I've added it back and hardcoded to the default.

proddy:
merging...

MichaelDvP:
I've tested now the latest dev, the maxTopicLength is only used here: https://github.com/emsesp/EMS-ESP32/blob/d080f5db0f317bfe521781fbbcedb6f69c6c5e4b/lib/framework/MqttSettingsService.cpp#L140
We can only set this line to FACTORY_MQTT_MAX_TOPIC_LENGTH (also tested)

I'm not happy with changing the command log to INFO, With Thomas' iobroker-adater getting device values every10 sec the log is flooded.

I'll make a few changes in a PR.

proddy:
> I'm not happy with changing the command log to INFO, With Thomas' iobroker-adater getting device values every10 sec the log is flooded.

good point, forgot about that. Can you change it back to `LOG_DEBUG(("Calling command....` then and add a comment so I remember in case I decide to change it again :-)

PR #793, opened 2022-12-06
option to enable multiple instances using a flag

proddy:


proddy:
I noticed the system MQTT topics may need to always be prefixed with the base. I'll check how it looks like in 3.4 and make sure its backwards compatible.

MichaelDvP:
Nice idea to keep backward compatibility.
But it's an option only affects HA discovery, isn't it better to place the checkbox to HA section like this (and hide with HA disabled):
grafik

proddy:
Yeah. I thought about that too after I implemented it.

Issue #795, opened 2022-12-07
CORS support


thegitarist:
Hello,

I would like to add control buttons to my grafana dashboard, that triggers actions via open api access. Currently this fails due to CORS limitations.

Would be great having this feature.

THNX & BR

MichaelDvP:
esp8266 have a [compile-option](https://github.com/rjwats/esp8266-react/blob/b9b9c1b2149b3d4472125c5684045f434df66d42/platformio.ini#L16-L17) to enable CORS. I've tried to use the same scheme, but add as a network setting. Please check if this works for you:
https://github.com/MichaelDvP/EMS-ESP32/releases/tag/latest

Issue #789, opened 2022-12-04
Increase counter precision (analogsensors)


NFis:
I wanted to use the counter functionality for my gas meter. Every 0.01 m³ a pulse comes from my hall sensor (very good signal quality, hardware debounced). My gas meter has about 13000 m³ on its mechanical counter.  So I created an analogsensor with counter functionlity, Start Value 13xxxx and Factor 0.01: 
![grafik](https://user-images.githubusercontent.com/48214097/205496629-740a9b94-8904-48b1-800a-7b69b85651b2.png)
But after a while the counter in ems-esp differs from the real counter. EMS-ESP sensors internally use float values, so with its counter adding a small value (0.01) to a relative big one leads to deviations. 
My first workarund: configure the input as digital input and count the pulses externally in Home Assistant. 
Second workaround is a software change: use double in the Sensor class in analogsensors.cpp/h instead if float. In helpers.cpp/h the according render_value function has also to be adapted.
My version: 3.4.4

MichaelDvP:
Please test with https://github.com/MichaelDvP/EMS-ESP32/releases/tag/latest
I've changed to double, but have no impulse source to count and test.

NFis:
Ok, thank you. I installed the sw and watch the counter tomorrow. This night heating is off.

NFis:
The counter runs perfectly after 11 hours of heating now.

Issue #785, opened 2022-12-03
Add heartbeat publish interval


proddy:
Currently heatbeat topic publish interval is fixed at 60 seconds. Add to MQTT settings to make it configurable.

Issue #784, opened 2022-12-01
Thermostat data issue when tap water is active


PopaSchKing:
**Bug description**

Hi everyone,

I'm running into an issue where data from the thermostat (Hc1) is completely corrupted when tap water is active (boiler starts when tap water is pulled).
Date/Time, Current room temp., Current room target temp., Hc1 summer mode, ... change as soon as tap water is pulled and goo back to normal values afterwards.

This happens without any human interaction on the thermostat/boiler/Home Assistant/EMS-ESP.
Tx write is disabled, 100% quality on Rx/Tx/MQTT, 0 fail on both Rx/Tx/MQTT (would mean it's not a communication issue with the EMS-ESP ?).

EMS-ESP v3.4.4
ESP32 / BBQKees interface board
Boiler: Condens 5000i/Greenstar 8000
Thermostat: RC200/CW100

I'm adding 2 screenshots:
- First one is a graph from Grafana (Home Assistant --> InfluxDB --> Grafana).
  It shows the current room temperature, temperature target and if heating is on/off.
- Second (and third) is a screenshot of various Home Assistant data (boiler and thermostat from EMS-ESP).
  You can see that Date/Time changes when tap water is active, and the graph is a zoom on temp. and temp. target that changes at the same time.

I'm thinking about a potential power issue on the EMS bus which would corrupt the data..
Tested with a beefier 5V adapter for the ESP32, same behavior.

This is an issue opened following a discussion on Discord, many thanks for the help !

---

**Steps to reproduce**
Have everything running normally.
Pull water from a tap.
Issue hap
pens.

---

**Expected behavior**
No change of behavior / no data corruption when tap water is being pulled.

---

**Screenshots**

Temperature, Target and Heating on/off
![image](https://user-images.githubusercontent.com/19823177/205081546-13c2cab3-1220-475a-b87d-fa5413bef154.png)

HA data (Date/Time is the strangest thing to me)
![image](https://user-images.githubusercontent.com/19823177/205081810-d96ac723-94f3-4588-bbb3-46d81f3ae488.png)

Temps again but zoomed in on the same time period as the previous HA data (current room temp and target)
![image](https://user-images.githubusercontent.com/19823177/205084569-e41a7afe-2c07-4421-b979-3c498e76bf44.png)

---

**Device information**
```
{
  "System Status": {
    "version": "3.4.4",
    "uptime": "000+01:33:14.923",
    "freemem": 128,
[...]

PR #797, opened 2022-12-07
Some actual additions/changes, see description

MichaelDvP:
Please wait before merging for feedback of the testers. This PR adresses:
- move settingof  multiple instances to HA settings
- add heartbeat intervall #785
- increase counter precision #789 (already tested, ok)
- CORS support #795
- an additional LOG_DEBUG with mqtt-payload for #784 (maybe change to #ifdef EMSESP_DEBUG or remove if not needed anymore)

Changelog will be updated if tests are ok.


proddy:
I tested it and its working fine, so Merge anytime.
nice work by the way doing all the translations and adding the network details to `System::command_info()`. We should probably also include the  `multple_instances` in that json for supporting later.

MichaelDvP:
#795 is confirmed working, i've added the `rebootNeeded`for it. 
#784 seems solved by updating to 3.5, the extra debug message to analyze  is not needed and i've set it to `#if defined(EMSESP_DEBUG)` 
`multiple instances` is already in the info-command (mqtt section).

Clear to merge!

proddy:
> Clear to merge!

can you do this? You should be able to

PR #808, opened 2022-12-14
add v3.4 entity id flag to MQTT settings

proddy:
to stop HA people complaining!


MichaelDvP:
With this extra option HA users have 75% chance to set wrong option. Isn't it better to make only one combined option setting
- v3.4 style (single instance)
- v3.5 multiple instances

proddy:
> With this extra option HA users have 75% chance to set wrong option. Isn't it better to make only one combined option setting
> 
> * v3.4 style (single instance)
> * v3.5 multiple instances

ah I see what you did there, very clever! Since the Entity IDs are changing anyway with the support for multiple-instances, bundling the two together makes sense.

So having a pull-down option when is shown when Discovery is enabled, called "Entity ID format" with two options "<v3.5, single-instance, descriptive EN naming" and "multi-instance, internal short name". And the default for all new installs is the newer format.

MichaelDvP:
Yes that's what i mean, only 2 options old/new, does not matter if checkbox or pulldown. Default should be the new format.

proddy:
merged the two, please check

MichaelDvP:
The selectbox does not work and always resets to default. I think it does not like bool values and needs number.
In mqttsettingsservice i realized that settings = newsettings is **after** mqtt reset, so new settings become valid in next change or restart. Some of the new settings does not need mqtt_reset (timings, ...), we can skip the changed = true for these.

MichaelDvP (approved):
Working good now.

There is only a little difference for v3.4 compatibility in analog/dallas.
In v3.4.4 main the ' uniq-id'  and ' object-id'  were not identical and `object-id` was `analog_sensor_name`, `temperature_sensor_name`. Now `object-id` uses the uniq-id.
But i think this is not a problem, it's only a few sensors and there were complains for v.3.4 because of the name use, so object-id with gpio/id is better.

I think merge this version now, but keep in mind that a few ha users will ask.

proddy:
> Working good now.
> 
> There is only a little difference for v3.4 compatibility in analog/dallas. In v3.4.4 main the ' uniq-id' and ' object-id' were not identical and `object-id` was `analog_sensor_name`, `temperature_sensor_name`. Now `object-id` uses the uniq-id. But i think this is not a problem, it's only a few sensors and there were complains for v.3.4 because of the name use, so object-id with gpio/id is better.
> 
> I think merge this version now, but keep in mind that a few ha users will ask.

thanks for checking. Yes I did look at the sensors and decided to keep the new naming as it's better. The problem with the old one is that if you renamed one of the sensors, it would create a new entity ID in Home Assistant and you'll loose all the historic data anyway.

Issue #803, opened 2022-12-14
Support for Hybrid specific settings and statistics in Nefit Bosch Heatpumps


tvdhorst:
Some settings and statistics available in Nefit Bosch heatpumps do not seem to be available yet in EMS-ESP on BBQKees' hardware. Some of these are specific for Hybrid installations.

I played around with these settings and forced some changes, while monitoring them manually on the heatpump and logging on EMS-ESP. These are the setting and changes in their values:

Sensor values:
- Status aux heater with mix
  - Add. heat source
    - ... Off
    - 09:36 On
    - 09:44 On
    - 09:51 On
    - 10:01 Off
  - Mixing valve
    - ... 0%
    - 09:36 1%
    - 09:44 2%
    - 09:51 3%
    - 10:01 0%
- Additional heater status
  - ... Off
  - 09:36 Heating
  - 10:01 Off

Settings (change initiated by user):
- Auxiliary heater with mixer
  - Outs. temp par. mode
    - 09:07 0 degrees centigrade
    - 09:10 -1 degrees centigrade
    - 09:25 2 degrees centigrade
    - 09:37 1 degrees centigrade
    - 09:44 -2 degrees centigrade
    - 10:00 -5 degrees centigrade
- Energy/ price ratio
  - 09:07 3.5
  - 09:10 4.0
  - 09:25 4.3
  - 09:36 4.7
  - 09:44 4.9
  - 10:00 1.7
- Low-noise operation (als mentioned in #802)
  - Min. outside temp.
    - 09:06 7 degrees centigrade
    - 09:09 5 degrees centigrade
    - 09:25 4 degrees centigrade
    - 09:37 3 degrees centigrade
    - 09:43 4 degrees centigrade

[log-ems-esp.txt](https://github.com/emsesp/EMS-ESP32/files/10226195/log-ems-esp.txt)

Would it be possible to add these? Happy to help by providing additional info!

MichaelDvP:
Please add your system info (from help page). We have hybrid settings with telegram 0xBB in thermostat, but for your thermostat it is shown as unknow. Please also mention if these are settings and the change is initiated by user (e.g. single value thermostats-W->boiler), or sensor-values that are published somewhere in a telegram.

tvdhorst:
Here's my system info: [emsesp_info.txt](https://github.com/emsesp/EMS-ESP32/files/10227579/emsesp_info.txt)
I edited the startpost to distinguish between the sensor values and settings.

[...]

Issue #802, opened 2022-12-12
support silent mode option in Nefit Bosch heatpump


derSchreifritz:
Some settings available on the heatpump are not available through the bbqkees (yet) or I can't see or find them.

One of them is the silent mode option.

I'd like to see the option "Minimum outside temperature for silent mode" or other recognizable name in homeassistant
I did set it from 6 to 8, then to 9, 7 and back to 6, and exported the log that I saw during that in system at EMS-ESP. See attached txt file and foto.
[log(3).txt](https://github.com/emsesp/EMS-ESP32/files/10212029/log.3.txt)
![signal-2022-12-12-220656_002](https://user-images.githubusercontent.com/105242771/207154691-7938f2ee-243b-49fd-bbba-52c51ef1008f.jpeg)
The panel model on the heatpump is called HPC 410, and it says EMS 2 on the instruction book.

proddy:
thanks, we'll take a look.

derSchreifritz:
thanks, let me know if I can provide better information. or help in other ways.
At the moment, I'm even more interested in another setting I cant find via homeassistant, can I post that here as well?
It's the "auxiliary heater only" binary switch. I changed it twice from yes to no and back. A foto (this time in English) and log are below.
[log(4).txt](https://github.com/emsesp/EMS-ESP32/files/10215762/log.4.txt)
![signal-2022-12-13-085746_002](https://user-images.githubusercontent.com/105242771/207261140-3675d3c9-ad7e-4371-b87f-da9f44098940.jpeg)
I have no idea how you can find anything in there, thanks in advance for any help!

MichaelDvP:
Thanks for the logs.
> I did set it from 6 to 8, then to 9, 7 and back to 6,

```
001+00:25:04.154 TRACE 33470: [emsesp] Thermostat(0x10) -> Boiler(0x08), ?(0x0484), data: 07 (offset 11)
001+00:25:13.616 TRACE 33559: [emsesp] Thermostat(0x10) -> Boiler(0x08), ?(0x0484), data: 09 (offset 11)
001+00:25:19.679 TRACE 33578: [emsesp] Thermostat(0x10) -> Boiler(0x08), ?(0x0484), data: 08 (offset 11)
001+00:25:27.979 TRACE 33640: [emsesp] Thermostat(0x10) -> Boiler(0x08), ?(0x0484), data: 06 (offset 11)
```
The complete telegram is very long, i thnk it is not only silent mode related.
[...]

PR #815, opened 2022-12-17
Additions for #802, #803

MichaelDvP:
Check, but wait for confirmation that this all works in the heatpumps.


proddy:
looks all ok.

proddy:
I'm happy to merge and fix the issues later. Nothing like living on the edge, right? :)

MichaelDvP:
Technical it's mostly ok, but namings are confusing. In the linked manual i've seen that the hysteresis is regarding heatpump and not the aux heater, and there are hysteresis-heating, cooling and pool (in 484 offset 37, 35 , 33 or 37, 33, 35? i think).

But sure we can adapt later, it is only wrong names, not wrong function.
Also missing translation we'll get faster with a release and more users contributing.

PR #827, opened 2022-12-22
Polish translation update

pswid:


proddy:
I think we need to start marking somehow the new translations that still need translating

pswid:
You're right. Maybe you could prepare a scrpit that will scan `locale_translations.h` during compilation (e.g. looking for "") and generate raport/file containing only names that require translation? `locale_translations.h` is a big file and it's easy to miss something.

proddy:
I'll just put // TODO tags in and hope they get filtered out during the node webpack

Issue #835, opened 2022-12-24
Unrecognized EMS device (device ID 0x09, no product ID)


stmb-100:
EMS-ESP Version
v3.5.0b12
Device (Platform / SDK)
ESP32 / v4.4.2

2022-12-24 13:03:58.176 E 82: [telegram] Last Tx Read operation failed after 3 retries. Ignoring request: 0B 8A 02 00 1B
2022-12-24 13:04:12.662 E 83: [telegram] Last Tx Read operation failed after 3 retries. Ignoring request: 0B 89 02 00 1B
2022-12-24 13:04:12.663 W 84: [emsesp] Unrecognized EMS device (device ID 0x09, no product ID). Please report on GitHub.
2022-12-24 13:04:23.301 W 85: [telegram] Incomplete Rx: 0A 88 16 00 1B 80
2022-12-24 13:04:23.978 W 86: [telegram] Incomplete Rx: 0A 88 16 00 1B 80 08 00 07 00 0F 00 00 00 00 00 00 00 00 00 00 00 00
2022-12-24 13:04:24.221 E 87: [telegram] Last Tx Read operation failed after 3 retries. Ignoring request: 0B 8A 02 00 1B
2022-12-24 13:04:39.118 E 88: [telegram] Last Tx Read operation failed after 3 retries. Ignoring request: 0B 89 02 00 1B
2022-12-24 13:04:39.118 W 89: [emsesp] Unrecognized EMS device (device ID 0x09, no product ID). Please report on GitHub.
2022-12-24 13:04:45.750 W 90: [telegram] Incomplete Rx: 0A 90 3F 36 1B 00
2022-12-24 13:04:49.459 W 91: [telegram] Incomplete Rx: 0A 90 A5 1B 1B 00

MichaelDvP:
@stmb-100 Please add the system info from help page. A controller without product-id is very uncommon. There is also a temininal attached to the bus producing inclomplete telegrams (timing?). Maybe the sender 09 is misinterpreted from 0A because of timing differs >9%.
A complete log (all) of about 1-2min would also help to understand what's going on..

stmb-100:
[emsesp162_info.txt](https://github.com/emsesp/EMS-ESP32/files/10299905/emsesp162_info.txt)

stmb-100:
[...]

Issue #826, opened 2022-12-21
Set temp not possible with 2 CR10 thermostats


marcstone-be:
**Bug description**
V3.4.4
Everything is working perfectly with 1 CR10 thermostat connected, either the one of HC1 or the other for HC2.
When both HC1 and HC2 are connected, the readings are all fine. But only the values can be written to HC1, and not to HC2.

**Steps to reproduce**
- Connect 1 thermostat to HC1. Set manual temperature works.
- Connect 1 thermostat to HC2. Set manual temperature works.
- Connect 1 thermostat to HC1 AND 1 thermostat to HC2. Set manual temperature only works on HC1.

**Expected behavior**
When both thermostats are connected, setting manual temp should work on both HC1 and HC2.

Sysinfo: [emsesp_info.txt](https://github.com/emsesp/EMS-ESP32/files/10279037/emsesp_info.txt)

**Logging when only HC2 thermostat (0x19) is connected:**
000+00:03:08.222 N 19: [emsesp] Me(0x0B) <- Thermostat(0x19), RC300Curves(0x029C), length: 0x19
000+00:03:08.269 N 20: [emsesp] Thermostat(0x19) -> Me(0x0B), RC300Curves(0x029C), data: 04 01 00 00 00 01 00 00 3C
000+00:03:13.657 N 21: [emsesp] Me(0x0B) -> Thermostat(0x19), RC300Set(0x02BA), data: 1D (offset 10)
000+00:03:14.250 N 22: [emsesp] Thermostat(0x19) -> All(0x00), RC300Monitor(0x02A6), data: 1D (offset 3)
000+00:03:14.436 N 23: [emsesp] Thermostat(0x19) -> All(0x00), RC300Set(0x02BA), data: 1D (offset 10)
000+00:03:15.747 N 27: [emsesp] Thermostat(0x19) -> All(0x00), RC300Monitor(0x02A6), data: 00 9C 21 1D 00 00 1D 00 00 00 00 00 00 00 00 00 00 00 00 00 01 00
000+00:03:16.517 N 28: [emsesp] Me(0x0B) <- Thermostat(0x19), RC300Monitor(0x02A6), length: 0x19
000+00:03:16.633 N 29: [emsesp] Thermostat(0x19) -> Me(0x0B), RC300Monitor(0x02A6), data: 00 9C 21 1D 00 00 1D 00 00 00 00 00 00 00 00 00 00 00 00 00 01 00 06 1C 00
[...]

Issue #823, opened 2022-12-19
WEB interface - Problem loading (507:)


HansRemmerswaal:
I upgraded today to the latest version of EMS-ESP-3_5_0b12-ESP32.bin, I already had v3.5.0b9 installed (the one with the defrost fix in the compressor status). In the web interface I see that the number of entities for the boiler has increased from 107 to 127. So far so good.

When I click in the WEB interface on one of the devices I can't see the sensors, instead I'm getting "Problem loading (507:).
I have tried with Chrome and Edge.

**Screenshots**
![Values not visible in WEB interface](https://user-images.githubusercontent.com/116959990/208538815-e59a6593-e1d9-4f45-8218-d280cbb27082.png)

**Device information**
[emsesp_info.txt](https://github.com/emsesp/EMS-ESP32/files/10263223/emsesp_info.txt)

HansRemmerswaal:
Not sure what happened but suddenly I'm now getting the values...

![Annotation 2022-12-20 001330](https://user-images.githubusercontent.com/116959990/208546177-35529b2c-2e54-46e6-8729-cac393f80d1e.png)

MichaelDvP:
Known issue, see #769 and #800. 
http:507 is `Insufficient Storage`
For web output of dashboard and cstomization we need a large json buffer (20k). Sometimes there is no continous 20k block free in heap and  the http:get was answered by null, making the web unresponsable. Now it shows 507 and keeps responsable.
Check system page for free heap, if the max alloc block goes below ~39k you'll get the 507. But ems-esp bus is still read and mqtt sent.

Actually there is no solution how to save memory, reduce heap fragmentation or  switch to ESP32 wrover with PSRAM (or ESP32S3).

MaartenRs:
My Heap free shows 101kB and I still get this 507 error.
![image](https://user-images.githubusercontent.com/121044285/209185368-845695fe-4b5e-45f2-91c0-4188ea7cbcf9.png)

proddy:
@MichaelDvP I think we're hitting the 20k buffer and maybe need to switch to event streaming in a future version for sending data to the dashboard, rather than POSTing it. 

@MaartenRs when you get this error can you go into your chrome/edge browser, turn on developer mode (usually F12 on Windows), go to Network, hit Record, reload the page, click on the Boiler device so the error occurs again and then see what is in the devicData payload, especially the Content-Length

[...]

Issue #820, opened 2022-12-19
Webinterface shows wrong values


MaartenRs:
Webinterdace shows values of 21.474.836 kWh, while Home Assistent shows -1.

EMS-ESP Version
v3.5.0b12

Apparaat (Platform / SDK)
ESP32 / v4.4.2

<html>
<body>

TYPE | DESCRIPTION | ENTITIES |  
-- | -- | -- | --
  | Boiler | Enviline/Compress 6000AW/Hybrid 3000-7000iAW/SupraEco/Geo 5xx/WLW196i | 122 |  
  | Thermostat | RC300/RC310/Moduline 3000/1010H/CW400/Sense II/HPC410 | 48 |  
  | Thermostat | RC100H | 4 |  
  | Sensors | Attached EMS-ESP Sensors | 0

</body>
</html>

<html>
<body>

cooling control starts | 29 |  
-- | -- | --
dhw control starts2 | 21.474.836 |  
pool control starts | 21.474.836 |  
total energy consumption | 1.946 kWh |  
total energy consumption compressor | 1.947 kWh |  
energy consumption compressor heating | 1.898 kWh |  
energy consumption compressor dhw | 21.474.836 kWh |  
energy consumption compressor cooling | 22 kWh |  
energy consumption compressor pool | 21.474.836 kWh |  
total auxiliary electrical heater energy consumption | 21.474.836 kWh |  
auxiliary electrical heater energy consumption heating | 21.474.836 kWh |  
auxiliary electrical heater energy consumption dhw | 21.474.836 kWh |  
auxiliary electrical heater energy consumption pool | 21.474.836 kWh |  
total energy supplied | 7.394 kWh |  
total energy supplied heating | 7.357 kWh |  
total energy warm supplied dhw | 21.474.836 kWh |  
total energy supplied cooling | 36 kWh |  
total energy supplied pool | 21.474.836 kWh |  
compressor power output | 3,5 kW |  
 

<br class="Apple-interchange-newline">
</body>
</html>

MichaelDvP:
Confusion in browser cache? Shut down and restart browser and (to be sure) also ems-esp.

MaartenRs:
Restart PC, empty browser cache and restart ems-esp.
Error is still there. If I connect with another browser (Firefox), the result is the same.

heennkkee:
Do you expect these values to be read? I have similar values being sent out, like for instance 

total energy supplied pool: 21.474.836 kWh

[...]

PR #836, opened 2022-12-24
fixes and enhancements #826, #835, #820

MichaelDvP:
and maybe a bit #823 with web log buffer dynamically reduced. 
The setting is now renamed to "max buffer size", real buffer size reduces one by one if maxAlloc goes below 45k and increases back to setting if maxAllox is > 50k. (Could happen if mqtt queue is filled with ha config messages).

A lot of commits, but i've tried to cleanup, it's not so much changes.

Merry Christmas.


proddy:
looks super, thanks. And a Merry Christmas to you too!

Issue #832, opened 2022-12-24
Support upgrade/downgrade between versions


proddy:
- store version in the settings file
- add more intelligent upgrade/downgrade between versions
- use the [semantic versioning](https://semver.org/), to make sure users know a dev version is a dev version and have breaking changes

proddy:
I've added this to my fork. We're going to change the naming, so 3.5.0b14 goes to 3.5.0-dev.14 to keep within the spec

proddy:
PR for review

PR #840, opened 2022-12-25
added version upgrade/downgrade #832

proddy:
includes some other minor things

when upgrade <3.5.0b13 it will set the MQTT HA Entity ID format to 0, which is the old method using the long names


MichaelDvP (approved):
The changes to ems-esp are looking ok. I have not understand what this library do. +30k code and 6k ram just for comparing two version strings? There must be something other.

proddy:
> The changes to ems-esp are looking ok. I have not understand what this library do. +30k code and 6k ram just for comparing two version strings? There must be something other.

yes, it's indeed a lot of code but uses Templates so only compiles the code it uses. I think the overhead in mem is in the std:: libraries it uses. I'll search around again for alternative libraries. What's important is that it can parse the major/minor/patch from a version string.

MichaelDvP:
I think ot's not the lib, most of the memory change comes from the c++17 update, also i have compiled with a updated dev-platform, that also produces more code. Will go back to standard...

proddy:
@MichaelDvP I rolled back to c++11, it indeed saves about 2k ram

Issue #828, opened 2022-12-23
Create table of all EMS-ESP entities, by device, shortname/fullname and characteristics


proddy:
> Thanks, but how can i find the mapping for the version EMS-ESP-3_5_0b10-ESP32 ?

What do you need exactly? The mapping of the old names to the new names?

I'll create a new GH issue to maintain a table of the shortname, fullname etc. We've done a few passes at this but its not consistent

_Originally posted by @proddy in https://github.com/emsesp/EMS-ESP32/issues/804#issuecomment-1363723789_

proddy:
My first idea for this is to do it in the code (test.cpp) which will simulate adding all devices and then printing out their details (similar to the API call for each entity).

sciurius:
I have the entities for my boiler+thermostat in an md file that I can easily transform into any form you like.

zibous:
I have the problem that I have to adjust my automation, scripts and graphics when I use the new version b12. So I'm looking for the mappings so I know which names to change.

|Device            |EMS-ESP-3_5_0b10-ESP32                                     |single instance b12                       |single instance  v3.4|Multi instance b12|FIELD6|FIELD7|
|------------------|-----------------------------------------------------------|------------------------------------------|---------------------|------------------|------|------|
|EMS-ESP           |sensor.system_ems_bus                                      |sensor.system_bus_status                  |                     |                  |      |      |
|                  |sensor.system_free_memory                                  |sensor.system_freemem                     |                     |                  |      |      |
|                  |sensor.system_mqtt_fails                                   |sensor.system_mqttfails                   |                     |                  |      |      |
|                  |binary_sensor.system_ntp_status                            |binary_sensor.system_ntp_status           |                     |                  |      |      |
|                  |sensor.system_rx_fails                                     |sensor.system_rxfails                     |                     |                  |      |      |
|                  |sensor.system_rx_received                                  |sensor.system_rxreceived                  |                     |                  |      |      |
|                  |entity: binary_sensor.ems_esp_status                       |entity: binary_sensor.system_status       |                     |                  |      |      |
|                  |sensor.system_tx_fails                                     |sensor.system_txfails                     |                     |                  |      |      |
|                  |sensor.system_tx_reads                                     |sensor.system_txreads                     |                     |                  |      |      |
|                  |sensor.system_tx_writes                                    |sensor.system_txwrites                    |                     |                  |      |      |
[...]

PR #851, opened 2022-12-27
#828 dump all out entities (standalone only)

proddy:


MichaelDvP (approved):
I was a bit shocked: 110 files changed! 
But the main change with dump is much smaller. 
Looks all ok, maybe also enclose the header of dump_all_values also in a #ifdef section.

PR #856, opened 2022-12-28
 HA firstrun only delete config, create HA in next loop (mqtt queue)

MichaelDvP:
In your generated entity list i've seen that heatpumps now have 164 possible entities. On first start HA queues 164 remove messages and direct a number of config messages. This could blow the queue with 300 entries. With this change first the remove messages were queued. On next cycle, when queue is empty, the HA config is generated. Also helps to keep maxAlloc a bit larger.


proddy:
good spot.

PR #858, opened 2022-12-28
fix for #828 - show thermostat hc1 and mixer wwc

proddy:


src/devices/thermostat.cpp
```diff
@@ -194,6 +194,11 @@ Thermostat::Thermostat(uint8_t device_type, uint8_t device_id, uint8_t product_i
     EMSESP::send_read_request(EMS_TYPE_RCTime, device_id);
     EMSESP::send_read_request(0x12, device_id); // read last error (only published on errors)
     EMSESP::send_read_request(0xA2, device_id); // read errorCode (only published on errors)
+
+#if defined(EMSESP_STANDALONE_DUMP)
+    // if we're just dumping out values, create a single dummy hc
+    register_device_values_hc(std::make_shared<emsesp::Thermostat::HeatingCircuit>(1, 0)); // hc=1, no flags
```
MichaelDvP, line 200: As mentioned in the issues, flags should contain the model: 
`    register_device_values_hc(std::make_shared<emsesp::Thermostat::HeatingCircuit>(1, model)); // hc=1`

src/emsesp.cpp
```diff
@@ -320,7 +320,14 @@ void EMSESP::dump_all_values(uuid::console::Shell & shell) {
         // go through each device type so they are sorted
         for (const auto & device : device_library_) {
             if (device_class.first == device.device_type) {
-                emsdevices.push_back(EMSFactory::add(device.device_type, 0, device.product_id, "1.0", device.name, device.flags, EMSdevice::Brand::NO_BRAND));
+                uint8_t device_id = 0;
+                // Mixer class looks at device_id to determine type, so fixing to 0x28 which will give all the settings except flowSetTemp
+                if ((device.device_type == DeviceType::MIXER) && (device.flags == EMSdevice::EMS_DEVICE_FLAG_MMPLUS)) {
+                    device_id = 0x28; // hard code
+                }
+
```
MichaelDvP, line 328: Do i understand right? This is only one mixer with wwc1, but no mixer with hc1? Then i prefere to have only hc1, this is more common when using mixers.

MichaelDvP (approved):

src/emsesp.cpp
proddy: yes, I had to choose one. I'll use hc1 then

src/devices/thermostat.cpp
proddy: fixed.

src/emsesp.cpp
proddy: fixed. will check in soon

Issue #865, opened 2022-12-30
web api not returning entity details for second mixer (heating circuit 2)


tglynx:
Hi, this issue may be related to #857. When using the web API, details of an entity can be queried using the full path of an entity in the web API.

As an example http://ems-esp/api/boiler/wwsettemp returns details of the entity wwsettemp.

```
{
  "name": "wwsettemp",
  "fullname": "dhw set temperature",
  "circuit": "ww",
  "value": 13,
  "type": "number",
  "uom": "°C",
  "readable": true,
  "writeable": false,
  "visible": true
}
```

This  does not work with the mixer entities for the second mixer in the system.
Scenario: 

http://ems-esp/api/mixer returns

```
{
  "hc1": {
    "flowtemphc": 28.3,
    "valvestatus": 6,
    "flowsettemp": 29,
    "pumpstatus": "on"
  },
  "hc2": {
    "flowtemphc": 29.4,
    "valvestatus": 0,
    "flowsettemp": 0,
    "pumpstatus": "off"
  }
}
```

http://ems-esp/api/mixer/hc1/flowtemphc returns

```
{
  "name": "flowtemphc",
  "fullname": "hc1 flow temperature (TC1)",
  "circuit": "hc1",
  "value": 28.1,
  "type": "number",
  "uom": "°C",
  "readable": true,
  "writeable": false,
  "visible": true
}
```

Now the problem: 
http://ems-esp/api/mixer/hc2/flowtemphc returns

```
{
  "message": "cannot find values for entity 'flowtemphc'"
}
```

EMS-ESP logs the following errors when using the ioBroker Adapter to enumerate the hc2 entities:

```
001+03:01:10.313 E 203: [emsesp] Call failed with error: cannot find values for entity 'flowtemphc' (Error)
001+03:01:10.325 E 204: [emsesp] Call failed with error: cannot find values for entity 'valvestatus' (Error)
001+03:01:10.325 E 205: [emsesp] Call failed with error: cannot find values for entity 'flowsettemp' (Error)
001+03:01:10.325 E 206: [emsesp] Call failed with error: cannot find values for entity 'pumpstatus' (Error)
```

I would expect that the details of hc2 entities can be queried the same way as the details of the hc1 entities.

http://ems-esp/api/system returns

```
{
  "System Info": {
    "version": "3.5.0-dev.14",
    "platform": "ESP32",
    "uptime": "001+03:18:21.677",
    "uptime (seconds)": 98301,
[...]

Issue #857, opened 2022-12-28
ahs / am200: web-api needs corrections


tp1de:
I got a GH issue in respect to my ioBroker adapter (https://github.com/tp1de/ioBroker.ems-esp/issues/12) opened by @tglynx.
The ioBroker Adapter is polling ems-esp by using web-api calls.  
I could find the following issues which need corrections:

1.  the  http://ems-esp/api/system command returns two times the device boiler. One for the boiler and one for the alternative heat source ahs. Device name should be unique either "boiler" and "boiler2" or even better using device name "ahs" for an am200 device
2. an http://ems-esp/api/boiler  returns the "ahs" as a substructure within json - I adjusted my ioBroker to get values. Better to get the ahs values by polling http://ems-esp/api/ahs
3. Reading single ahs values e.g. http://ems-esp/api/boiler/ahs/altflowtemp do return an error: { "message": "cannot find values for entity 'ahs/altflowtemp'" }. So single field attributes can't be read.

From my point of view I recommend to adjust.

MichaelDvP:
Please ask the user to participate here and post the system info. 
We can setup a own device-class ahs, but need someone with this device to test and give feedback.
Or only fix the prefix `ahs/` in point 3?

tglynx:
Hi, 

i am here and ready to test and give feedback. 

I think it would be the right way to implement a separate device class for the am200 module - as for mixers (MM100 and so on) or solar modules (SM100...).
However, it is a heat source module ... the alternative heat source module can be used to directly control a pellet stove with target temperature (or just a simple wood stove) and has the ability to block the primary heat source for a given time.

As i can see the implementation that has been done already also happened on user request and i really appreciate all the work you put into this. Great project!

MichaelDvP:
Yes, it's better to create a own device class. The boiler have a lot of entities making problems with buffer sizes.
I'll call the class `heatsource` and move the am200. 
For now we know only the AM200 with device-id 0x60, i'll add it unnested. If more modules are possible we will need to rename with tag HS1... 
The AM200 can be used as single boiler, pos 9/10. I think then it will use devcie_id 8 as a boiler and other telegrams. We have to check if someone uses this.

Boilers connected to a cascade module will still go to boiler class with tag HS1.., there are only a few values and the boilers keep there product-id, only changes device-id to 70...

I'll make a pr. @proddy please check if you agree that this is the right way.

MichaelDvP:
[...]

Issue #841, opened 2022-12-26
console command crashes EMS-ESP


proddy:
On 3.5.0b13 and 3.5.0-dev.14 

```
ems-esp:# call thermostat hc1 mode eco

Connection closed by foreign host
```

The entire device was just restarted automatically.

_Originally posted by @akanto in https://github.com/emsesp/EMS-ESP32/issues/773#issuecomment-1365040363_

MichaelDvP:
I'm not sure what is the best fix. 
Because the commandline only uses a prefix with no command a nullptr is returned here: https://github.com/emsesp/EMS-ESP32/blob/d08a1224e2df894b937bbe575faf8f53faae8e5c/src/command.cpp#L228-L231
It does not crash if we return the empty string instead.

The crash happens with nullptr here: https://github.com/emsesp/EMS-ESP32/blob/d08a1224e2df894b937bbe575faf8f53faae8e5c/src/command.cpp#L253
in this strcmp: https://github.com/emsesp/EMS-ESP32/blob/d08a1224e2df894b937bbe575faf8f53faae8e5c/src/emsdevice.cpp#L280
We can check for nullptr or use std::string instead of const char* (like in is_readonly here: https://github.com/emsesp/EMS-ESP32/blob/d08a1224e2df894b937bbe575faf8f53faae8e5c/src/emsdevice.cpp#L636)

But with other commandlines like `call thermostat hc1`  it also crashes here: https://github.com/emsesp/EMS-ESP32/blob/d08a1224e2df894b937bbe575faf8f53faae8e5c/src/command.cpp#L265

I think an easy fix is inserting at the start of Command::call: 
```
    if (cmd == nullptr) {
        return CommandRet::NOT_FOUND;
    }
```
and the critical (crashing) functions are not called.

proddy:
thanks for hunting that one down. Yes, the easiest fix is as you suggested. Do you mind fixing it?

proddy:
seeing similar behaviour with testing the mixer

```
ems-esp:# call mixer info
{
  "wwc1 current temperature": 56.5,
  "wwc1 pump status in assigned wwc (PC1)": "off",
  "wwc1 temperature switch in assigned wwc (MC1)": 0,
  "wwc2 current temperature": 62.0,
  "wwc2 pump status in assigned wwc (PC1)": "off",
  "wwc2 temperature switch in assigned wwc (MC1)": 0,
  "hc1 mixing valve actuator (VC1)": 0,
[...]

PR #870, opened 2022-12-31
Changes to heatsource, api, console

MichaelDvP:
- fix console crashes #841 
- AM200 to heatsource as ahs1 #857
- cascaded boilers to heatsource hs1..hs16
- fix api output for multiple devices #865


proddy:
I reviewed the code and tested it (the best I could). looks good. There is a change I need to make in the standalone arduino string library to support `api_data` but it's minor and I can do it after you've merged this

MichaelDvP:
I've merged the latest dev, solving some merge conflicts and include the latest user inputs.
Also added a translation for "Sensors" in Dashboard device list.
I think we can merge now.

proddy:
ok!

PR #873, opened 2023-01-01
Additional SE translations and corrections

mvjt:
If you add the possibility to change the device types on the dashboards (Boiler, Thermostat etc) I`ll make another commit to this PR.


proddy:
Because of #874 I had to merge this in manually. They'll be in the next PR

Issue #874, opened 2023-01-01
Plural of "days" under EMS Status not working for Swedish


mvjt:
**Bug description**
It seems plural for "days" for EMS Status is not working for Swedish but works for other languages. 
Day Singular = Dag
Day Plural = Dagar

In index.ts
NUM_DAYS: '{num} dag{{ar}}',

Screenshot where it should say "dagar" and not "dag"
<img width="449" alt="image" src="https://user-images.githubusercontent.com/34924009/210168820-6c14d0c7-3d4e-4171-8e58-8b883eca2bf4.png">

Version: v3.5.0-dev.14

MichaelDvP:
I think this is because the pl base have no fixed part: `NUM_DAYS: '{num} {{dni|dzień|dni|dni|dni}}',`
Changing the pl to: `NUM_DAYS: '{num} d{{ni|zień|ni|ni|ni}}',` should work,
 or set all other languages to `NUM_DAYS: '{num} {{dag|dagar}}',`

mvjt:
> I think this is because the pl base have no fixed part: `NUM_DAYS: '{num} {{dni|dzień|dni|dni|dni}}',`
> Changing the pl to: `NUM_DAYS: '{num} d{{ni|zień|ni|ni|ni}}',` should work,
> or set all other languages to `NUM_DAYS: '{num} {{dag|dagar}}',`

Thanks Michael, I'm not sure I understand the syntax but I used the EN one as reference...

proddy:
I think it's something else that is causing this. And it only happens in the Bus Status and only for SE. It renders correctly in the Dashboard for the "* burner operating time".

MichaelDvP:
For many days like in "burner operating time" the SE also renders correctly. And for minutes, hours also all languages work and PL defines cases for 1, 2, few, many, other languages only single/many. The only difference i see is that days are not defined as fixedpart/suffix, but use cases for each plural. My emsesp rarly reaches 2 days uptime without updating. Difficult to check.

proddy:
yeah, this is a weird one and related to how the page is rendered are the variables loaded. I have a workaround which I'll push shortly.

proddy:
I found the problem after tracing it down to internal JS library `Intl.PluralRules`. Turns out we're using the wrong locale for Swedish. It's SV and not SE (ISO 639). doh! Will do a PR...

mvjt:
Nice work. Translations finally look really good too. The only missing piece is #872.

Issue #875, opened 2023-01-01
Small typo i18n/de & missing link in docs


MichScha:
**Bug description**

### Typo in DE translation

https://github.com/emsesp/EMS-ESP32/blob/dev/interface/src/i18n/de/index.ts
Current
-   ENABLE_ANALOG: 'Aktiviere Analogsensorsen',

Corrected
-   ENABLE_ANALOG: 'Aktiviere **_Analogsensoren_**',

Current
-   MQTT_NEST_1: 'Als Nester in in einem Gesamttopic',

Corrected
-   MQTT_NEST_1: 'Eingebettet in einem Gesamttopic',

### Missing anchor in docs for "see the table below"

- https://emsesp.github.io/docs/Commands/
_an EMS device entity also referred to as an <entity>. See the [table below](https://emsesp.github.io/docs/Commands/Command?id=ems-device-entity-names.md) for the complete list_

proddy:
thanks, fixed in the next PR

PR #876, opened 2023-01-01
Fix Swedish - #874 and #875

proddy:


MichaelDvP (approved):
Looks all ok, but i doubt that this fixes #874.
If this was the cause all plurals should wrongin SE, but all plurals working for SE, except days. Also many days are working for SE, only for 2, few it's wrong. 
Now look at the plural declaration in PL and find the difference between days and all other num declarations.

proddy:
> Looks all ok, but i doubt that this fixes #874. If this was the cause all plurals should wrongin SE, but all plurals working for SE, except days. Also many days are working for SE, only for 2, few it's wrong. Now look at the plural declaration in PL and find the difference between days and all other num declarations.

The plurals are handled by the Intl JS library base do rules per locale, which typescript-i18n used. When it was SE it was using the Sami language. I'm not sure why it worked in the dashboard and not in the status. It's also not just NUM_DAYS but also NUM_MINUTES and the others too.

proddy:
For PL it's weird how it's implemented and I think incorrectly. I found out that numeric plurals in Polish are tricky (see https://www.quora.com/How-do-Polish-plurals-work) but the library should handle this and not be implemented the way it is now.

MichaelDvP:
> When it was SE it was using the Sami language. I'm not sure why it worked in the dashboard and not in the status. It's also not just NUM_DAYS but also NUM_MINUTES and the others too.

Very strange, i did not see issues for the plurals when switching language. But plural rules for se and sv are identical. See http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html

I also don't know what have caused it and did not see it in dashboard for any wrong plural (including days)  and in status minutes and seconds were ok for all languages (including SE), i thought it's just days, let's wait 2 days if it is fixed.

proddy:
I hope it's fixed but we'll see. I quick test I did was
- add `**{LL.NUM_MINUTES({num: 2})}**` to the bottom of `SignIn.tsx`
- Make the PL translation the same as EN in `i18n/pl/index.ts` just to rule out the base translation messing the formatting up
- `npm run standalone` and click on the flags and see how the number is rendered
- it didn't work for Swedish when it was set to `se`

Issue #872, opened 2023-01-01
Translate Device Name


proddy:
Request from Discord to translate the device names (boiler, thermostat etc)

![image](https://user-images.githubusercontent.com/1230712/210166802-027ddd43-0a82-4850-bd10-68066fd1a14f.png)

It's a change in EMSdevice::device_type_name()

MichaelDvP:
Think about the lot of usages: mqtt topic, system-info, api device, command device and which should be translated without braking to much. Dashboard/Customizaton and HA friendly name are safe, i think.

proddy:
yes, good point. I did check for all those cases

proddy:
I think I'll only make this change in the WebUI and leave the rest (console, mqtt, API as it is)

proddy:
merged into dev

PR #885, opened 2023-01-03
#872

proddy:


src/web/WebSettingsService.cpp
```diff
@@ -240,10 +240,12 @@ StateUpdateResult WebSettings::update(JsonObject & root, WebSettings & settings)
 
     String old_local = settings.locale;
     settings.locale  = root["locale"] | EMSESP_DEFAULT_LOCALE;
+    check_flag(prev, settings.low_clock, ChangeFlags::RESTART);
```
MichaelDvP, line 243: This seems like a copy&paste, it makes no sense here.

interface/src/project/DeviceIcon.tsx
```diff
@@ -13,32 +13,35 @@ import { AiOutlineAlert } from 'react-icons/ai';
...
-const DeviceIcon: FC<DeviceIconProps> = ({ type }) => {
-  switch (type) {
-    case 'Boiler':
-      return <CgSmartHomeBoiler />;
-    case 'Sensor':
+// matches emsdevice.h DeviceType
+const DeviceIcon: FC<DeviceIconProps> = ({ type_id }) => {
+  switch (type_id) {
+    case 1:
+    case 2:
```
MichaelDvP, line 23: Can we use a enum here, then it's easier to check against emsdevice.h.
```
// matches emsdevice.h DeviceType
const enum DeviceType {
  SYSTEM = 0,
  DALLASSENSOR,
  ANALOGSENSOR,
  BOILER,
  THERMOSTAT,
  MIXER,
  SOLAR,
  HEATPUMP,
  GATEWAY,
  SWITCH,
  CONTROLLER,
  CONNECT,
  ALERT,
  PUMP,
  GENERIC,
  HEATSOURCE,
  UNKNOWN
}

const DeviceIcon: FC<DeviceIconProps> = ({ type_id }) => {
  switch (type_id) {
    case DeviceType.DALLASSENSOR:
    case DeviceType.ANALOGSENSOR:
      return <MdOutlineSensors />;
    case DeviceType.BOILER:
    case DeviceType.HEATSOURCE:
      return <CgSmartHomeBoiler />;
    case DeviceType.THERMOSTAT:
      return <MdThermostatAuto />;
    case DeviceType.MIXER:
      return <AiOutlineControl />;
    case DeviceType.SOLAR:
      return <FaSolarPanel />;
    case DeviceType.HEATPUMP:
      return <GiHeatHaze />;
    case DeviceType.GATEWAY:
      return <AiOutlineGateway />;
    case DeviceType.SWITCH:
      return <TiFlowSwitch />;
    case DeviceType.CONTROLLER:
    case DeviceType.CONNECT:
      return <VscVmConnect />;
    case DeviceType.ALERT:
      return <AiOutlineAlert />;
    case DeviceType.PUMP:
      return <AiOutlineChrome />;
    default:
      return null;
  }
};
```

proddy: yes much better. I'll update

PR #887, opened 2023-01-05
README update and one change to api_data

proddy:


proddy:
@MichaelDvP ok with the api_data change?

MichaelDvP (approved):

PR #901, opened 2023-01-10
added dhw alternating operation

pswid:
In the thermostat menu options look like this:

dhw alternating
prioritise dhw
prioritise heating

### DIFF
--- a/src/helpers.cpp
+++ b/src/helpers.cpp
@@ -596,14 +596,16 @@ bool Helpers::value2bool(const char * value, bool & value_b) {
         return false;
     }
 
-    std::string bool_str = toLower(value); // convert to lower case
+    std::string bool_str = value;
 
-    if ((bool_str == std::string(Helpers::translated_word(FL_(on)))) || (bool_str == "on") || (bool_str == "1") || (bool_str == "true")) {
+    if ((bool_str == std::string(Helpers::translated_word(FL_(on)))) || (bool_str == std::string(Helpers::translated_word(FL_(ON))))
+        || (bool_str == "on") || (bool_str == "1") || (bool_str == "true")) {
         value_b = true;
         return true; // is a bool
     }
 
-    if ((bool_str == std::string(Helpers::translated_word(FL_(off)))) || (bool_str == "off") || (bool_str == "0") || (bool_str == "false")) {
+    if ((bool_str == std::string(Helpers::translated_word(FL_(off)))) || (bool_str == std::string(Helpers::translated_word(FL_(OFF))))
+        || (bool_str == "off") || (bool_str == "0") || (bool_str == "false")) {
         value_b = false;
         return true; // is a bool
     }
--- a/src/helpers.cpp
+++ b/src/helpers.cpp
@@ -596,15 +596,15 @@ bool Helpers::value2bool(const char * value, bool & value_b) {
         return false;
     }
 
-    std::string bool_str = value;
+    std::string bool_str = toLower(value);
 
-    if ((bool_str == std::string(Helpers::translated_word(FL_(on)))) || (bool_str == std::string(Helpers::translated_word(FL_(ON))))
+    if ((bool_str == std::string(Helpers::translated_word(FL_(on)))) || (bool_str == toLower(Helpers::translated_word(FL_(ON))))
         || (bool_str == "on") || (bool_str == "1") || (bool_str == "true")) {
         value_b = true;
         return true; // is a bool
     }
 
-    if ((bool_str == std::string(Helpers::translated_word(FL_(off)))) || (bool_str == std::string(Helpers::translated_word(FL_(OFF))))
+    if ((bool_str == std::string(Helpers::translated_word(FL_(off)))) || (bool_str == toLower(Helpers::translated_word(FL_(OFF))))
         || (bool_str == "off") || (bool_str == "0") || (bool_str == "false")) {
         value_b = false;
         return true; // is a bool
--- a/src/helpers.cpp
+++ b/src/helpers.cpp
@@ -678,6 +678,41 @@ std::string Helpers::toUpper(std::string const & s) {
     return lc;
 }
 
+// capitalizes one UTF-8 character in char array
+// works with Latin1 (1 byte) and Polish (2 bytes) characters
+// TODO add special characters that occur in other supported languages
+void Helpers::CharToUpperUTF8(char * c) {
+    switch (*c) {
+    case 0xC3:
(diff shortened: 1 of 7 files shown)


src/helpers.cpp
```diff
@@ -596,14 +596,16 @@ bool Helpers::value2bool(const char * value, bool & value_b) {
         return false;
     }
 
-    std::string bool_str = toLower(value); // convert to lower case
+    std::string bool_str = value;
 
-    if ((bool_str == std::string(Helpers::translated_word(FL_(on)))) || (bool_str == "on") || (bool_str == "1") || (bool_str == "true")) {
+    if ((bool_str == std::string(Helpers::translated_word(FL_(on)))) || (bool_str == std::string(Helpers::translated_word(FL_(ON))))
```
proddy, line 601: there are a number of places in Helpers.cpp where we test for boolean and it's translations and alternatives. Maybe good to create a separate function for this like `Helpers::isBool(const char * bool_s)` ?

proddy (changes requested):

src/helpers.cpp
pswid: I don't understand what the new function you propose is supposed to do. Can you explain your idea?
In Helpers.cpp (except `value2bool`) I only see `value2enum` functions that test translated boolean names. I missed it. So, `value2enum` should also be rewritten to not use `toLower` helper, but tests both uppercase and lowercase variants. Would you do it?

proddy: ok, I'll take a look later and refactor myself. it's not that important.

code looks fine, @MichaelDvP ok to merge?

MichaelDvP (changes requested):
I see the intension to add "OFF" to the bool input with pl language, but we have to keep the toLower(value) to match all mixed case writings like "True", "On". 
Please re-add `bool_str = toLower(value)` and change in the if's to `bool_str == toLower(Helpers::translated_word(FL_(OFF))` then all options should work.

pswid:
Done. Sorry for delay. It works but I noticed that `toLower()` and `toUpper()` do not work on Polish characters (ĄĘĆŁŃÓŚŹŻ). It should be fixed but I don't know how. I read somewhere that `locale` needs to be set properly.

MichaelDvP:
I think you have to check the special charcater here explicide:
https://github.com/emsesp/EMS-ESP32/blob/88404bcb56d6be0e79cb2dda2fc1931be0102e4d/src/helpers.cpp#L665

pswid:
toupper() is used e.g. here:
https://github.com/emsesp/EMS-ESP32/blob/88404bcb56d6be0e79cb2dda2fc1931be0102e4d/src/mqtt.cpp#L1135
and it doesn't work. The first letter of HA friendy names starting with Polish character is not capitalized (e.g. "łączny czas grzania"), so I assume that std::tolower() will not work also.

MichaelDvP:
Yes the std::tolower  only converts A-Z. I thought of something like a Helpers::tolower:
(i have not tested if it works)
```
int Helpers::tolower(int _c) {
    switch (_c) {
    case 'Ż': return 'ż';
    // insert other special characters here
    default: return std::tolower(_c);
    }
}

std::string Helpers::toLower(std::string const & s) {
    std::string lc = s;
    std::transform(lc.begin(), lc.end(), lc.begin(), [](unsigned char c) { return tolower(c); }); // use own tolower() here
    return lc;
}
```

proddy:
> Yes the std::tolower only converts A-Z. I thought of something like a Helpers::tolower: (i have not tested if it works)
> 
> ```
> int Helpers::tolower(int _c) {
>     switch (_c) {
>     case 'Ż': return 'ż';
>     // insert other special characters here
>     default: return std::tolower(_c);
>     }
> }
> 
> std::string Helpers::toLower(std::string const & s) {
>     std::string lc = s;
>     std::transform(lc.begin(), lc.end(), lc.begin(), [](unsigned char c) { return tolower(c); }); // use own tolower() here
>     return lc;
> }
> ```

that would work. Or just don't bother with toLower/Upper because of the different charsets. I don't think we'll miss much.

pswid:
I wouldn't want special characters to be ignored. This can cause various problems in other places, especially if more languages using special characters will be added in the future.

The main cause of problems with special characters (e.g. Polish letters) is due to the fact that in UTF-8 they are not encoded in 1 byte, but in 2 bytes (in UTF-8 there are characters that require even 4 bytes!).

My commit [4f05dda](https://github.com/emsesp/EMS-ESP32/pull/901/commits/4f05ddab93d4321c236e7489b63a9716b45fda54) doesn't "break" anything, but fixes a bug with the boolean "wł./wył." option, so I think it can be merged (and possibly improved in the future).

Since there are currently no entities starting in Polish with a special character other than "ł", temporary solution could by replacing https://github.com/emsesp/EMS-ESP32/blob/88404bcb56d6be0e79cb2dda2fc1931be0102e4d/src/mqtt.cpp#L1135 by:
```
    if (F_name[0] == 0xC5 && F_name[1] == 0x82)  //if first letter is Polish "ł" (UTF-8: 0xC5 0x82)...
        F_name[1] = 0x81;   //change it to "Ł" (UTF-8: 0xC5 0x81)
    else
        F_name[0]     = toupper(F_name[0]); // capitalize first letter
```

Ultimately, however, own `toupper` & `tolower` procedures should be created that will also work with special characters (are not based on 8-bit variables such as `char`).

MichaelDvP (approved):

src/helpers.cpp
```diff
@@ -678,6 +678,41 @@ std::string Helpers::toUpper(std::string const & s) {
...
 }
 
+// capitalizes one UTF-8 character in char array
+// works with Latin1 (1 byte) and Polish (2 bytes) characters
+// TODO add special characters that occur in other supported languages
+void Helpers::CharToUpperUTF8(char * c) {
+    switch (*c) {
+    case 0xC3:
+        if (*(c + 1) == 0xB3) //ó (0xC3,0xB3) -> Ó (0xC3,0x93)
+            *(c + 1) = 0x93;
```
MichaelDvP, line 688: You can change this to match more special characters to:
```
        // grave, acute, circumflex, diaeresis, etc.
        if ((*(c + 1) >= 0xA0) && (*(c + 1) <= 0xBE)) {
            *(c + 1) -= 0x20;
        }
```

proddy: when this is done I'll merge

pswid: done. sorry for the delay

proddy: thanks!

proddy: ok to merge @MichaelDvP ?

MichaelDvP: Yes.

PR #909, opened 2023-01-13
weblog send queue one by one, no json buffer needed

MichaelDvP:
This saves some flash and a lot of heap. Weblog now only uses heap for the deque.


proddy:
Thanks, I'll check later this weekend when I'm back (as part of the memory profiling work)

PR #923, opened 2023-01-22
fixes and the tested heatpump/heatsource entities

MichaelDvP:
looks like a lot of changes, but mostly small commits. 
Turkish is not included, also not the blacklisting of entities and not the weblog buffer modification. These are in my blacklist branch.


proddy:
thanks. I'll also merge into the tec branch and update the issues

PR #929, opened 2023-01-23
PL translation update

pswid:


interface/src/framework/system/SystemLog.tsx
```diff
@@ -226,10 +226,10 @@ const SystemLog: FC = () => {
...
               name="max_messages"
               marks={[
-                { value: 25, label: '25' },
-                { value: 50, label: '50' },
-                { value: 75, label: '75' },
-                { value: 100, label: '100' }
+                { value: 25, label: '25%' },
+                { value: 50, label: '50%' },
+                { value: 75, label: '75%' },
+                { value: 100, label: '100%' }
```
MichaelDvP, line 232: It's not really a percent value, it's the maximum number of log messages stored in RAM if the weblog tab is not active.
If the tab is active, there is no limit from esp, only from browser memory. 
It's not the absolute number of messages stored in RAM, only the maximum, the real number depends on log and free heap (up to this max).
But it's to complicated to explain to users, so it does not matter if a scale point is 25, 25 msg ´, or 25% (i prefere wihout uom).

proddy: I also prefer without the %

pswid: Now I understand. Sorry for the confusion but the name of the option was not very clear. I removed "%". I also changed the Polish name of this option to "**Maximum buffer capacity (number of entries)**". It's not a perfect name, but I think it's more understandable. Maybe it should be modified in other languages as well?

proddy: think it's just easier to explain the help page (which I noticed is missing in https://emsesp.github.io/docs/Configuring/#logging)

pswid: If you think so. However, I personally prefer solutions that do not require the use of help/manual. 
What now? Should I undo the change of the option name in Polish? If you agree, I'd like to leave it as it is...

Issue #932, opened 2023-01-27
thermostat mode 'manual' missing as option


Peerke1:
**Bug description**
thermostat (hc1 in my case) is missing the 'manual' mode 

**Steps to reproduce**
put thermostat in manual mode (on Moduline 400 - press hand icon)
Adjust the heat index (number.thermostat_hc1_selected_room_temperature) in HA
temperature is adjusted and heater starts
adjust temperature on thermostat
heating isn't affected, burner state remains the same
put thermostat in manual mode
adjust temperature once more as previous adjustment is not used
heating is now adjusted to new temperature 
(heating either starts or stops depending on the room- and selected temp)

**Expected behavior**
thermostat should remain in manual mode and accept new temp setting from both HA and manual operation

![image](https://user-images.githubusercontent.com/48998939/215093948-fb0985ae-654f-4936-ab7d-da44e71c6212.png)

**Device information**
{
  "System Status": {
    "version": "3.5.0b7",
    "platform": "ESP32",
    "uptime": "000+05:26:35.359",
    "freemem": 118,
    "free_app": 74,
    "reset reason": "Software reset CPU / Software reset CPU"
  },
  "Network Status": {
    "connection": "Ethernet",
    "hostname": "ems-esp",
    "MAC": "0C:B8:15:4A:EA:0F",
    "IPv4 address": "*****",
    "IPv4 gateway": "****",
    "IPv4 nameserver": "208.67.222.222",
    "phy type": 1,
    "eth power": 16,
    "eth phy addr": 1,
    "eth clock mode": 0,
    "static ip config": false,
    "enable IPv6": false,
    "low bandwidth": false,
    "disable sleep": false,
    "AP provision mode": "disconnected",
    "AP security": "wpa2",
    "AP ssid": "ems-esp"
  },
  "NTP Status": {
    "network time": "disconnected",
    "enabled": false,
    "server": "time.google.com",
    "tz label": "Europe/Amsterdam"
  },
  "OTA Status": {
    "enabled": true,
    "port": 8266
  },
  "MQTT Status": {
    "MQTT status": "connected",
    "MQTT publishes": 6446,
    "MQTT queued": 0,
    "MQTT publish fails": 0,
    "MQTT connects": 1,
    "enabled": true,
    "client_id": "ems-esp",
    "keep alive": 60,
    "clean session": false,
    "base": "ems-esp",
    "discovery prefix": "homeassistant",
    "nested format": 1,
    "ha enabled": true,
[...]

Issue #891, opened 2023-01-06
Blacklist / Remove Entities


HansRemmerswaal:
You have done a great job in adding many entities to the EMS-ESP. In my case I see 125 for the Boiler and 53 for the Thermostat. But due to all these entities the memory of my BBQKEES module is getting full and my system started to restart due to this, see #859. I have disabled as much as possible all the functionality to free up memory and now it's running stable again for a few days. Soon you will add a few new entities and my module will start restarting again. 

Part of these entities are useless for my configuration, I have full electric so all the ones related to gas are irrelevant. 
Part of these entities will never change or are less relevant, e.g. I don't need the current date/time of my thermostat.
Luckily you have thought about that and added the option to hide these.  Unfortunatly hiding these will not reduce the memory.

**Proposed solution**
In the ideal world we would have in the settings the option to indicate the kind of configuration you have (hybrid / full electric / ...) and automatically all the irrelevant entities are excluded and not part of the memory. But I doubt that is a feasible solution as someone needs to categorize all the entities of all the devices.

A more workable solution could be a kind of black list. From the customization tab you can blacklist an entity which will remove it from the list of entities. A new blacklist tab will show all the entities that are blacklisted, from there it can be deleted from the blacklist. During the receiving of the telegrams the blacklist is checked and blacklisted entities are ignored. I think it's acceptable that the user needs to restart the EMS-ESP after blacklisting entities to free up the already occupied memory.

My expectation is that I can eliminate a part of the entities and with that free up memory. That allows me to enable some disabled functionality again, and will make it possible for you to add new entities again without worrying about the available memory.

proddy:
I've created a specific GH issue to look into optimizing the memory (#869) as it's not clear to me what the thresholds and lower limits are for Heap Memory and Max Alloc block before it causes EMS-ESP to run out of RAM and crash. I'm writing some code to simulate complex setups (like yours) which loads multiple device and entities to see where the memory goes. I have a hunch the problem is in the Max Alloc which is the largest amount of continuously aligned memory blocks which can be used to store a buffer. Often this becomes fragmented and gets smaller and smaller. We use large buffers for MQTT and the Web. If we solve that I think we should be fine.

Having all the entities loaded is important to be able to enable/disable them from the customization screen. I don't think they take too much internal memory as the long string names are stored in Flash memory. If the results of the memory investigation gets us nowhere we can implement the black-list as you suggested. 

[...]

PR #940, opened 2023-01-31
Entitiy blacklist (remove from mem, #891) and RC20/30 modes/seltemps #932

MichaelDvP:
dev-17 with changelog


proddy:
nice. I'll review and manually merge into the tec branch too.

BTW I always thought the C++ precompiler would automatically make smaller functions inline? Specially when using the -Os flag. hmmm

Issue #910, opened 2023-01-13
Make the mqtt icon for ha discovery customizable


tp1de:
Every ems-esp restart will re-eset changes done within HA in respect to icon changes. Please make the icon changeable within ems-esp customization ... next to changing the entity name.

proddy:
as a workaround, you could add a line to the `customization.yaml` file (https://www.home-assistant.io/docs/configuration/customizing-devices/)

tp1de:
> as a workaround, you could add a line to the `customization.yaml` file

Thanks @proddy that' what I have done.

tp1de:
I tested with my km200 workflow in HA using node-red and mqtt discovery. Manual icon changes are kept persistent when re-sending mqtt config.

You said in discord:
>when EMS-ESP restarts it removes all the discovery config topics and rebuilds them, and some of them contain the 'ic' which is the icon, so any custom settings in HA are overwritten.

Why do you need to remove all discovery config topics on ems-esp restart? 
Is this linked to the "Set Clean Session" flag in mqtt settings? (without config topics will not be removed?)

proddy:
> Why do you need to remove all discovery config topics on ems-esp restart?
Is this linked to the "Set Clean Session" flag in mqtt settings? (without config topics will not be removed?)

If there is a Discovery Config topic in MQTT that points to another topic and data element and that doesn't exist, HA will vomit hundreds of warnings in its log because it can't find and render the value. This is because the data topic (e.g. boiler_data, thermostat_data) is dynamically built-up and only shows the values that exist in EMS-ESP. The same logic applies when a value is removed or becomes 'dormant', we want to exclude it from the payloads and also delete the Discovery topic to avoid warnings.

Clean Session is something different and internal to MQTT. See http://www.steves-internet-guide.com/mqtt-clean-sessions-example/

Issue #1067, opened 2023-02-20
MQTT crashes ems-esp while getting disconnected to server and reconnecting afterwards


tp1de:
@MichaelDvP @proddy  as discussed on Discord:

When MQTT (HA) gets disconnected and reconnects after a while, all entities are removed and created new.
Michael mentioned on Discord:  
"this generates a lot of messages (EMSdevice::ha_config_clear() {), because it's for all devices at once, and fills the queue".

When queue becomes too long then ems-esp crashes and reboots. Then the same procedure starts again, but due to discovery of devices / entities while reading telegrams needs some time, the start-up works.

I tested that some seconds of "unavailability" of the mqtt-server brings my ems-esp system to crash.
As a result any short-time  wifi-disconnects as well as longer offline backup-sessions will result in crashes.

I recognized that ems-esp wifi disconnects and reconnects during the day and changing the wifi access point on my mesh-network.
This takes some seconds. I might believe that this is the reason for my 2-3 crashes a day.

The automatic removal of the entities should be avoided wherever possible. It takes 1-2 minutes to recreate them by mqtt-discovery on HA. During this time dashboards and automations do not work or even crash. A MQTT reconnect does not need this removal function.

MichaelDvP:
@tp1de Could you compile/test the tech-branch? Or shoud i add a bin, or merge to my 3.5.1-dev? With the RC310 you have more entities, i get not restarts with my system, even if i enable HA discovery.

tp1de:
Hi guys I am lost for the moment how to get the tech upgrade.

When I clone with command: **git clone https://github.com/emsesp/EMS-ESP32.git**
I do not see any branches anymore to select the tech branch.

So I used Michaels repository and selected the tech-upgrade branch and compiled to: EMS-ESP-3_5_0-tec_4-ESP32.bin
This bin I can't upload with the webui - error upload failed (507).

Please advice how to go forward .....

proddy:
Here's a bin file : [EMS-ESP-3_5_0-tec_4-ESP32_4M.zip](https://github.com/emsesp/EMS-ESP32/files/10787534/EMS-ESP-3_5_0-tec_4-ESP32_4M.zip)

The tec branch has various optimizations under the hood which I'm working on. There are no breaking changes or new features.

proddy:
[...]

PR #1077, opened 2023-02-22
added "availability" section in HA Discovery

pswid:
This will allow to not remove discovery topics during each ems-esp restart (MQTT init), so it first step to solve issue #910.

After applying this fix (and removing from the code commands that delete discovery topics) HA no longer reports errors/warnings in the log. I'm testing if for over a week.
Now it is enough to delete discovery topics only when it is really needed (e.g. the entity has been removed by customization, discovery prefix has been changed or the HA option has been disabled in the configuration).


proddy:
err, hang on, this could be a solution to https://github.com/emsesp/EMS-ESP32/issues/1067 which we've been working on

are you saying by introducing the discovery availability attribute we no longer need to clean up old HA discovery config topics when EMS-ESP starts? That's quite cool.

proddy:
could you merge this into the tech-upgrade branch?

proddy:
> could you merge this into the tech-upgrade branch?

never mind, I've added it manually so I can test it. I'll report back here.

proddy:
merged with tech-upgrade branch

JokerGermany:
Pls be good, this would be huge for me.

pswid:
> err, hang on, this could be a solution to #1067 which we've been working on
> 
> are you saying by introducing the discovery availability attribute we no longer need to clean up old HA discovery config topics when EMS-ESP starts? That's quite cool.

It seems so. If no specific value exists in the topic `ems-esp/xxx_data`, then such an entity becomes "unavailable" in HA. This does not result errors in HA log.

Besides, all EMS-ESP entities become unavailable if topic `ems-esp/status` is different from `online`. During (re)generation of HA discovery topics, you could publish the status as, for example, `starting`. It will be visible in HA. Of course `system_status` entity would have to be changed from `binary_sensor` to `sensor`.

It is only necessary(?) to ensure that in a situation where HA discovery is generated (at the start of EMS-ESP or when the user changes format in the EMS-ESP MQTT configuration), data topics do not have value inconsistent with previous discovery definition. In such situations, it would be the best(?) to delete `ems-esp/xxx_data` topics, publish new (modified) HA discovery, and only then start publishing `ems_esp/xxx_data` again.

PR #1082, opened 2023-02-27
schedule commands timing, retry

MichaelDvP:
I've changed the timings: timer commands to esp-uptime, calender to RTC. Also added a retry counter, we can define if we like a retry for startup commands. Still wonder for what these startup commands are usefull, but i know `<device>/<cmd>` will not work before the bus is scanned.

### DIFF
--- a/src/web/WebSchedulerService.cpp
+++ b/src/web/WebSchedulerService.cpp
@@ -92,10 +92,10 @@ StateUpdateResult WebScheduler::update(JsonObject & root, WebScheduler & webSche
 }
 
 // execute scheduled command
-bool WebSchedulerService::command(std::string cmd, std::string data) {
+bool WebSchedulerService::command(const char * cmd, const char * data) {
     StaticJsonDocument<EMSESP_JSON_SIZE_SMALL> doc_input;
     JsonObject                                 input = doc_input.to<JsonObject>();
-    if (data.length()) {
+    if (strlen(data)) { // empty data queries a value
         input["data"] = data;
     }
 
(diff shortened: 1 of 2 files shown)


proddy:
I don't personally see the need for Boot up either. The use case came from https://github.com/emsesp/EMS-ESP32/discussions/1065

proddy:
I closed #127 because there was no interest from the original author for 15 months

src/web/WebSchedulerService.cpp
```diff
@@ -84,75 +83,107 @@ StateUpdateResult WebScheduler::update(JsonObject & root, WebScheduler & webSche
...
+        for (ScheduleItem & scheduleItem : *scheduleItems) {
+            if (scheduleItem.active && scheduleItem.flags == SCHEDULEFLAG_SCHEDULE_TIMER && scheduleItem.elapsed_min == 0) {
+                scheduleItem.retry_cnt = command(scheduleItem.cmd.c_str(), scheduleItem.value.c_str()) ? 0xFF : 0;
+            }
+        }
+        last_tm_min = 0; // startup done, now use for RTC
+    }
+    // check timer every minute, sync to EMS-ESP clock
+    uint32_t uptime_min = uuid::get_uptime_sec() / 60;
+    if (last_uptime_min != uptime_min || last_uptime_min == -1) {
```
proddy, line 157: last_uptime_min is unsigned so this will always be false

MichaelDvP: Ups, yes, can be removed, left from a first try to put startup and timer in one for().

Issue #701, opened 2022-10-25
Scheduler


HerrRiebmann:
First of all: Thank you very much for this incredible project!

### Idea
A Scheduler to set a specific value at several predefined times.

### Implementation
```
- [x] Scheduler Active

- [x] Monday
- [x] Tuesday
- [x] Wednesday
- [x] Thursday
- [x] Friday
- [ ] Saturday
- [ ] Sunday

- [x] 07:30 `thermostat` `{"cmd":"hc1/mode", "data":"day"}`
- [x] 17:30 `thermostat` `{"cmd":"hc1/mode", "data":"night"}`
```

Or some kind of script-configuration: weekday;minutesOfDay;DeviceType;Cmd,Data
```
1;450;1;hc1/mode;day
1;1050;1;hc1/mode;night
2;450;1;hc1/mode;day
2;1050;1;hc1/mode;night
5;450;1;hc1/mode;day
5;960;1;hc1/mode;night
```

### Background
My Buderus Thermostat has about 6 different automatic-programs. None of them is suitable for a workingspace.
I´d just like to change the thermostat `hc1/mode` from **_day_** to **_night_** and vice versa, on five days a week.
Since the ESP is running all day long and I´m currently not planning to additionally introduce another (_constantly running_) device, I thought this could be realized within this great project.
As saied, the EMS-ESP is used in my office, where no Home Assistant will take place.

### Symbiosis
With this approach, it would also be possible to lower or raise the temperature. Change the state of the heatpump. Disable the warmwater cycle, and much more.

### Alternatives
Installing [Mosquitto](https://mosquitto.org/), [NodeRed](https://nodered.org/) and a [time scheduler](https://flows.nodered.org/node/node-red-contrib-ui-time-scheduler/in/590bc13ff3a5f005c7d2189bbb563976), which needs a Computer, Raspberry PI, Synology NAS or whatever.

Using any home automation.

Or manually changing the state with [HTTP Shortcuts](https://http-shortcuts.rmy.ch/) on Android.
(_My current approach_)

### Similar Request
Adding a scheduler for setting GPIOs was requested here:
https://github.com/emsesp/EMS-ESP32/issues/648
But I´m still hoping for a simple solution on a single device.

I´d love to see something like that, adjustable from the web interface!
In this huge project and without any experience in react, it´s hard to contribute this feature myself.

proddy:
[...]

PR #1103, opened 2023-03-05
#701 scheduler web dialog updates

proddy:


proddy:
still need to look at the backend C code to make sure id's are generated in sequence and unique

src/web/WebSchedulerService.cpp
```diff
@@ -37,14 +37,14 @@ void WebSchedulerService::begin() {
...
-        JsonObject si     = schedule.createNestedObject();
-        si["id"]          = scheduleItem.id; // name, is unqiue
-        si["active"]      = scheduleItem.active;
-        si["flags"]       = scheduleItem.flags;
-        si["time"]        = scheduleItem.time;
-        si["cmd"]         = scheduleItem.cmd;
-        si["value"]       = scheduleItem.value;
-        si["description"] = scheduleItem.description;
+        JsonObject si = schedule.createNestedObject();
+        si["id"]      = scheduleItem.id; // name, is unqiue
```
MichaelDvP, line 41: The id is undefined here, go back to `smallitoa(counter)`, or skip id here and generate all unique ids in webpage on load.

proddy: yes, that's next to do. see the previous comment

proddy: fixed.

MichaelDvP (approved):
There are a lot of small changes i don't understand in all consequences, but all seem to work well.
I liked the checkicon more than the circle, but it's also ok.

proddy:
> There are a lot of small changes i don't understand in all consequences, but all seem to work well. I liked the checkicon more than the circle, but it's also ok.

yeah i know what you mean. There's always a debate on how to show active/no-active states. Like on https://ux.stackexchange.com/questions/100504/how-should-i-show-disabled-state-in-table

The problem with a CheckIcon/Tick the user will think it's actually selectable, when it isn't. 

I'll make a chage to use round colored circles and ticks

PR #1105, opened 2023-03-06
scheduler and fixes

MichaelDvP:
These are the scheduler api-commands for info/values/commands
Also when checking for domoticz i've found a bug i've made when merging the avty from pswid. 
And the climate creations had a buffer overflow.


proddy:
thanks Michael, you found a lot of fixes.

Issue #1112, opened 2023-03-10
Optimize Web libraries to reduce bundle size (3.6.x)


proddy:
Look into replacing or optimizing the web components to reduce the amount of Flash used. In 3.6.0 we moved to vite/yarn which has better webpack bundling that helped reduce the size of the web code, but there are some components that don't fully support tree shaking and taking a large amount of the precious ram. 

Running `yarn run build-hosted` will generate a `stats.html` file in the `interface` folder showing the usage. 

![image](https://user-images.githubusercontent.com/1230712/224304525-bc0dd076-b2e8-4573-8e0e-b85a1cb656d2.png)

one culprit is notistack which takes 12.78KB gzipped. Maybe replace with https://github.com/fkhadra/react-toastify

proddy:
latest PR:

replace notistack with react-toast
removing the bold font (was only used in teble headers)
change minify to terser
add HTML minifying as a backup
other minor fixes
in total saves 25.5KB in flash

proddy:
running `yarn run build-hosted` creates the rollup visualizer file in `.../interface/stats.html` which I've been using to find the larger libraries and find alternatives. For example I replaced the Slider in the Web Log Settings to a pulldown menu which saves 5KB.

Next big ones are:

async-validator = 8.36KB
msgpack = 10.44
react-dropzone = 12.26
axios = 29.26 (https://alova.js.org/ is an interesting alternative)

proddy:
async-validator - I looked at 5 other alternatives and this one had the smallest bundle size, so sticking with it.
msgpack - there are many alternatives, as JS libraries. The best one I think is `msgpackr` which supports tree-shaking and is very fast and actively maintained but I couldn't get it to work since it doesn't accept `ArrayBuffer` needed by axios
react-dropzone - I looked at alternatives and this was the smallest, so keeping it
axios - I'm playing with alova to see if I can port the code over. Hopefully this could save 10KB

proddy:
i stole unpack() from https://github.com/kriszyp/msgpackr and made it a typescript function as the library didn't do tree shaking

proddy:
I'm going to look at replacing axios with https://alova.js.org . Then we can close this issue

proddy:
I've spent the last 3 weeks porting axios to alova...almost there. Just need to fix the upload code and it's done. This will save approx 4KB in flash, a lot less than I had hoped but still enough to use the latest framework-arduinoespressif32

proddy:
[...]

PR #1114, opened 2023-03-11
#1112 - web optimizations

proddy:
Make the Web pack smaller by
- replace notistack with react-toast
- removing the bold font (was only used in teble headers)
- change minify to terser
- add HTML minifying as a backup
- other minor fixes

in total saves 25.5KB in flash


interface/src/project/SettingsCustomization.tsx
```diff
@@ -507,61 +503,59 @@ const SettingsCustomization: FC = () => {
...
+                            type="deleted"
+                            isSet={(de.m & DeviceEntityMask.DV_DELETED) === DeviceEntityMask.DV_DELETED}
+                          />
+                        </ToggleButton>
+                      </ToggleButtonGroup>
                     </Cell>
-                    <Cell>{!deviceEntity && formatName(de)}</Cell>
-                    <Cell>{!deviceEntity && !(de.m & DeviceEntityMask.DV_READONLY) && formatValue(de.mi)}</Cell>
-                    <Cell>{!deviceEntity && !(de.m & DeviceEntityMask.DV_READONLY) && formatValue(de.ma)}</Cell>
-                    <Cell>{!deviceEntity && formatValue(de.v)}</Cell>
```
MichaelDvP, line 564: On my old laptop it takes ~5 seconds for each character input in dialog if the list is rendered in background. With empty list i can just type a new name without waiting.

MichaelDvP (commented):
Tried to compile and upload, but can not access the webpage, body of all pages seems empty.
Will test with a new branch...

interface/src/project/SettingsCustomization.tsx
proddy: ok, I wondered why that clause was in there. I'll see if I can optimize it

MichaelDvP: Not rendering the table jumps to top of table after editing, the trick was to render the table, but not the content of the cells. Then scolling stays at the same place.

proddy: i know, but I'd like to code it properly. the biggest problem with React is that the DOM and all its children are re-rendered when a state changes (in this case the deviceEntity). The solution is to use a pattern like https://www.developerway.com/posts/react-re-renders-guide for the Modals where the props and states are separated. Some coding to do....

MichaelDvP:
Now it works, After reboot/upload the snackbar is red but says "Upload successfull".

proddy:
> Now it works, After reboot/upload the snackbar is red but says "Upload successfull". Te difference is ~17kB, mainly the size of the font, zipped 11kB. But it gives a little room for v3.5.1.

fixed that. The last dev with notistack v3 used 2026865 bytes, so its around 25K if you count that version upgrade. Still not bad for 60 mins of work

PR #1124, opened 2023-03-15
fix device icons in dashboard

MichaelDvP:


interface/src/project/DeviceIcon.tsx
```diff
@@ -17,6 +17,7 @@ const enum DeviceType {
   SYSTEM = 0,
   DALLASSENSOR,
   ANALOGSENSOR,
+  SCHEDULER,
```
proddy, line 20: I don't understand this - what was broken? Scheduler is not a device type

MichaelDvP: It's a enum and must match the emsdevice.h enum, the icons are shifted by one before, e.g boiler shows up with thermostat icon.
Scheduler is listed as device because commands/api check for valid device from this list, system, analog, dallas are also no ems-devices, but devices for api/command.

proddy: ok I see. Then it's needed. 

But later we should refactor the code. Scheduler is not a device or a device-type and if we want to extend with API calls it should have it's own end-point.

PR #1137, opened 2023-03-22
improve view of settings pages (app and mqtt)

pswid:
The appearance of the configuration pages has been unified. Additionally on very narrow screens (e.g. smartphones) only one parameter per row is displayed (on wider screens up to three per row).


proddy:
really nice and responsive, thanks for the PR

Issue #1126, opened 2023-03-16
support numeric IDs in react-table-library web tables


proddy:
The latest react-table-library v4.1.0 supports typescript generics for the id, so in tables we don't need to always use Strings. This issue is to see where we had to force using string IDs and revert them back to Numbers.

proddy:
think I've covered all of them in 3.6.0-dev.10

PR #1139, opened 2023-03-23
use number as table id's - #1126

proddy:


proddy:
I used numeric IDs where possible in the Table. Makes no real difference.

interface/src/project/SettingsScheduler.tsx
```diff
@@ -52,15 +52,7 @@ import { useI18nContext } from 'i18n/i18n-react';
...
-  let result = '';
-  const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
-  const charactersLength = characters.length;
-  let counter = 0;
-  while (counter < 4) {
-    result += characters.charAt(Math.floor(Math.random() * charactersLength));
-    counter += 1;
-  }
-  return result;
+  return Math.floor(Math.random() * (Math.floor(200) - 100) + 100);
```
MichaelDvP, line 55: This seems a bit strange to me: floor(200) is always 200, but i think it willalways generate a number between 100 and 200.

MichaelDvP (approved):

interface/src/project/SettingsScheduler.tsx
proddy: yes I wanted a number between 100 and 200. Math.random() returns a floating point. I would probably work without the floor(200). I should test it in a small program and use 'node prog.js' to see the output.

MichaelDvP: I think `Math.floor(Math.random() * 100 + 100)` is the same, or `Math.floor(Math.random() * 100) + 100`

PR #1140, opened 2023-03-23
Dev

Vikingn:
Index.ts Norwegian translated


proddy:
think something went wrong here.

PR #1143, opened 2023-03-24
Norweian translated index.ts

Vikingn:
Norwegian Translated index.ts


Vikingn:
Hope this is ok now !

PR #1145, opened 2023-03-25
revert another wrong translation

MichaelDvP:


Vikingn:
Its just o remove: se innkommende telegrammer in line 62

se innkommende telegrammer is right in line 63

PR #1149, opened 2023-03-27
fixed PL and NO translation of "haclimate"

pswid:


Vikingn:
Should it also be some more lines around line 216 in the document for thesee languages? sv,pl,no,tr

PR #1150, opened 2023-03-27
Fix #1147 - Exhaust temperature always zero on GB125/MC110/RC310

### LINKED ISSUE #1147: Exhaust temperature always zero on GB125/MC110/RC310
Hi,

I noticed that the exhaust gas temperature reported by EMS-ESP32 is always zero on my system while the RC310 shows the actual value. Th3M3 notes (https://github.com/Th3M3/buderus_ems-wiki/blob/master/Quelle_08.md) align with the data I found on my system and I implemented a now tested fix which can be found here: https://github.com/minusdreidb/EMS-ESP32/commit/982a43b8d438e3d2926bd038f4c971139702b66b.

The location at offset 6 in telegram 0xE5 is always zero on my system causing the initial problem, therefore I commented it out.
I don't have telegram 0x19 which also seems to carry the exhaust temp. according to boiler.cpp.

What is the right strategy to handle this situation (assuming that offset 6 in 0xE5 is a known good location for the exhaust temperature and not a bug)?

My system setup including SW-Versions is documented in the above linked commit's message.

minusdreidb:
Fix always zero exhaust temperature reading by adding location telegram 0xe4, offset 31 as an additional source for the exhaust temp. and update from 0xe5 only if the value is unequal 0.

### DIFF
--- a/src/devices/boiler.cpp
+++ b/src/devices/boiler.cpp
@@ -1098,7 +1098,13 @@ void Boiler::process_UBAMonitorSlowPlus(std::shared_ptr<const Telegram> telegram
     has_bitupdate(telegram, ignWork_, 2, 3);
     has_bitupdate(telegram, heatingPump_, 2, 5);
     has_bitupdate(telegram, wwCirc_, 2, 7);
-    //has_update(telegram, exhaustTemp_, 6);
+    
+    /* Update exhaust temperature if it is unequal 0 */
+    const uint8_t exhaustTempOffset{6};
+    uint16_t exhaustTemp{0};
+    if (telegram->read_value(exhaustTemp, exhaustTempOffset))
+        has_update(telegram, exhaustTemp_, exhaustTempOffset);
+    
     has_update(telegram, burnStarts_, 10, 3);   // force to 3 bytes
     has_update(telegram, burnWorkMin_, 13, 3);  // force to 3 bytes
     has_update(telegram, burn2WorkMin_, 16, 3); // force to 3 bytes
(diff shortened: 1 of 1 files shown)


MichaelDvP (commented):
@proddy I'm not sure what is the right way to handle it. The exhaustTemp in E4 is now verified by @minusdreidb . The temp in E5 was just a guess and is maybe something different. 
I think it's better to remove the reading in E5 and check again if users have the E5 value and can verify what it is. We can leave a comment in E5 with reference to issue #1147.
If we really want read both, @minusdreidb should add the same logic to E4-telegram to avoid toggling between 0 and value. But if both telegrams report a nonzero value we also have toggling.

proddy:
> @proddy I'm not sure what is the right way to handle it. The exhaustTemp in E4 is now verified by @minusdreidb . The temp in E5 was just a guess and is maybe something different. I think it's better to remove the reading in E5 and check again if users have the E5 value and can verify what it is. We can leave a comment in E5 with reference to issue #1147. If we really want read both, @minusdreidb should add the same logic to E4-telegram to avoid toggling between 0 and value. But if both telegrams report a nonzero value we also have toggling.

I agree. I don't really like conditional statements in the processing of the EMS telegrams. Let's remove E5 and see who starts complaining. @minusdreidb are you ok to adjust this PR?

minusdreidb:
@proddy I made the changes and updated the PR.

proddy:
ok with me. @MichaelDvP  ?

MichaelDvP:
> ok with me. @MichaelDvP ?

Yes.

Issue #1148, opened 2023-03-26
Power entity


Olgierd-1:
Is it possible to add a boiler entity of actual power in kW?
Now I can see the modulation of the burner (in %).
I know that it is very simple to calculate but it would be usefull to have information about actual power of the heat source.

I have e32 gateway with latest firmware 3.5.1
It is connected to Bosch Condens 9000i boiler.

Regards
Olgierd

proddy:
It's possible, but goes against the design on EMS-ESP which is to read and process the raw data from the EMS bus and pass it on to more sophisticated automation systems (Home Assistant, IOBroker, etc) to do the calculations.  Currently we don't have any custom entities other than a special class to calculate if the hot-water/dhw is running (for the shower alert). There was some discussion earlier to make custom telegrams (https://github.com/emsesp/EMS-ESP32/issues/922).

Olgierd-1:
Thank you for your reply.
Can you check if the value of the current boiler output is available in the data on the EMS bus.
This information depends on two properties:
1. the current burner modulation
2. the maximum boiler output, which is a property of the specific boiler model. The boiler output is stored in the boiler settings by the boiler manufacturer.
Therefore, I presume that the value of the current boiler power should be available for reading from the EMS bus.

proddy:
I haven't been able to find the max KW. Even on my Dutch Nefit TrendLine HRC 30 which is rated for 30Kw the front-panel only shows the % burner power and not the KW.

Vikingn:
Same here   on boiler system pressure ,exhaust temperature , input 1 options , input 2 options , input 4 options 
aux heater max limit  is always 0 I think.
But I have a Bosch 6000AV with just electricity.  But when it find this elements should it not also give values ?

Olgierd-1:
The front panel of my boiler shows the % burner modulation and value of the maximum boiler output (kW) only.
Yes, you are right I can not find the the current boiler output. I finded this entity in the Bosch thermostat integration in the Home Assistant. That is why I thought it was possible to add such an entity to the EMS-ESP.

proddy:
Ah good, that means the kW is somewhere hidden in one of the telegrams most likely. We need to go hunting again!

proddy:
The kw can be calculated, possibly, using the new Telegram entity customization. But it's still experimental so wait a while.

Olgierd-1:
Good news. Thank you.

MichaelDvP:
[...]

PR #1151, opened 2023-03-31
Custom telegram handler #1079

MichaelDvP:
and #1148 

This is my appoach for this. But the 4MB symmetric is again on 99,4% flash. 
I've added environments to platformio.ini for 4MB with asymmetric partitons with loader in the smaller one. These are using platform 6.1.0.
I also added write to the bus-entities (simply not working if the telegram is not writable, but everyone can try).
HA-discovery only adds sensor without write.


proddy:
its a big change. I'll take a look tonight. Does this mean the dev won't fit into 4MB anymore so people still on older dev builds and not using the boot partition, the upload will not work?

MichaelDvP:
It fits on esp32 with 4kB free flash. Web-upload works. The esp32_4M env is unchanged.

(For future change there is now a env:esp32_4M+ with partion check/usb-upload for a second smaller loader partition. This uses platform 6.1)

PR #1159, opened 2023-04-15
add writeable setting to custom entities

MichaelDvP:
and fix the add dialog with empty values (string/number mismatch)
still fit's/uploads in 4MB standard partitions


proddy:
I'll merge. I've been spending the last weeks deep into React memo() and callbacks to reduce the rendering of the web pages and had made changes to the SettingsEntity which I'll add later. I found a better way to keep the type and device as numbers (but still input as hex strings) and a more efficient library for validation. Hopefully I'll get around to finishing it.

Issue #1116, opened 2023-03-12
Optimize WebUI rendering when using Dialog Boxes


proddy:
When a Dialog (Modal) is open in the WebUI (e.g. editing an entity, renaming a customization) the whole screen is still refreshed which is slow and inefficient. It's because of how React renders the nested children in the DOM due to the state changes. The correct way is to enclose the state changes in a separate React component. 

https://www.developerway.com/posts/react-re-renders-guide

proddy:
dialogs in the settings pages done in 3.6.0-dev10

proddy:
Whilst working on this I moved Sensors (Temperature & Analog) to it's own tab, so not in DeviceData anymore. It makes it easier to navigate and the code is much cleaner.

and...I renamed dallas to temperature also in the C code to be consistent with the front-end

proddy:
in dev11

PR #1162, opened 2023-04-22
web render update - #1116

proddy:
lots of minor refactoring to improve performance in the Web. There was a lot of re-rendering when inputting data into a modal dialog box. See https://github.com/emsesp/EMS-ESP32/issues/1116. 
The big change is making separate react component classes for the dialog modals, preventing excessive rendering of the parent classes. it should improve performance.

Note:
It does break backward compatibility with 3.6.0-dev-9 with Entity Customization as the payload format has changed. The device id and type id are now numbers (not strings). And the JSON and endpoints are called `entities` to keep with the standard naming convention.


MichaelDvP:
Wow, 120 files changed! Thats a bit to much to review all the code. I've only tested.
The rendering is really fast now. Only things i've seen:
- system log now have a update button, but this button is always there. Better use dirty flags and show it as `Apply changes()`.
Another thing, old, but maybe need some rework: the manage user page shows the update button outlined and always, after changing a user-password you have to press the update button. Also here the `Apply changes()` button on demand will be more clear.

proddy:
> Wow, 120 files changed! Thats a bit to much to review all the code. I've only tested. The rendering is really fast now. Only things i've seen:
> 
> * system log now have a update button, but this button is always there. Better use dirty flags and show it as `Apply changes()`.
> 
> Another thing, old, but maybe need some rework: the manage user page shows the update button outlined and always, after changing a user-password you have to press the update button. Also here the `Apply changes()` button on demand will be more clear.

yes, I think every single file was touched because of new ESLint rules, one of which re-orders the imports via `yarn run lint-fixall` does that. The main changes are in the Settings*.tsx and the way the Dialogs work. So I wouldn't bother doing a code review, maybe check that the Custom Entities still works as I changed the type of device_id and type_id (they were strings first to accept hex numbers in the TextField input. 

I'll work on the DashboardData next as this is also really inefficient when you have many entities showing on the screen.

Good call on the System Log Update button and also adding the Apply Updates to the User management screens, thanks.

interface/src/project/SettingsEntitiesDialog.tsx
```diff
@@ -0,0 +1,235 @@
...
+              margin="normal"
+              fullWidth
+              type="number"
+              value={editItem.offset}
+              onChange={updateFormValue}
+            />
+          </Grid>
+          <Grid item xs={4}>
+            <ValidatedTextField
+              name="val_type"
```
MichaelDvP, line 160: name should be also `value_type`

MichaelDvP (approved):
Works good. I've changed a lot of settings, but maybe not all.

interface/src/project/SettingsEntitiesDialog.tsx
proddy: fixed

interface/src/project/SettingsEntities.tsx
```diff
@@ -129,147 +108,135 @@ const SettingsEntities: FC = () => {
...
     }
-    setEntityItem(undefined);
+  };
+
+  const addEntityItem = () => {
+    setCreating(true);
+    setSelectedEntityItem({
+      id: Math.floor(Math.random() * (Math.floor(200) - 100) + 100),
+      name: '',
+      device_id: 11,
```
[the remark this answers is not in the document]
proddy: fixed

MichaelDvP:
The update button in weblog does not work proper. I get a Offset is out of bounds error. Also the blocker is not shown, navigation is blocked while the update button shows, but the banner is missing.

proddy:
Ok. I'll fix.

proddy:
> Ok. I'll fix.

fixed.

proddy:
Think it's mostly done now. Now I need to test it.

proddy:
@MichaelDvP I think its working ok. Let me know what you think. Noticeable changes are speed in the web pages, Sensors have moved to a new tab and I made the Device Data scrollable.

MichaelDvP:
Some first impressions:
- Speed is fine. After updating packages with react-router 6.11 routing is broken, i have to go back to 6.10.
- Renaming dallas to temperaturesensors breaks mqtt and api (including @tp1de iobroker adapter). Not sure if this is really worth it.
- Sensors in the new tab is ok (i like the old style on one page more, maybe with some time...)
- entities in the scrollbox is not good, with some more devices the box is completly vanished, or only half a line in height:
Screenshot 2023-05-01 at 13-09-07 EMS-ESP
- Remove the `<Grid> from `edit dialog: 
- Screenshot 2023-05-01 at 13-22-03 EMS-ESP
- Edit dialog is only string input, number input with up/down arrows is missing (step should be string again, float is not working properly)
- Update button in weblog: when saving there should be a `setOrigData(data)`to allow changing setting back. 

to be continued...

proddy:
this PR is getting big so I'm planning to push the changes in a few days, which means different UI and the rename of dallas.

PR #1201, opened 2023-07-02
Merge dev2

MichaelDvP:
only a few manual edits nedded to merge, mainly mqtt.
It compiles, but i havn't tested.
The standalone build still needs some changes for the espmqttclient.


interface/src/project/DashboardDevicesDialog.tsx
proddy, line 1: I optimized this dialog code when doing my alova port so need to check whether these changes are still valid.

proddy: I'll merge and check afterwards

MichaelDvP: This is the change for number input of custom entities without UOM. It does not conflict with your changes.

proddy: ok! cool.

PR #1202, opened 2023-07-02
github action

MichaelDvP:
I named it "test build".
Also tried to fix the standalone, but i never get `pio run -e standalone -t exec` to run without throwing 1000 errors and give up to analyse the cause.


proddy:
I fixed the standalone

PR #1203, opened 2023-07-03
reset dirty when saving

MichaelDvP:


proddy:
that was on my list to fix (again - as I keep missing it out). Thanks Michael

PR #1211, opened 2023-07-08
report on invalid extension - fixes #1210

### LINKED ISSUE #1210: Validate file name extension on file upload
As reported by @swa72 when you upload a file with unrecognized extension (not bin, json or md) it will report back an Unauthorized HTTP error. It's better to use something like '406 Not Acceptable' and have the web code pick this up and report back. The upload should also be immediately stopped (Upload.abort()).

proddy:


lib/framework/UploadFileService.cpp
```diff
@@ -122,18 +124,17 @@ void UploadFileService::uploadComplete(AsyncWebServerRequest * request) {
...
         root["md5"]         = md5;
         response->setLength();
         request->send(response);
-        // AsyncWebServerResponse * response = request->beginResponse(201, "text/plain", md5); // created
-        // request->send(response);
         return;
     }
 
-    handleError(request, 403); // send the forbidden response
+    handleError(request, 406); // send the forbidden response
```
MichaelDvP, line 137: You should also change the comment, forbidden is 403, 406 is not acceptable

lib/framework/UploadFileService.cpp
```diff
@@ -143,9 +144,9 @@ void UploadFileService::handleError(AsyncWebServerRequest * request, int code) {
     }
 
     // send the error code to the client and record the error code in the temp object
-    request->_tempObject              = new int(code);
     AsyncWebServerResponse * response = request->beginResponse(code);
     request->send(response);
+    handleEarlyDisconnect();
```
MichaelDvP, line 149: maybe here also `request->client()->close();` to stop the browser uploading. `handleError`is also called for wrong bin target, and uploading a bin takes a while.

MichaelDvP (approved):

lib/framework/UploadFileService.cpp
proddy: yes, will change

lib/framework/UploadFileService.cpp
proddy: I wanted to send a response code back first, gracefully, but when calling client()->close() it disconnects quickly and the whole response is lost. So I had to work around it and catch the network error in the web code.

I also noticed the upload is slower than with axios. I'll log an issue with the alova team to see what is causing this, along with a benchmark.

MichaelDvP: > and the whole response is lost

Seems my computer or network is to slow, i always get the response. I also dont see a slower upload speed. But for most people uploading is very rare, so it does not matter if it takes a few seconds more.

Issue #1224, opened 2023-07-12
replace react with preact


proddy:
Inspired by https://github.com/rjwats/esp8266-react/issues/327 I thought I'd have another go at trying out preact. preact is a lightweight version of react, with a much smaller bundle size and ideal for slow connections where loading speed could be a problem. There are drawbacks obviously (see https://www.merixstudio.com/blog/what-preact-and-when-should-you-consider-using-it) but for our project, I think it's safe and worth experimenting with. 

Using the compatibility mode (`preact/compat`) and the vite plugin `preact/preset-vite` I only needed to replace `react()` with `preact()` in the `vite.config.ts`. 

It saves 36KB in Flash so now we're able to use the espressif IDF4.4.4/pio 6.3.2 library on an 2MB ESP32. Pretty impressive for changing one line.

Thanks @rjjrbatarao :-)

PR #1225, opened 2023-07-12
replace react with preact #1224

proddy:


MichaelDvP (approved):
Have only compiled and tested, works good, no issues found.

proddy:
I'm running on my production system and so far no issues. The nice thing is that its also free'd up 4KB of heap somehow. But that could be because I'm on 6.3.2 espressif and no longer 5.3.0. 

So TLS should now in theory work on all platform builds?

MichaelDvP:
> So TLS should now in theory work on all platform builds?

Maybe fit to 2M partition, but i think TSL takes to much heap for esp32 without PSRAM.

proddy:
ok. let's keep dev2 running for a few more days and then merge into dev, then a week or two and then into main. Then I can go on holiday (with my new aprilia)

proddy:
nice! RSV Mille Factory R. It was semi-road legal but I put a race exhaust on it

PR #1231, opened 2023-07-23
fix mqtt `on_message`

MichaelDvP:


src/mqtt.cpp
```diff
@@ -211,18 +211,15 @@ void Mqtt::incoming(const char * topic, const char * payload) {
...
-    // This causes a compile error "error: variable-sized object may not be initialized"
-    // so commenting out and assuming with the new mqtt lib it's always null terminated
-
-    /*
-    // sometimes the payload is not terminated correctly, so make a copy
+void Mqtt::on_message(const char * topic, const char * payload, size_t len) const {
+    // the payload is not terminated correctly, so make a copy
     // convert payload to a null-terminated char string
-    char message[len + 2] = {'\0'};
+    char message[len + 1];
```
proddy, line 217: I removed this because the compiler cannot determine the size (error variable-sized object may not be initialized). Is it necessary? If so, use memcpy()

MichaelDvP: You had mentioned this compiler error in the comment, so i removed the initializer.
It s necessary to have a valid termination to the string, mqtt only use a array of uint8_t, not a terminated string.
It is not necessary to initialize the char[], i have removed the initializer and added the terminator in case of nullptr.
(I don't think the nullptr check is necessary, empty payload should have length zero, but valid array pointer. So the strlcpy should also give a terminated char[] which is "\0". But have not tested yet, so leave the nullptr check)

I don't think this PR will give a compiler error.

Issue #1076, opened 2023-02-21
Turkish language support


proddy:
### Discussed in https://github.com/emsesp/EMS-ESP32/discussions/868

<div type='discussions-op-text'>

<sup>Originally posted by **dreixin** December 31, 2022</sup>
I am trying to add Turkish language support to EMS-ESP. Anyone willing to help is welcome.</div>

MichaelDvP:
When merging the it-translations i see that there are a lot of tr-translations missing. 
Reopening, maybe someone can help and write into this excel sheet.
[locale_translations_w_it.xlsx](https://github.com/emsesp/EMS-ESP32/files/11929185/locale_translations_w_it.xlsx)

proddy:
@dreixin could you please take a look

dreixin:
Hi everyone, 
As you know we had a big earthquake and personally I had so many other stuff to deal with during this time and I had to leave this project on a lover priority for a long time. attached is a quick touch with all the translations done until line 270.
I will spare more time on the project and try to do my best to finish as soon as possible.
[locale_translations_w_it.xlsx](https://github.com/emsesp/EMS-ESP32/files/11936715/locale_translations_w_it.xlsx)

proddy:
We'll include the ones up to line 270 for now, as I want to release the next version of EMS-ESP soon.

MichaelDvP:
The translations up to line 270 are included in dev since 27.feb. See: https://github.com/emsesp/EMS-ESP32/commit/94749e46154b7cf62129fa170e6b6d3df013375b

proddy:
oh, I thought more were added

dreixin:
[locale_translations_w_it.xlsx](https://github.com/emsesp/EMS-ESP32/files/12206538/locale_translations_w_it.xlsx)
on going process, I hope it will be completed in a week or so.

MichaelDvP:
I've added the new translations to my dev2, you can test here https://github.com/MichaelDvP/EMS-ESP32/releases/tag/test
I'll make a PR when complete.

dreixin:
[locale_translations_w_it_v2.xlsx](https://github.com/emsesp/EMS-ESP32/files/12211665/locale_translations_w_it_v2.xlsx)
last 30-40 translations left

proddy:
We'll add the remaining translations later in a minor release of 3.6.x

dreixin:
Finished all the translations on Turkish. 
Attached is the final file. 
[...]

PR #1237, opened 2023-07-31
more tr translations #1076

MichaelDvP:
BTW: better remove the test release, Now dev is newer and it may confuse the useres.


proddy:
> BTW: better remove the test release, Now dev is newer and it may confuse the useres.

ok, do we need to remove the GH actions too?

MichaelDvP:
> do we need to remove the GH actions too?

Why not leave it, if you push a dev2 again it will create a new test build.

Issue #1240, opened 2023-08-01
Enhance version check to differentiate between ESP32 and ESP32-S3 bin versions


proddy:
### Discussed in https://github.com/emsesp/EMS-ESP32/discussions/1238

<div type='discussions-op-text'>

<sup>Originally posted by **tp1de** August  1, 2023</sup>
The System / Version Check from web-ui is a nice function. 
I am currently using the S3 version. The download functionality does not  differentiate between the esp32 and the esp32-s3 bin file. 

Since @bbqkees is working on a S3 version of his gateway, it might be a good idea to download the correct bin file from web-ui. 
Version 3.6. should be productive soon. I would recommend to add this functionality. (3.6 might stay for a couple of months)

@MichaelDvP @proddy  - What's your opinion?</div>

proddy:
I have a PR coming

PR #1245, opened 2023-08-04
Enhance version check to differentiate between ESP32 and ESP32-S3 bin versions #1240

proddy:
Haven't tested on real ESP32 variants yet


MichaelDvP:
Tested on esp32S3 and esp32, working.
For the changelog it is maybe better to redirect to the download page, instead of downloading the md-file. I think most people does not have a reader/editor associated with *.md and don't know how to open.

proddy:
> Tested on esp32S3 and esp32, working. For the changelog it is maybe better to redirect to the download page, instead of downloading the md-file. I think most people does not have a reader/editor associated with *.md and don't know how to open.

thanks for testing.

I couldn't find a way to show the md file in the browser (using the raw URL or installing another plugin). Any ideas here or shall we just leave it?

MichaelDvP:
We can leave it, but i think https://github.com/emsesp/EMS-ESP32/blob/dev/CHANGELOG_LATEST.md shows the formated dev changelog and https://github.com/emsesp/EMS-ESP32/blob/main/CHANGELOG.md the release. But it's not the file from the release asset and may be different (normally not).

platformio.ini
```diff
@@ -75,8 +74,8 @@ build_unflags = ${common.unbuild_flags}
 
 ; for github Actions actually standard platform is buggy, use dev-platform
 [env:ci_s3]
+extends = espressi32_base
 platform = https://github.com/platformio/platform-espressif32.git
```
MichaelDvP, line 78: You need to remove this line 78, it overwrites the platform from the base.

proddy:
merged by accident, I still need to put in direct links to the md files

Issue #1267, opened 2023-08-26
Add Shower Duration to MQTT Discovery topic


proddy:
Add the shower_duration as an MQTT topic which makes it easier to add as an HA entity

(from the discussion https://github.com/emsesp/EMS-ESP32/discussions/1248)

proddy:
See https://emsesp.github.io/docs/Home-Assistant/#shower-integration

proddy:
Added a new command called boiler/coldshot which can be called via the API and HA

rene865:
can you see what i am doing wrong? I have just started with home assistant and yaml.
EMS-ESP Version 3.7.0-dev.0

i added your example in configuration.yaml, at first I thought the code here was wrong _ems_esp_shower_DATA_duration but what I try it doesn't work.
![image](https://github.com/emsesp/EMS-ESP32/assets/127300660/7a082797-813e-43ce-89c1-ff3ad33cb9ed)
![image](https://github.com/emsesp/EMS-ESP32/assets/127300660/1366405e-eec6-4cf6-a501-4da02699fe07)

but for the new entity the shower duration is not availible
![image](https://github.com/emsesp/EMS-ESP32/assets/127300660/027c02ac-5f1f-4240-8582-81f693b0fbfc)

proddy:
Sorry! I hadn't made the build yet. Doing it now.

proddy:
> Sorry! I hadn't made the build yet. Doing it now.

https://github.com/emsesp/EMS-ESP32/releases/tag/latest

PR #1272, opened 2023-08-29
Shower re-work - #1267

proddy:


MichaelDvP (approved):
Looks good to me. I have a buffered dhw system, now flowsensor, so can't do a real life test.

Issue #1271, opened 2023-08-29
Show Wifi SSID and strength (RSSI) in Network Status web page


proddy:
Network Status page is missing some important data, like the SSID and Wifi Strength.

MichaelDvP:
added in my dev1d (SSID was there, only RSSI missing)

MichaelDvP:
BTW: Should we color the icon to show wifi quality, <-85dBm => red, <-75dBm orange, else green (because of #1264)?

proddy:
excellent idea, let me know if you need help

MichaelDvP:
added to PR #1284

PR #1273, opened 2023-08-31
update Mqtt Client

MichaelDvP:
Bert has fixed the issues in mqtt. But i leave the default mem level in the library and added a check in mqtt.c. Let's hope that this was the issue from #1264 
#1271 is included


src/web/WebStatusService.cpp
```diff
@@ -166,9 +166,9 @@ void WebStatusService::webStatusService(AsyncWebServerRequest * request) {
     if (Mqtt::enabled()) {
         statJson       = statsJson.createNestedObject();
         statJson["id"] = 5;
-        statJson["s"]  = Mqtt::publish_count();
+        statJson["s"]  = Mqtt::publish_count() - Mqtt::publish_fails();
```
proddy, line 169: nice!

proddy (approved):

proddy:
really great work hunting down the root cause for this freeze/restarts Michael.

proddy:
And @MichaelDvP I think we need to make a v3.6.1 soon, based on what we have in dev

src/web/WebStatusService.cpp
MichaelDvP: statJson["s"] is for success, publish_count() is all messages. Same correction for quality calculation.

MichaelDvP:
> And @MichaelDvP I think we need to make a v3.6.1 soon, based on what we have in dev

Yes.

PR #1284, opened 2023-09-06
fix for espressif@6.4.0

MichaelDvP:
and stay on 6.3.2 for the esp32 envs


proddy:
@MichaelDvP espressif 6.5.0 is out, but I had to add a ` -D CONFIG_ETH_ENABLED` to get it to compile, otherwise it would complain about ETH.h not finding WiFi.h

just for your information if you stumble across the same thing

MichaelDvP:
Thanks. Seems they have also added W5500 and ETH_SPI to the ETHClass.

Issue #1297, opened 2023-09-17
/api/custom show {} if no Custom Entities configured instead of an error


proddy:
### Discussed in https://github.com/emsesp/EMS-ESP32/discussions/1294

<div type='discussions-op-text'>

<sup>Originally posted by **harvey637** September 14, 2023</sup>
Hi,
the iobroker adapter for ems-esp requested until v 2.4.0 for "custom" entities,
I guess, that the resposnse to that request somehow changed with the "new" 3.7.0-dev0,
at least I did not spot the logging of this error before.
In the ems-esp-iobroker-adapter this was remedied with introducing v2.4.1 with the switch to turn on/off the call for custom entities.
https://github.com/tp1de/ioBroker.ems-esp/issues/34

May be it is a better way to answer the api call to not existing entities (like custom entities, analogue sensors, ...) with "INFO" and the answer "nothing configured for <api-call customs|sensor|analogue|...>" instead of errors.

just my idea - no problem today, as v2.4.1 has this configurable.

Thank You+++
Harvey</div>

proddy:
So the change here is that when you have no Custom Entities configured, a call to http://<hostname>/api/custom/ should return an empty JSON {} and not log an Error with the message "unknown device".

proddy:
did a PR with this change (3.6.2-dev-1). Now if there are no custom entities it'll come back with a message 'no entities'.

PR #1298, opened 2023-09-17
some refactoring and change #1297

proddy:
In the C++ code we were using class and object names like 'customEntity' in the Customizations logic, which was getting confused with the new Custom Entity feature. So I renamed the classes, objects, descriptors so you can tell if its Customization or CustomEntity piece of code.


MichaelDvP (approved):
It's to much to read/check every change, i just merged and checked on a esp. Looks ok.
The "no entries" message should also go to scheduler, temperaturesensor and analogsensor. I have this in my dev2-test, so we can add later with the power entities.
Only the analogsensor have a enable setting (is this usefull?), so it should give error message if disabled and "no entries" if enabled.

proddy:
good catch, I forgot about the scheduler, temperaturesensor and analogsensor and it should have consistent behaviour. I'll merge now so you can continue with your updates.

PR #1302, opened 2023-09-20
network connection and powerentities

MichaelDvP:
it's some more commits now. includes
- powerentities, setting of 32bit values and max-limit to 32 bit, save to NVS
- network, bssid and full scan
- tr language additions, some web translationsstill missing
- mqtt queue check
- a mqtt connect change from bert melis i havn't merged before
- changed handling of flag-svg in webpage due to package update
- empty json for a/t-sensors, custom and scheduler, fix the `api/device/commands` for custom/scheduler
- the missing dhw tags for some boiler values


proddy:
some of the changes are amazing @MichaelDvP - you clearly spent a lot of time on this PR

proddy:
@MichaelDvP in `boiler::process_UBAFactory()` we're only reading `nomPower` at offset 4. Do we also need to add the min and max at offsets 5 and 6? I'm missing these.

image

MichaelDvP:
Do we need them as entities? 
We can read them and set these values as min/max to burnMinPower/burnMaxPower.

proddy:
my burnminpower and burnmaxpower are 0 and 50 respectively. Not 23 and 115, so I thought they were different entity values.

MichaelDvP:
> my burnminpower and burnmaxpower are 0 and 50 respectively. Not 23 and 115, so I thought they were different entity values.

Yes, but i think you can't set burnMaxPower lower than 23 or higher than 115. 

Btw there seems to be a bug in min/max settings in customizations if min is <0 like in hysteresis the max is always set avery high value. Have not found the cause yet.

PR #1304, opened 2023-09-21
burner settings use min/max from telegram 4, fix max for values < 0

MichaelDvP:
This sets the min max limits of burner settings to the values read from telegram 4.


proddy:
tested, works well. thanks!

PR #1306, opened 2023-09-23
small changes

proddy:
@MichaelDvP could you look at the TODO in `thermostat.cpp` 

    // TODO: should this be >= 87 to avoid out of bounds?
    if (telegram->message_length + telegram->offset >= 92 && telegram->offset <= 87) {
        char data[sizeof(wwVacation_)];


src/devices/thermostat.cpp
```diff
@@ -843,6 +843,8 @@ void Thermostat::process_RC35wwTimer(std::shared_ptr<const Telegram> telegram) {
     if (telegram->type_id != 0x38) {
         return;
     }
+
+    // TODO: should this be >= 87 to avoid out of bounds?
```
MichaelDvP, line 847: No, this is the right check, we make sure that all offsets 87-92 are in the telegram. (e.g. read 10 38 90 gives offsets 90-117, so 87, 88, 89 are missing, o processing)

src/emsdevicevalue.cpp
```diff
@@ -331,7 +331,7 @@ bool DeviceValue::get_custom_min(int16_t & val) {
         if (fahrenheit) {
             v = (v - (32 * (fahrenheit - 1))) / 1.8; // reset to °C
         }
-        if (max > 0 && v > max) {
+        if (max > 0 && v > (int16_t)max) {
```
MichaelDvP, line 334: Could be wrong to shorten max, if you want to compare identical types do 
`if (max > 0 && v > 0 && (uint32_t)v >max) {`

Maybe it's much easier to set max as int32_t and use 31 bit range and handle the inputs for min/max also in int32_t, only shorten min to int16_t for storage.

proddy: thanks, this was just so the code would compile on my MacBook - the compiler would throw an error. I'll go for option 1 or look at a compiler option to ignore these checks.

src/devices/thermostat.cpp
proddy: yes sorry, not sure what I was thinking

PR #1316, opened 2023-10-03
arduino 2.0.13 and 3.0.0 ready

MichaelDvP:
I've merged the changes from jason2866 and updated to 2023.10.02. The platformio now is set to the tasmota development build, for releases we should use a fixed one. In `pio_local.ini_example` the path to the arduino 3 from jason is set.
Also some cleanup: for C3mini use the integrated arduino neopixel instead of adafruit library (tested), The pwm generation ledc is different for idf5, code checks the idf and use the right.
Additions to ventilation and thermostat need to be confirmed by the users.


proddy:
nice work, interesting changes in the async* libs too. I assume this has the wifi flush() fix too. I'm back on the weekend to test and will interested in the memory footprint

PR #1332, opened 2023-10-15
Changes from my testbuild

MichaelDvP:


MichaelDvP:
Or should we also use dev2 as testbuild? S3 is now wthout SSL option, and mqtt takes 5k heap for own task. I hope this will not stay to next release.

proddy:
i think a dev2 is better to hunt down the cause of the freezes

MichaelDvP:
ok, i close this.

proddy:
@MichaelDvP created a dev2. https://github.com/emsesp/EMS-ESP32/tree/dev2

PR #1343, opened 2023-10-17
show entity count in device list, remove type

proddy:
small change. what do you think? Showing the Type in a column on the right didn't make sense as most of the times it was hidden.

image


MichaelDvP:
The entity count is good, but i would leave the type. Yes, it's hidden when showing the entities, but for new users it makes the types more clear when showing up frst time.

MichaelDvP (approved):
A lot of changes, i've just merged and tested in my S3. Looks good.

Issue #1382, opened 2023-11-04
Energy/power entities need to have "diagnostic" as category set in HA to avoid warnings


proddy:
In recent updates of HA, any Energy entities must have the category "ent_cat" set to "diagnostic" and not "config". We have two in EMS-ESP in the Boiler device called `nrgheat` (Energy Heating) and `nrgww` (dhw Energy dhw). Both have a command associated to reset the counters making them automatically set to "config".

The change is in `Mqtt::publish_ha_sensor_config()` to ignore the commands if the UOM is KWH.

tp1de:
As far as I understood the entity category is not needed. It works for energy states as well without.

proddy:
It is required for HA now. The owner of Home Assistant had contacted us asking us to include it.

tp1de:
Thanks for info. I will include  in my own ha entities as well.

proddy:
Paulus (HA creator) emailed @bbqkees and asked if we could support this. "A configuration entity provides a setting/configuration option for the device, while a diagnostic entity provides information about the device.". So when devices are added by Discovery all the config sections are consistent.

codetrembler:
After upgrading the EMS-Gateway to 3.6.3-dev5b and Home Assistant 2023.11.1 with HA Auto-Dicovery there are still some values missing for me:

sensor.thermostat_datetime
sensor.thermostat_hc1_holidays
sensor.thermostat_hc1_switchtime1
sensor.thermostat_hc1_switchtime2
sensor.thermostat_hc1_vacations
sensor.thermostat_wwholidays
sensor.thermostat_wwswitchtime
sensor.thermostat_wwvacations
sensor.thermostat_wwcircswitchtime

sensor.boiler_maintenancedate

sensor.ems_esp_shower_duration
sensor.ems_esp_shower_timestamp

Could you please check if the category for these entities has to be adapted too,
Since I only have one HC I would think that the HC2-values are also missing but cannot check this with my systems.

Thank you very much and keep up the good work!

tp1de:
sensor.thermostat_datetime is now text.thermostat_datetime for me. I do not have witchtimes or holidays. Maybe they are text as well?

proddy:
[...]

PR #1383, opened 2023-11-04
no commands for nrgheat & nrgww - #1382

proddy:
Changed on line

```C
    // determine if we're creating the command topics which we use special HA configs
    // unless the entity has been marked as read-only and so it'll default to using the sensor/ type
    // or we're dealing with Energy sensors that must have "diagnostic" as an entity category (e.g. negheat & nrgww)
    bool has_cmd = dv.has_cmd && !dv.has_state(DeviceValueState::DV_READONLY) && (dv.uom != DeviceValueUOM::KWH);
```

to ignore commands sent to HA for UOMs with KWH. There are only 2 so I think it's safe


MichaelDvP (commented):
Is this complete, only removing kWh? As i understand all with `set_ha_classes` are ' diagnostic'

proddy:
it works, I think but the code can be tidied up. I see now the workaround that was added now for DeviceValueType::ULON . I'll change a few things.

MichaelDvP (approved):
If it works i think this is ok until HA change it again.

PR #1418, opened 2023-11-12
show UOM in console when 'show values'. And added a new system command called 'values' which dumps the same to JSON

proddy:
- had to move some functions around to be in scope
- fixed the show UOM function that didn't work anymore since we added the shortnames in brackets. This was breaking 'show values' in the Console
- the system command 'values' will dump all values (just name + value) in a JSON which is handy for a dump. I think. Anyway, there is no memory management so it may blow up with users having a shit load of entities
-


src/emsesp.cpp
```diff
@@ -386,13 +386,19 @@ void EMSESP::show_device_values(uuid::console::Shell & shell) {
...
-                    const char * key = p.key().c_str();
-                    shell.printf("  %s: ", key);
+                    std::string key = p.key().c_str(); // this will be fullname and the shortname in brackets
+                    shell.printf("  %s: ", key.c_str());
                     JsonVariant data = p.value();
                     shell.print(COLOR_BRIGHT_GREEN);
                     shell.print(data.as<std::string>());
+
+                    // extract the shortname from the key, which is in brackets
+                    std::size_t first_bracket = key.find_first_of('(');
```
MichaelDvP, line 396: We have some long names with brackets lke `temperature (TS1)` or `valve (VS2)`, better check for last open/close brackets.

src/system.cpp
```diff
@@ -106,6 +106,35 @@ bool System::command_response(const char * value, const int8_t id, JsonObject &
     return true;
 }
 
+// output all the EMS devices and their values, plus the sensors and any custom things
+// basically dump everything we know about
```
MichaelDvP, line 110: Nice, but the other `device/values` commands uses the shortnames (`info` the long names). `system` is always a bit special, so it's ok.

src/system.cpp
```diff
@@ -106,6 +106,35 @@ bool System::command_response(const char * value, const int8_t id, JsonObject &
...
+                std::string title = emsdevice->device_type_2_device_name_translated() + std::string(" ") + emsdevice->to_string();
+                device_output     = output.createNestedObject(title);
+                emsesp::EMSdevice::export_values(emsdevice->device_type(), device_output, id, EMSdevice::OUTPUT_TARGET::API_VERBOSE);
+            }
+        }
+    }
+
+    // Custom entities
+    device_output = output.createNestedObject("Custom Entities");
+    EMSESP::webCustomEntityService.get_value_info(device_output, "");
```
MichaelDvP, line 127: add `webSchedulerService` here.

MichaelDvP (approved):

src/emsesp.cpp
proddy: Thanks for reviewing. I'll see if I can scan backwards for the last bracket as you suggested

src/system.cpp
proddy: hmm, that's a good point. So how about calling the system command 'allinfo' instead of 'values' ?

Or stick to `values` and show the shortname, which is less readable, and not translated, but still consistent.

src/system.cpp
proddy: I decided to leave the Scheduler out since it doesn't have any entity values - they are just records.

Maybe add a new system command that shows the scheduler details? That's useful since the current 'call scheduler info' only shows the name and whether its on/off, which isn't that useful.

MichaelDvP: The scheduler have entity/command to activate/deactivate the schedule. This was for replace the `forceheatingoff`, sending `selflowtemp 0` every minute to stop the boiler, and disable scheduler to activate the boiler again.
Maybe it's a nice addition to also manipulate the schedule command value via command. But difficult to add in the command structure. 

I wonder if we can add custom RAM entities only stored in ems-esp and use them for setting a variable value. E.g. in custom use devcie/telegram type 0x0 for storing 256 bytes. Custom entity: `myseltemp, device 0, type 0, offset 0, UINT`, Scheduler: `every minute, command: boiler/seltemp, value: custom/myseltemp`. Then the flowtemp can be set via mqtt: custom/myseltemp and is used by the scheduler.

src/emsesp.cpp
proddy: fixed.

src/system.cpp
proddy: I see, That's quite clever. How about instead of using custom entities to store variables we could create a new type called Variables. Maybe limit to 10 max, as strings with a default value that is stored in LittleFS and loaded at boot.

src/system.cpp
proddy: made it allvalues

src/system.cpp
proddy: moved to https://github.com/emsesp/EMS-ESP32/issues/1423

PR #1422, opened 2023-11-13
Add options to invert RX and TX and to enable pullup on RX

ellepdesk:
These option can be individually enabled at compile time by the following defines:

EMSUART_RX_INVERT
EMSUART_TX_INVERT
EMSUART_RX_PULLUP


MichaelDvP:
> @MichaelDvP you good with this?

Yes, maybe better than the approach with negative gpio numbers. I think RX_PULLUP is not needed, but do not hurt. Afaik the uart sets an internal pullup when configuring.

ellepdesk:
Hmm, I'll go test if the pullup is already on by default (make sense to me), no need to add this if it is already there

proddy:
We just need to document it now in https://github.com/emsesp/docs/blob/main/docs/FAQ.md

PR #1426, opened 2023-11-14
fix bug when traversing ems devices when using cmd allvalues

proddy:
use emsdevice->uniqueid for displaying values instead of searching the factory for registered types. when displaying more than 1 thermostat it got messed up

### DIFF
--- a/src/emsdevice.cpp
+++ b/src/emsdevice.cpp
@@ -817,7 +817,7 @@ std::string EMSdevice::get_value_uom(const std::string & shortname) const {
     return std::string{}; // not found
 }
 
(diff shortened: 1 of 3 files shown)


MichaelDvP (approved):

MichaelDvP:
I'm not sure if this is this commit, but the devices are mixed now, api/thermostat/values gives me the mixer values, etc.

MichaelDvP:
Oh, now i see, i've missed that: 
It is intended that all thermostats go to **one** publish, each thermostats uses different tags/hcs and there are no identical values.
Also we have one thermostat_data, mixer_data,, etc.  in mqtt. Checking the device_type is right.

proddy:
yes, I see as well. Missed that. I'll revert back

proddy:
will do a new PR

PR #1428, opened 2023-11-15
latest changes

MichaelDvP:
This includes all of your changes to dev and all of my dev2, except the new water/pool devices.
New entites for water related solar module are not included in this build.
There sholud be no merge conflicts also to dev build.


MichaelDvP:
> shall I test and merge into dev and call this 3.6.3 ?

Yes. Then we can add the braking changes with water module to the dev2/test. I think as 3.6.4-test, there are only few people with mixer/solar as water module so for most users there is no breaking change.

proddy:
done

PR #1430, opened 2023-11-15
fix for export_values and new command_allvalues()

proddy:
fixes https://github.com/emsesp/EMS-ESP32/pull/1426


proddy:
hopefully working now

proddy:
pushing this as its a major bug and don't want too many people to start complaining

PR #1438, opened 2023-11-18
fix crash on wrong thermostat mode command (if not using HA)

MichaelDvP:
Too late for the main, but not big issue.


proddy:
@MichaelDvP amazing how you traced that bug down to that one line!

PR #1451, opened 2023-11-22
next fix RC300 mode

MichaelDvP:
another patch for the main. Sorry, I wonder why this was not recognized in the test builds.


proddy:
no worries, I'll update 3.6.3

I wanted to call it 3.6.3.x but my semver version checker was built to only support major.minor.patch

PR #1455, opened 2023-11-23
fix boiler commands to ems telegrams

MichaelDvP:


proddy:
@MichaelDvP thanks - this was a critical bug, I had problems also setting boiler values. Shall I do a 3.6.4 now?

MichaelDvP:
Yes, better count up to 3.6.4 and declare all 3.6.3 as broken.

proddy:
just saw that message on tab-routing. damn.

proddy:
i have fix, will do it tonight after work

MichaelDvP:
The 3.6.5-dev0 bin have the routing bug, To upload new software, don't use the tab, only manual URL input: http://ems-esp.local/system/upload

PR #1461, opened 2023-11-24
revert to react-router-dom 6.19.0 to fix tab-routing-issue

MichaelDvP:


proddy:
thanks @MichaelDvP for the quick fix. The looping bug is gone when navigating tabs and the other bug I'm still working on. This is where the tabs are selected (a blue bar is shown under the tab in the webui)

PR #1464, opened 2023-11-26
revert to react-router-dom 6.18.0 to bring blue tab line back

MichaelDvP:
tested some older build to see what makes it disappear.


proddy:
I was looking at the routing logic, which I wrote a while back and it's not great - so today I'll spend some time to refactor it and hopefully this will fix some of those issues we've been seeing

PR #1465, opened 2023-11-26
fix router loops

proddy:
not perfect, but it fixes the problem

### DIFF
--- a/interface/package.json
+++ b/interface/package.json
@@ -43,7 +43,7 @@
     "react-dom": "latest",
     "react-dropzone": "^14.2.3",
     "react-icons": "^4.12.0",
-    "react-router-dom": "6.20.0",
+    "react-router-dom": "^6.20.0",
     "react-toastify": "^9.1.3",
     "sockette": "^2.0.6",
     "typesafe-i18n": "^5.26.2",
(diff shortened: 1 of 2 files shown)


interface/package.json
```diff
@@ -43,7 +43,7 @@
     "react-dom": "latest",
     "react-dropzone": "^14.2.3",
     "react-icons": "^4.12.0",
-    "react-router-dom": "6.18.0",
+    "react-router-dom": "6.20.0",
```
MichaelDvP, line 46: you can use ^ again=> `"^6.20.0"`

MichaelDvP (approved):
works good.

proddy:
dom-router reverted there changes in 6.20.1 so we'll do the same

MichaelDvP:
The absolute path works with 6.20.0 and 6.20.1. Is there any advantage in using relative pathes?

proddy:
> The absolute path works with 6.20.0 and 6.20.1. Is there any advantage in using relative pathes?

not really, it just saves a few bytes of flash and makes the code a little easier to read. I was hoping the route matching would work again and catch invalid paths, so preventing the browser warnings when going to for example /dashboard/devices222. but it didn't. I can easily go back to the last implementation, its just a few seconds of work

PR #1471, opened 2023-11-29
Some new/fixed entities

MichaelDvP:
Also sending the step as string, this works for me for RC35 internal offset temperature 0.1 step.
The entities are from my dev2, for some i get feedback that it's working, for others no feedback.

### DIFF
--- a/CHANGELOG_LATEST.md
+++ b/CHANGELOG_LATEST.md
@@ -6,7 +6,12 @@
 
 ## Added
 
+- thermostat boost mode and boost time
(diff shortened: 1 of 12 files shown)


proddy:
strange that steps as numerical didn't work for everyone. perhaps the decimal separator.

MichaelDvP:
IMO it is a float issue: in float 1/10 is not 0.1 and it seems step rounds this float to zero. Step 0.5 works as float.
I think with sending double the numerical steps should work.

PR #1493, opened 2023-12-10
update standalone server to bun and itty-router, add back router paths, add support download button

proddy:


MichaelDvP (approved):
Have not checked the mock-api changes, need to setup wsl for bun on windows first.
Support info buttons are good. Maybe use filetype .json only for the settings that can be uploaded and for support/allvalues .txt?

proddy:
> Have not checked the mock-api changes, need to setup wsl for bun on windows first. Support info buttons are good. Maybe use filetype .json only for the settings that can be uploaded and for support/allvalues .txt? And use first letter upper case for the `Remember ...`

added

PR #1496, opened 2023-12-11
Update locale_translations.h

misa1515:
added new language - slovak


proddy:
nice, that was quite a lot of work to do. Are you also planning to translate the web interface?

And you'll need to add the language to the Web settings. See https://emsesp.github.io/docs/Adding-Languages/#how-to-add-language-xy-to-the-webui

MichaelDvP:
I've tried to merge to a testbuild. Platformio seems to hang up on the characters `č` and 'ď', not finding some includes after that and stops compiling. Maybe a charset setting, but i have not found a solution yet. The other special characters do not have this effect.
BTW: We also should include the special characters in webfont.

misa1515:
@proddy one question: how to create new directory in interface/src/ "sk"?

proddy:
I can create it for you and all you need to do is update the index.ts file

proddy:
can you merge to dev please?

PR #1497, opened 2023-12-12
add Tasmota/Arduino ver. and CPU info to System Status

pswid:


proddy:
no problems with this PR. @MichaelDvP ?

MichaelDvP:
It's ok for me, but maybe there is a better solution for the untranslated `based on`. E.g. set primary=SDK / IDF and show only with separator?

proddy:
@pswid thanks for the PR. If you remove the 'based on' translation text we're good to go. It'll save on flash mem too.

MichaelDvP:
BTW: i've made something similar for the arduino 3.0 tests and the strange reboots to wrong partition. But i found it to 'techi' for normal users and the wrong boot seems fixed. (In my build, not translated, not in system info, just for check before updating)
grafik

proddy:
> BTW: i've made something similar for the arduino 3.0 tests and the strange reboots to wrong partition. But i found it to 'techi' for normal users and the wrong boot seems fixed. (In my build, not translated, not in system info, just for check before updating) grafik

I like that - can we merge the two changes and then close this PR?

MichaelDvP:
How about this?
- show for SDK: arduino core and IDF
- show for CPU: type, rev, cores and frequency 
- show in Application the partitions name and used/free space
grafik

proddy:
> How about this?
> 
> * show for SDK: arduino core and IDF
> * show for CPU: type, rev, cores and frequency
> * show in Application the partitions name and used/free space
>   grafik

perfect!

pswid:
> @pswid thanks for the PR. If you remove the 'based on' translation text we're good to go. It'll save on flash mem too.

I apologize for the late response. I wasn't able to react earlier.

My idea was to show the names of the SDKs in a form identical to the official releases. See e.g.:
https://github.com/espressif/arduino-esp32/releases/tag/2.0.14
https://github.com/tasmota/platform-espressif32/releases/tag/2023.10.03

If you think this is a bad approach, the "based on" can of course be removed. Should I do it or will you change it yourself?

MichaelDvP:
> Should I do it or will you change it yourself?

I have already merged your PR to my development and make some small changes to show partition, etc as mentioned. Also testing with with different cores, e.g. arduino 3.0 have a different definition of `ESP_ARDUNO_VERSION_STR` that conflicts, so better rename it here.
I'll make a new PR with your and my changes. Check if you are ok with it.

pswid:
Great. Of course I have nothing against it.

PR #1499, opened 2023-12-13
latest entities

MichaelDvP:
Here are the changes for some issues


src/devices/boiler.cpp
```diff
@@ -1309,7 +1322,7 @@ void Boiler::process_UBAMonitorFastPlus(std::shared_ptr<const Telegram> telegram
     uint8_t syspress = sysPress_;
     telegram->read_value(syspress, 21); // 0 means no sensor
     if (syspress == 0) {
-        sysPress_ = EMS_VALUE_UINT_NOTSET;
```
proddy, line 1312: wow!

PR #1514, opened 2023-12-17
ethernet with arduino_3.0, sort translations alphabetical

MichaelDvP:
Playing around with BBQKees' new gateway i get now arduino 3.0 working with ethernet. 

Also sort the translations. The entity translation is missing? 
Do you have the same compile errors (include not found) when translations contains the special sk-chars ď or č?
Hmm, seems to be a windows issue. I pushed the translations to a branch dev_sk, it does not compile in VScode on windows here, but github action runs fine.

The boiler wwmode seems still not working, but the included fix  is logcal for the reverse order in enum.


proddy:
I don't have compile errors with those strange characters, I do everything in WSL on Windows, only because its like twice as fast to build than pio on native Windows.

Great that you got the Ethernet working. Which IDF 5.x are you using? Would be good to add to te platform.ini file for future reference. Are you using Tasmota still? (https://github.com/Jason2866/platform-espressif32/tree/Arduino/IDF5)

MichaelDvP:
I'll try with wsl. Do you have tested with the sk translations in `locale_translations-h`? It's not in the dev. You can check with https://github.com/MichaelDvP/EMS-ESP32/blob/dev_sk/src/locale_translations.h
But i think it's a windows issue because GH-actions work.

For Arduino 3.0 i've used `platform = https://github.com/Jason2866/platform-espressif32.git#Arduino/IDF5` which is synced with 
`platform = https://github.com/tasmota/platform-espressif32.git#Arduino/IDF5`
The SSL is missing, so the mqtt library and S3 needs some small mods to compile.

proddy:
Just tried on wsl and MacOSX and both compile without warnings. I've added the translations.h file to the latest dev, forgot that one.

proddy:
and if you're switching to WSL for vscode and pio I can give you some tips on the setup, like getting the USB to work

MichaelDvP:
>  I can give you some tips on the setup,

Yes, please. I have now just opend a wsl commandline, installed nodejs and platformio and started yarn and pio from commandlne. Works, but integration in vscode is nicer. 
BTW: On my old (ivy bridge) latop the wsl build is much slower than windows build in vscode, i think wsl needs too much resources.

proddy:
I'm using Ubuntu as the WSL2 distro
```sh
❯ uname -a
Linux Desktop 5.15.133.1-microsoft-standard-WSL2 #1 SMP Thu Oct 5 21:02:42 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

❯ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 23.04
Release:        23.04
Codename:       lunar
```

I also use Windows Terminal (Version: 1.18.3181.0) for accessing the WSL shell.

And indeed, it takes up a lot of memory. The process VmmemWSL loads at 1000MB and can easily climb to 3000MB after lots of VSCode compiling. If you want to turn it off use `wsl --shutdown` from Windows-R Run command.

I'm not sure how familiar you are with Linux, but use 
```sh
sudo apt update
sudo apt upgrade
```
to keep the Linux system up to date.

To install bun use `curl -fsSL https://bun.sh/install | bash`, which is need for our standalone.

To get USB to work:

1. Go to a windows Admin cmd window
2. `usbipd list` to find the attached USB. Note the <bus-id>. See if its Shared
3. `usbipd bind --busid <bus-id>`
4. In your WSL2 shell window you'll now see `/dev/ttyUSB0` which you can use in pio.ini
5. the changes are persisted over a boot so it only needs to be done once

(see https://github.com/dorssel/usbipd-win/wiki/WSL-support)

MichaelDvP:
Thanks, 
btw: usbipd is in win11, in win10 it has to be installed from github.
I think i found the issue in wndows, pio system info gives for windows:
```
Platform                    Windows-10
File System Encoding        utf-8
Locale Encoding             cp1252
```
and for wsl:
```
Platform                    Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
File System Encoding        utf-8
Locale Encoding             UTF-8
```
I'm now searching how to change the local encoding to utf-8.

MichaelDvP:
Just to have i documented if we search for it:
Found, check the utf-8 box in region settings:
grafik
then pio system info gives:
```
Platform                    Windows-10
File System Encoding        utf-8
Locale Encoding             cp65001
```
and compilng with sk translations works.

proddy:
I got fed up with loosing the USB all the time I went back to my old way of getting USB in WSL. Download and run this program called 'WSL USB' from https://gitlab.com/alelec/wsl-usb-gui. When you start it you choose with ports to mount. And it works. 

The only thing you need to do on the WSL side, and only once, is run this command:

```
sudo apt install linux-tools-virtual hwdata
sudo update-alternatives --install /usr/local/bin/usbip usbip `ls /usr/lib/linux-tools/*/usbip | tail -n1` 20
```

Issue #1516, opened 2023-12-19
MQTT: Home assistant discovery - analog sensor invalid dev section


arturzx:
### PROBLEM DESCRIPTION

Invalid dev section for analog sensor in HA discovery leads to HA displaying all sensors prefixed with analog sensor name.

### TO REPRODUCE

Add analog sensor.

### EXPECTED BEHAVIOUR

Added analog sensor have separated device ids like others and not leads to change other entities names.

### SCREENSHOTS

![image](https://github.com/emsesp/EMS-ESP32/assets/42235373/76f0bc02-feba-43d1-8179-89876a5e6238)
![image](https://github.com/emsesp/EMS-ESP32/assets/42235373/b8b78a5b-7c6f-452b-b478-693160194e3a)

### ADDITIONAL CONTEXT

Discovery topic value for this sensor
`{
  "stat_t": "ems-esp/analogsensor_data",
  "val_tpl": "{{value_json['05'].value if value_json['05'] is defined and value_json['05'].value is defined else wyłączono}}",
  "obj_id": "analogsensor_05",
  "uniq_id": "analogsensor_05",
  "name": "CWU potrzeba grzania",
  "pl_on": "włączono",
  "pl_off": "wyłączono",
  "dev": {
    "name": "CWU potrzeba grzania",
    "ids": [
      "ems-esp"
    ]
  },
  "avty": [
    {
      "t": "ems-esp/status",
      "val_tpl": "{{'online' if value == 'online' else 'offline'}}"
    },
    {
      "t": "ems-esp/analogsensor_data",
      "val_tpl": "{{'online' if value_json['05'] is defined and value_json['05'].value is defined else 'offline'}}"
    }
  ],
  "avty_mode": "all"
}`

arturzx:
Screenshot from HA devices list:
![image](https://github.com/emsesp/EMS-ESP32/assets/42235373/107397d1-a2ef-4891-ab1a-2ff9bd7135ff)

proddy:
See #1528

proddy:
will fix with 1528

PR #1517, opened 2023-12-19
fix: Fixed dev section in analog sensor HA discovery

arturzx:
Fixing #1516 filled by me a few minutes ago. This commit changes generating dev section similiar to mixer or boiler entities - dev name is "ems-esp-analog" and name is "ems-esp Analog". All entities from analog sensors shows in HA with this prefix by default.

### DIFF
--- a/src/analogsensor.cpp
+++ b/src/analogsensor.cpp
@@ -116,7 +116,7 @@ void AnalogSensor::reload() {
             }
             if (!found) {
                 sensors_.emplace_back(sensor.gpio, sensor.name, sensor.offset, sensor.factor, sensor.uom, sensor.type);
-                sensors_.back().ha_registered = false; // this will trigger recrate of the HA config
+                sensors_.back().ha_registered = false; // this will trigger recreate of the HA config
                 if (sensor.type == AnalogType::COUNTER || sensor.type >= AnalogType::DIGITAL_OUT) {
                     sensors_.back().set_value(sensor.offset);
                 } else {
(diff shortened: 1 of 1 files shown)


proddy:
That was quick. I'll take a look and see if I can first reproduce. And also using the translation for Analog in the name, or is that not important you think ?

arturzx:
I've made this change like for other entities. For name generation in mqtt.cpp is used device_type_2_device_name:
image

I've suppose that method returns normal name because there are second device_type_2_device_name_translated.

I don't think that is very important - this is only device name. If it were to be changed, it would be good to unify it across all devices.

MichaelDvP:
Maybe a good time for checking the other ems-esp internal devices and make all fit in one scheme. See:
https://github.com/emsesp/EMS-ESP32/blob/08c5fd8f64c19ff7b76b672a2914825d51758381/src/temperaturesensor.cpp#L547
https://github.com/emsesp/EMS-ESP32/blob/08c5fd8f64c19ff7b76b672a2914825d51758381/src/web/WebSchedulerService.cpp#L291
https://github.com/emsesp/EMS-ESP32/blob/08c5fd8f64c19ff7b76b672a2914825d51758381/src/web/WebCustomEntityService.cpp#L430
https://github.com/emsesp/EMS-ESP32/blob/08c5fd8f64c19ff7b76b672a2914825d51758381/src/shower.cpp#L214
https://github.com/emsesp/EMS-ESP32/blob/08c5fd8f64c19ff7b76b672a2914825d51758381/src/shower.cpp#L244

proddy:
yes, lets clean up the whole thing

proddy:
added the rest in dev.7

PR #1523, opened 2023-12-22
Polish translation update

pswid:


proddy:
thanks @pswid . I'll review.

Issue #1538, opened 2024-01-04
Upgrade ArduinoJson 6 to 7


proddy:
ArduinoJson 7 is out. Would be interesting to test both overall performance and memory usage. Now that it uses dynamic memory allocation I'm curious what happens with mem fragmentation (Max Alloc).

https://arduinojson.org/v7/how-to/upgrade-from-v6/

proddy:
Added to 3.6.5-dev.8

proddy:
I also removed all the & references from JsonObject and JsonArray as there are already semantically references to JsonDocument. And added JsonObjectConst where necessary. It doesn't affect the code size or performance, perhaps prevent dangling references if they ever happen.

PR #1539, opened 2024-01-04
ArduinoJson version 7.0.0 - #1538

proddy:
don't see any difference at all. perhaps a little more responsive. I'll keep running for a few more days.

code is cleaner
 - no more buffers as all memory is dynamically allocated on the heap


MichaelDvP:
Yes, looks good, much clearer code, also no issues if allocated buffer is too small. Only code size is a bit bigger.

proddy:
i found a few errors where I mixed up `.as<JsonObject>()` and `.to<JsonObject>()` which I'm fixing now. Then I'll optimize String a little to decrease fragmentation and then I think its good to go

Issue #1536, opened 2024-01-04
Unknown Device ID (Bosch C1200W) Gas Boiler


Raybom:
### PROBLEM DESCRIPTION

Using custom build, and my device is Bosch C1200W. It was working okay for one week and after that it did not work. The RX and TX readings are okay but, device ID is unknown.

### REQUESTED INFORMATION

- [x] Searched the problem in [issues](https://github.com/emsesp/EMS-ESP32/issues)
- [x] Searched the problem in [discussions](https://github.com/emsesp/EMS-ESP32/discussions)
- [x] Searched the problem in the [docs](https://emsesp.github.io/docs/Troubleshooting/)
- [x] Searched the problem in the [chat](https://discord.gg/3J3GgnzpyT)
- [x] Provide the output of http://ems-esp.local/api/system :

```lua
  System information output here:
[emsesp_settings (2).json](https://github.com/emsesp/EMS-ESP32/files/13827733/emsesp_settings.2.json)

``
{
  "type": "settings",
  "System": {
    "version": "3.6.4"
  },
  "Network": {
    "ssid": "",
    "bssid": "",
    "password": "",
    "hostname": "ems-esp",
    "static_ip_config": false,
    "enableIPv6": false,
    "bandwidth20": false,
    "tx_power": 20,
    "nosleep": false,
    "enableMDNS": true,
    "enableCORS": false,
    "CORSOrigin": "*"
  },
  "AP": {
    "provision_mode": 1,
    "ssid": "ems-esp",
    "password": "ems-esp-neo",
    "channel": 1,
    "ssid_hidden": false,
    "max_clients": 4,
    "local_ip": "192.168.4.1",
    "gateway_ip": "192.168.4.1",
    "subnet_mask": "255.255.255.0"
  },
  "MQTT": {
    "enabled": false,
    "host": "",
    "port": 1883,
    "base": "ems-esp",
    "username": "",
    "password": "",
    "client_id": "ems-esp",
    "keep_alive": 60,
    "clean_session": false,
    "entity_format": 1,
    "publish_time_boiler": 10,
    "publish_time_thermostat": 10,
    "publish_time_solar": 10,
    "publish_time_mixer": 10,
    "publish_time_other": 10,
    "publish_time_sensor": 10,
    "publish_time_heartbeat": 60,
    "mqtt_qos": 0,
    "mqtt_retain": false,
    "ha_enabled": false,
    "nested_format": 1,
[...]

PR #1542, opened 2024-01-06
fix #1495 and #1536

### LINKED ISSUE #1495: Wrong reading and setting of Dhw comfort mode
The reading and setting of setting of **Dhw comfort mode** is reversed.
When **eco** mode is set on the boiler then on the ems-esp gateway dashboard it is visible as **comfort mode** (see attached screenshots)
Firmware 3.6.4, boiler Junkers GC2300
Switching the mode to **eco** via gateway dashboard result with boiler going into **comfort mode** and vice versa 

Screenshot_20231209_082948_Samsung Internet

MichaelDvP:
1495 seems fixed now.
no feedback from 1535, but no complain is a good sign.


proddy:
I've updated the individual issues too

Issue #1338, opened 2023-10-16
unnecessary dhw prefixes


pswid:
I just noticed that the Energy Dasboard in HomeAssistant no longer shows the energy used by DHW. This is because in version [3.6.2](https://github.com/emsesp/EMS-ESP32/releases/tag/v3.6.2) the names of several entities related to DHW have changed (more DHW tags added). What is this change for? In my opinion, the _statistical_ entities of the entire system (such as the number of starts, operating time, energy consumed and supplied for heating, cooling, pool and dhw) should not contain any prefixes (neither DHW nor HC). They introduce _confusion_ in the nomenclature and spoil the view (see below). 
In addition, HomeAssistant does not have the option of maintaining continuity of energy monitoring of a specific device if the entity name changes (there is no option to copy the history of the old entity to the new one nor merge two entities). So, device name/id modification of long-term recorded entity is a breaking change!

I strongly suggest to revert https://github.com/emsesp/EMS-ESP32/commit/ad81f84c02dcca0a95bc4740b510e2a736fef930 commit.

<img width="386" alt="dhw prefixes" src="https://github.com/emsesp/EMS-ESP32/assets/78219494/4590d8f3-9c65-4608-9ec8-7a379b2dcaba">

proddy:
You have a valid point about the naming. We'll look into it.

Can you share the Energy Dashboard setup screen? I don't think the sensor names have changed use in HA discovery.

MichaelDvP:
Ok, this is the complain we have expected in #1301. I'll remove from the texts what i can identify.

bluegaspode:
this 'dwh prefix change' not only changed wordings in the UI, also the API keys

/api/boiler/info
have changed (breaking my data ingestion to influx for some values).

fields where I see it a change and 'dwh' is used as prefix and suffix.

dhw operating time compressor dhw
dhw energy consumption compressor dhw
dhw aux elec. heater energy consumption dhw
dhw total energy warm supplied dhw
dhw hp dhw
dhw heat limit dhw

all these value have 'siblings' for heating/cooling/pool  where the differenciation happens only via the suffix.

proddy:
@MichaelDvP shall we just revert the change and look again at how to fix the translations?

MichaelDvP:
[...]

Issue #1301, opened 2023-09-20
minor translation issue, missing Warmwater in some names


harvey637:
Hi,

just a reminder:

some (german) translation use the same wording for eg. "total consumption" AND "total consumption warmwater"
... the additional word ".... Verbrauch Warmwasser" would clarify this without checking the internal names. Also this leads to some duplicate german names. With memory constraints this could also be an additional WW :-)

![grafik](https://github.com/emsesp/EMS-ESP32/assets/10156426/52062f99-e4c9-4d81-b397-2ad865f0891f)

thank you
harvey

MichaelDvP:
I think they should be tagged with DHW/WW, added to my testbuild

MichaelDvP:
With the tags added we have some dhw values with double dhw in name now:
![grafik](https://github.com/emsesp/EMS-ESP32/assets/59284019/d60a8669-a4cc-40d9-929b-0adecf4643b1)
Sometimes it also depends on translation.

proddy:
I think that should be fine, unless someone complains

PR #1559, opened 2024-01-14
add tag after fullname instead of before

proddy:
For https://github.com/emsesp/EMS-ESP32/issues/1338

### DIFF
--- a/interface/package.json
+++ b/interface/package.json
@@ -20,20 +20,20 @@
     "lint": "eslint . --cache --fix"
   },
   "dependencies": {
-    "@alova/adapter-xhr": "^1.0.2",
+    "@alova/adapter-xhr": "^1.0.3",
     "@babel/core": "^7.23.7",
     "@emotion/react": "^11.11.3",
     "@emotion/styled": "^11.11.0",
     "@mui/icons-material": "^5.15.4",
-    "@mui/material": "^5.15.4",
+    "@mui/material": "^5.15.5",
     "@table-library/react-table-library": "4.1.7",
     "@types/imagemin": "^8.0.5",
     "@types/lodash-es": "^4.17.12",
-    "@types/node": "^20.11.0",
-    "@types/react": "^18.2.47",
+    "@types/node": "^20.11.5",
+    "@types/react": "^18.2.48",
     "@types/react-dom": "^18.2.18",
     "@types/react-router-dom": "^5.3.3",
-    "alova": "^2.16.2",
+    "alova": "^2.17.0",
     "async-validator": "^4.2.5",
     "history": "^5.3.0",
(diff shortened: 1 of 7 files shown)


proddy:
@MichaelDvP would you mind checking this to see if there is anything broken downstream that I may have missed? I've added TODOs in 3 places to track the changes.

MichaelDvP:
Tested, works so far. It's a bit unfamiliar with tags appended, especally the hc, the commands/mqtt have them in front, but ok. Sorting alphabetical now shows different hcs in line, e.g. all seltemps together. 

But this does not solve #1338. There the tag itself leads to the complain. He dont want have these dhw entities taged. 
For boiler we have all dhw related taged with dhw, but some heatpump entities were introduces withot tag, e.g. 
boiler/wwworkm was taged "dhw active time", heatpumps boiler/uptimecompww "operating time compressor" was not taged but named with fullname "operating time compressor dhw". Because of #1301 i've added the missing dhw tags to have all in line, but this leads to this complain, i think because the tag also is used in HA.  
But i still think it's right to tag these dhw related entities and every change will have someone who complains.

The other complain https://github.com/emsesp/EMS-ESP32/issues/1338#issuecomment-1774154213 is the boiler/info command shows fullnames and will always break his database on a renaming.

proddy:
> Tested, works so far. It's a bit unfamiliar with tags appended, especally the hc, the commands/mqtt have them in front, but ok. Sorting alphabetical now shows different hcs in line, e.g. all seltemps together.
> 
> But this does not solve #1338. There the tag itself leads to the complain. He dont want have these dhw entities taged. For boiler we have all dhw related taged with dhw, but some heatpump entities were introduces withot tag, e.g. boiler/wwworkm was taged "dhw active time", heatpumps boiler/uptimecompww "operating time compressor" was not taged but named with fullname "operating time compressor dhw". Because of #1301 i've added the missing dhw tags to have all in line, but this leads to this complain, i think because the tag also is used in HA. But i still think it's right to tag these dhw related entities and every change will have someone who complains.
> 
> The other complain [#1338 (comment)](https://github.com/emsesp/EMS-ESP32/issues/1338#issuecomment-1774154213) is the boiler/info command shows fullnames and will always break his database on a renaming.

I see. Then we should just leave it be, and not change anything. Unless we want to tag every entity. Shall I revert this PR?

MichaelDvP:
> Shall I revert this PR?

As you like, the change does not make a big difference and there are always some users complaining on any change. 
Actually i prefere the prefix, but i think only because it's familiar.

proddy:
I'll leave it the old way, and we can always make a new Settings option to prefix or suffix the tag - its cosmetics only

Issue #1423, opened 2023-11-13
custom variables


proddy:
From a discussion with Michael:

The scheduler have entity/command to activate/deactivate the schedule. This was for replace the `forceheatingoff`, sending `selflowtemp 0` every minute to stop the boiler, and disable scheduler to activate the boiler again.
Maybe it's a nice addition to also manipulate the schedule command value via command. But difficult to add in the command structure. 

I wonder if we can add custom RAM entities only stored in ems-esp and use them for setting a variable value. E.g. in custom use devcie/telegram type 0x0 for storing 256 bytes. Custom entity: `myseltemp, device 0, type 0, offset 0, UINT`, Scheduler: `every minute, command: boiler/seltemp, value: custom/myseltemp`. Then the flowtemp can be set via mqtt: custom/myseltemp and is used by the scheduler.

_Originally posted by @MichaelDvP in https://github.com/emsesp/EMS-ESP32/pull/1418#discussion_r1390784057_

proddy:
I see, That's quite clever. How about instead of using custom entities to store variables we could create a new type called Variables. Maybe limit to 10 max, as strings with a default value that is stored in LittleFS and loaded at boot.

proddy:
Something like:
- In the Custom Entity web page you can choose to read/write the value from EMS (with a deviceID, typeID, Offset) or as a Variable with default value. It means extending Custom Entity to include a Custom Entity type, and it's persisted in the FS. The value could be written to NVS or just kept in memory (which means loosing after a reboot)
- Add a new API call to read or write a value to the Custom Entity
- Modify the Scheduler to also accept a Custom Entity as a value, instead of a hardcoded value

MichaelDvP:
Comments to:
1. `Custom variable` only with name and default value. The value can be stored as string (no need to set the type), the command-values are always processed as char*. Default value should be stored in FS, changing the value by API should only use RAM (to avoid wear of FS). Setting new default only va web page.
2. API command is already implemented for custom values (`call custom <name> <value>`, `/api/custom/<name> {"value":"<value>"}`)
3. The scheduler already accepts commands (that return a value) as value. e.g. `boiler/flowtemp/value` returns the actual flowtemp can be processed as value parameter.

proddy:
excellent, so we only need to implement #1 !

proddy:
@MichaelDvP my idea with Variables was to be able to use them in Values

[...]

PR #1570, opened 2024-01-20
custom variables #1423

MichaelDvP:


proddy:
It looks ok, I'll need to a deep dive to see how it works later but trust it's coded well like all your PRs.

I'll update the handler.ts for standalone testing and also a section to the docs/wiki explaining how to use it.

one question: what does 'ram' standfor ?

PR #1575, opened 2024-01-22
small changes, optimizations and fixes

proddy:
Lots of little changes
 - refactoring and fixing typos
 - adding more test data for sensors and custom entities
 - added back old express/muter so can test eventstreaming and file uploading (was a bug in Alova)

I'll keep this PR open as I'm not finished. I still want to

- finish the Variables (custom entities) to make them true variables and add some more tests.
- fix bug when Custom Entities web screen is loaded on first render
- remove the info and commands and make it fixed (so for example get_value_info() replaces command_info() in sensors)
- removing all customizations in web prompts a button to reboot


proddy:
I'm now going to clean up the 'info' and 'commands' for the analog and temp sensors so it's consistent. Then I'll merge the PR.

@MichaelDvP FYI

PR #1589, opened 2024-01-29
feat: add 'heatingfailsafe' option

Ian-Clowes:
Some boilers cannot be activated over EMS only, and require that the traditional 230V activation switching is wired to "permanently on" in order to allow EMS control.  EMS messages can then be used to turn the boiler off when no heat is required.  A side-effect of this is that if the EMS messages to the boiler cease for some reason it defaults back to running at full power after a short period of minutes.  In a typical home automation system failure of the controlling PC / uP would result in no HTTP, MQTT, etc messages to EMS-ESP, which in turn would create no messsages to the boiler, leading to the incorrect running.  This change adds a timeout watchdog to EMS-ESP so that if no messages to actively control the boiler heat output are received the boiler will be shut off.


proddy:
@Ian-Clowes this is a really solid PR! I can't fault it. You figured out the main code, how to update the web, the translations and also the test mock-data which is not easy.

@MichaelDvP you approve?

Ian-Clowes:
Good to hear, and very kind.  Thanks.

For clarity, I'm not sure how the changes to these files snuck in.  I didn't make them and they seem to be a result of some of the git commands.  They may need removing / correcting if they are wrongly bringing in snippets from other PRs:

- .github/workflows/github-releases-to-discord.yml
- CHANGELOG.md
- sonar-project.properties

MichaelDvP:
I think it is a dangerous option to switch off heating if wifi is somehow disturbed. In frost periods it could cause freezing of pipes, etc. 
Bosch has a lot of experience with heating and i think there is a reason to fallback to ON on missing thermostat communication.

- We should add a warning that users only activate it, if they really know what they do (this warning box should also include the forceheatingoff option). 
- The default should be OFF, not as now 120 sec!
- `lastSendHeatingOff = uuid::get_uptime_sec();` is removed, so it's never updated and `forceheatingoff` does not work anymore.

Ian-Clowes:
So maybe adding `&& curFlowTemp_ > 50` to the tests?

MichaelDvP:
> The goal is perhaps to do the same as what somehting like the Comfort II RF controller would do if it lost the connection from the remote thermostat. 

I don't know what the boiler do, but i think the same as a broken ems connection: heating to the setting on the frontpanel.
But if you get a RF and test it let us know the result.
> An option might be to inspect the current flow temperature, which appears to remain valid when the boiler is stopped. The failsafe stop could be prevented unless that was over (say) 5C.

Difficult, nofrost mode of thermostats is more complicated, i think there is a logic with pump running periodically, checking flowtemps and start the boiler on too low temperature, let pump running in medium temperature and shut off pump on high temp. But this is done by thermostat, not in boiler.
> I selected the 120s to make the feature on by default because with prior firmware versions I have had crashes that take the unit back to factory settings - wi-fi. etc config lost

Then we should check what causes the reset and the factory reset and fix this, but not a workarund that make new problems.

BTW: I see you only reset the timer on ems-esp commands (set_flowtemp, etc) and overwrite other ems-commands to 0x1A with heating off.
This means for all users with a thermostat that the boiler stops 2 minutes after connection ems-esp( with the default enabled). This will give a lot of complains, i'm sure.  I think you should also reset the timer on 0x1A telegram when the thermostat send a new flowtemp.

Ian-Clowes:
Thanks for the follow up.  It's my first foray into EMS-ESP so I'll look at the 0x1A comments, etc to see if I can cover relevant cases appropriately.

I generally agree that fixing the crash & reset bug is the situation to aim for.  I saw that behaviour a year or so ago, and the current version I'm running has been a lot more stable.  Despite that I'm stil inclined to have this feature dong what the title suggests - provide a failsafe preventing the boiler running permanently if any part of the EMS side glitches significantly.  But I'm happy to compile my local version with the 120s default - so will set it to zero on future PR.

While waiting to get a suitable WB control device to observe its EMS messages to the boiler I asked their tech team what was the impact of suppressing boiler activaiton via EMS.  The reply was:

"The boilers internal frost protection will be active regardless of the types of controls installed and whether there is a demand from EMS controls or voltage on the 230vac central heating demand.

Internal frost protection will protect the appliance based on the internal flow temperature of the main heat exchanger, this will run the pump if the water temperature drops below 7 degrees and fire the burner if it drops below 5 degrees."

So it sounds like the boiler frost protection is not a problem.

PR #1591, opened 2024-01-30
fix #1590 and more checks for device-telegram mapping

### LINKED ISSUE #1590: ems-esp returns an error when sending a broadcast message
Oh, this is a bug. Ems-esp checks reads for the reply and writes for `<ack>/<nack>` from the destination. The broadcast is wrongly handled as write and there is no ack.

_Originally posted by @MichaelDvP in https://github.com/emsesp/EMS-ESP32/issues/1588#issuecomment-1916141156_

MichaelDvP:
and some small changes to analog/temperature sensors to show commands always.


proddy:
thanks for also fixing the 'commands' - missed that

PR #1615, opened 2024-02-11
fix WiFi TxPower, plus a ton of other refactoring

proddy:
added harcoded dBm values into Web's Network Settings page along with an extra setting called Auto which picks the lowest Wifi TX power based on the RSSI value. This is calculated and set after the WiFi has connected (STA); as before, it was never called. The idea is to save on power and reduce interference in noisy areas. Not sure if it's needed or if it works, but I stole the idea from Tasmota and ESPEasy. 

I also fixed a bug where settings did not persist when a Restart was required.

-- EDIT --

I should really have done this in a new PR, but anyway...I've added a few more changes inspired by @kam821

- Fixed uninitialized variable usage (NetworkSettingsService::_stopping).
- Added resource cleaning in MqttSettingsService.
- Mark SecuritySettingsService as final to prevent potential call of virtual function in constructor (wrapRequest).
- BSSID format logic moved to separate function
- Removed std::bind/std::placeholders, use lambda instead
- Removed c style casts (still more to go)
- Use auto in variable initialization from cast/new expression.
- Use braced initialization in return statements.
- Added missing std::move for non-trivial type
- Use const& in range for loop instead of value (User)
- Replace malloc by new expression
- Single argument constructors marked as explicit
- Added missing 'override'
- Removed VLA used in ArduinoJsonJWT::decode
- Wrapped while loop assignment expression in parenthesis
- Wrapped WIFI_RECONNECTION_DELAY in parenthesis
- Changed [&] closure capture to more explicit capture of this parameter.
- Change find_first_of(".") to find_first_of('.')
- Replace String == "" by isEmpty().
- Replace push_back by emplace_back
- Fixed unused parameters warning
- Renamed NTPSettingsService connected_ to _connected to match convention, moved initialization to constructor in CPP file
- Replace chained ternary operators by immediately invoked lambda to improve readability (NTPStatus.cpp)
- Use std::vector instead of std::list that hasn't already been modified
- Reworked some places to change mutable references to be able to use const references in parameters.
- Replace empty destructor by default one
- NULL -> nullptr
- time.h -> ctime
- Replace hand-written array length by sizeof

thanks to @kam821. Changes taken from:

https://github.com/kam821/EMS-ESP32/commit/d265ab8468ead42a04e5c3b1dd0b75845ebe580d
https://github.com/kam821/EMS-ESP32/commit/ed3ea06a56facafd61353ae357601ecf1672ddd4
https://github.com/kam821/EMS-ESP32/commit/2225aa4a4289e28c71617c75592812d79158c5b7


lib/framework/NetworkSettingsService.cpp
```diff
@@ -83,31 +86,364 @@ void NetworkSettingsService::manageSTA() {
...
+        p = WIFI_POWER_13dBm;
+    else if (set_power >= 11)
+        p = WIFI_POWER_11dBm;
+    else if (set_power >= 8)
+        p = WIFI_POWER_8_5dBm;
+    else if (set_power >= 7)
+        p = WIFI_POWER_7dBm;
+    else if (set_power >= 5)
+        p = WIFI_POWER_5dBm;
+
```
MichaelDvP, line 169: This table set a too low value, e.g. if calculated min is 12.5dBm it sets 11, this results in very low connection rates.
It should set 11dBm for calculate of >8.5 <=11, etc.

lib/framework/NetworkSettingsService.cpp
```diff
@@ -83,31 +86,364 @@ void NetworkSettingsService::manageSTA() {
...
+#ifdef EMSESP_DEBUG
+    emsesp::EMSESP::logger().debug("Recommended set WiFi Tx Power (set_power %d, new power %d, rssi %d, threshold %d", set_power, p, rssi, threshold);
+#else
+    char result[10];
+    emsesp::EMSESP::logger().info("Setting WiFi Tx Power to %s dBm", emsesp::Helpers::render_value(result, (double)(p / 4), 1));
```
MichaelDvP, line 174: Wrong bracket, => ((double)p) / 4 to get the x.5 settings

lib/framework/NetworkSettingsService.cpp
```diff
@@ -83,31 +86,364 @@ void NetworkSettingsService::manageSTA() {
...
+                                         info.wifi_sta_disconnected.reason); // IDF 4.0
+        emsesp::EMSESP::system_.has_ipv6(false);
+        break;
+
+    case ARDUINO_EVENT_WIFI_STA_GOT_IP:
+        char result[10];
+        emsesp::EMSESP::logger().info("WiFi connected with IP=%s, hostname=%s, TxPower=%s dBm",
+                                      WiFi.localIP().toString().c_str(),
+                                      WiFi.getHostname(),
+                                      emsesp::Helpers::render_value(result, (double)(WiFi.getTxPower() / 4), 1));
```
MichaelDvP, line 309: Same wrong bracket => ((double)p) / 4 to get the x.5 settings

MichaelDvP (commented):
There is a lot of code moved within the change commits, difficult to find and review the new code. 
The tx-auto seems to give very low results, i get on the router rssi < -70 dbM receive from ems-esp with datarate <10Mbit/s. . log is ` Recommended set WiFi Tx Power (set_power 4, new power 8, rssi -530, threshold -690`
I think we should set the auto to tx-power resulting in ~-65dBm (or higher) in the router to have 150mBit/s connection. (For me this needs ~ 11dBm tx-power ( giving rssi -67dBm, 135 mBit/s).

lib/framework/NetworkSettingsService.cpp
proddy: that would explain the low values

MichaelDvP: Yes, partly. I tested now with:
```
    int threshold  = WIFI_SENSITIVITY_n + 70; // Margin in dBm * 10 on top of threshold
```
(goal -65 dBm -> threshold 650)
and 
```
wifi_power_t p = WIFI_POWER_2dBm;
    if (min_tx_pwr > 185)
        p = WIFI_POWER_19_5dBm;
    else if (min_tx_pwr > 170)
        p = WIFI_POWER_18_5dBm;
    else if (min_tx_pwr > 150)
        p = WIFI_POWER_17dBm;
    else if (min_tx_pwr > 130)
        p = WIFI_POWER_15dBm;
    else if (min_tx_pwr > 110)
        p = WIFI_POWER_13dBm;
    else if (min_tx_pwr > 85)
        p = WIFI_POWER_11dBm;
    else if (min_tx_pwr > 70)
        p = WIFI_POWER_8_5dBm;
    else if (min_tx_pwr > 50)
        p = WIFI_POWER_7dBm;
    else if (min_tx_pwr > 20)
        p = WIFI_POWER_5dBm;
```
(using min_tx_power for the limits with decimals)
now i get tx_power 11 dBm and connection rate of 80-130 MBit/s, so i think lower limit of usefll setting.

lib/framework/NetworkSettingsService.cpp
proddy: fixed.

lib/framework/NetworkSettingsService.cpp
proddy: ok, I'll make those changes and test locally.

proddy:
> I noticed I broke the eventStreaming so will need to look into that

fixed.

proddy:
with the latest changes a few days ago I managed to squeeze out 3-4KB extra heap, now it seems to have gone back. So now I need to trace back in the code what caused it. When that's done I'll merge this PR

Issue #1605, opened 2024-02-04
Expose EMS-ESP commands (like reset) via MQTT Discovery


proddy:
Special commands in EMS-ESP are not created in Home Assistant. We only have one (type `DeviceValueType::CMD`) for the Boiler Reset command. We could treat all CMDs as enums (so lists in Discovery).

proddy:
in dev14 PR

PR #1626, opened 2024-02-20
Expose EMS-ESP commands (like reset) via MQTT Discovery #1605

proddy:
All CMDs are treated as ENUMs

and

Send MQTT heartbeat immediately after connection [#1628](https://github.com/emsesp/EMS-ESP32/issues/1628)


MichaelDvP (commented):
Looks ok to me (but i don't know what HA shows).

proddy:
> Looks ok to me (but i don't know what HA shows).

as a select list in the HA UI 

image

PR #1631, opened 2024-02-23
small fixes

MichaelDvP:


lib/AsyncTCP/src/AsyncTCP.h
```diff
@@ -46,11 +46,20 @@ extern "C" {
...
+#define CONFIG_ASYNC_TCP_USE_WDT 0       //if enabled, adds between 33us and 200us per event
+#endif
+
+#ifndef CONFIG_ASYNC_TCP_TASK_PRIORITY
+#define CONFIG_ASYNC_TCP_TASK_PRIORITY 5
 #endif
 
 #ifndef CONFIG_ASYNC_TCP_STACK_SIZE
-#define CONFIG_ASYNC_TCP_STACK_SIZE 8192 * 2
+#define CONFIG_ASYNC_TCP_STACK_SIZE 5120
```
proddy, line 57: the stack is set in the platformio.ini file with ` -D CONFIG_ASYNC_TCP_STACK_SIZE=8192`. should we be using 5120? How did you get to this value?

MichaelDvP: Oh,awhile ago i've tested a bit with `uxTaskGetStackHighWaterMark(NULL)` and found with this value i have ~1700 bytes left in S3 and ~2800 in esp32. I've tested with 4k, but  think >1k left is good reserve for future.
From my S3:
grafik

proddy: so remove the define from platformio.ini  ?

MichaelDvP: Ups, i thought i have only formatted the lib, but i've also changed the values for stack (8k->5k), priority (3->5) and queue (32->128), because i first thought it was forgotten with lib change. But you've said the 32 queue is enough now, so should i remove the lib formatting/change commit, or change the values to old?
I think we can leave the platformio define, maybe set to 5k, it overwrites any value in the AsyncTCP.h.

MichaelDvP: Yes, i know it's from esphome, but found only difference to AsyncTCP to add some PRs not merged to AsyncTCP, like IPv6 which we also have already merged. So i think we still need the larger queue, but let's wait for complains, it was only for fast connections with ETH to linux. 

BTW: tasmota has changed IPv6 handling in arduino3 and later also arduino2. Does not work with AsyncTCP. I've tested what to adapt, but now we have different behaviour for 2.0.14 tasmota and 2.014 official, can't be checked with arduino_major_version.  There is also some difference in IPv6 handling of 2 and 3 versions, For now i can connect tasmota v2 with IP6, but not showing the ip correct, tasmota v3 compiles, but does not connect ipv6.

proddy: ok - I'll merge this PR and we keep to our modified version.

Let me know if you need help with the IPv6 stuff. I have some time this weekend for coding.

PR #1639, opened 2024-02-27
merge dev changes/fixes, new entities, env for N32R8 chip, fix custom board profile on boot

MichaelDvP:
also new partition scheme for 16/32MB. Check if you are ok with this.
Enlarging nvs is not possible, the flasher wants ota_data @ 0xe000, so i added a second nvs1 with 250kB. emsesp trys this partition first and use nvs otherwise.


proddy:
Looks fine, can't really test. Let's see what that user comes back with

PR #1646, opened 2024-03-01
enable IPv6 for tasmota-arduino

MichaelDvP:
and some other changes. I'm not sure what to put in dev, what in dev2. please check first.
The 16M partitions are just suggestions, don't know what we need in future, Also nvs1 is only addressed in dev2.
For the esp32 16M with PSRAM i've enabled the SSL (30k heap), i think it should work.


proddy:
thanks @MichaelDvP - compiles ok, can't test since I'm in LA this week but code looks fine. One learning from my side is to not always update to the latest libraries   like Tasmota builds without thorough testing. I guess the same applies to the Web libraries as we've had regression when I moved to new versions of React-Router and Alova. That's a lesson learned for me.

I think at some point we should invest some time into some real tests, copying concepts from Tasmota, ESPHome etc

MichaelDvP:
> to the latest libraries like Tasmota builds without thorough testing. 

We and the dev users are testing the tasmota 2024.01.00 and 24.01.01 since two month now, difficult to find such small changes. I normally use ip addresses for mqtt server and ntp, so i would never see the dns not working after ~1 min. (Also difficult to understand the issue description of discord users)

But we could fix a dev as release candidate with no updates, except small fixes, for a few weeks before release, larger fixes reset RC-time.

Issue #1657, opened 2024-03-12
add telnet command to reboot to previous version


proddy:
From @K3vb3rt in https://github.com/emsesp/EMS-ESP32/issues/1656#issuecomment-1991500769_
 
Sometimes a bad firmware can brick the EMS-ESP and if its the Web UI and OTA is not working you'll need to flash manually using the Flasher Tool or via the command line. This feature is to extend the `restart` command to allow booting to the other partition which has a smaller boot-loader firmware.

@MichaelDvP was working on some code for this.

proddy:
in 3.6.5

PR #1661, opened 2024-03-13
add telnet command to reboot to previous version #1657

MichaelDvP:
Don't know what isthe best keyword, `other` is maybe not selfexplaining.


proddy:
It works, however I would add the following:

- rename 'other' to 'bootloader', and see if it's possible to always put the bootloader .bin in ota_0 (or ota_1), as a failsafe firmware only does uploading.
- if there is no available partition then the  command `restart other` should not restart, but log a warning and do nothing.
- if possible, add a log message to show which partition is being used (shown after the message "[emsesp] Starting EMS-ESP version 3...")

MichaelDvP:
>  rename 'other' to 'bootloader', and see if it's possible to always put the bootloader .bin in ota_0 (or ota_1), as a failsafe firmware only does uploading.

For failsafe we need a factory partition, but i think it is only activated on bootloop, or if we set the ota_data to boot in factory. A factory can not be written by OTA process, so this is fixed until USB-flash. For 4M with loader it is a bit nasty to first reboot to loader (factory) before uploading a new OTA,
> if there is no available partition then the command restart other should not restart, but log a warning and do nothing.

ok.
>  if possible, add a log message to show which partition is being used (shown after the message "[emsesp] Starting EMS-ESP version 3...")

Yes, the label from the partition scheme, e.g. app0, app1

We can add a factory partiton and use two ota-partitions on the 16M chips. We have enough space, so use the actual master as factory (2.5M) two OTA (5,8M), nvs1 (256k), LittleFS (2M), coredump (64k)
```
# Name,   Type, SubType, Offset,   Size,    Flags
nvs,      data, nvs,     0x9000,   0x005000,
otadata,  data, ota,     ,         0x002000,
boot,     app, factory,  ,         0x280000,
app0,     app,  ota_0,   ,         0x590000,
app1,     app,  ota_1,   ,         0x590000,
nvs1,     data, nvs,     ,         0x040000,
spiffs,   data, spiffs,  ,         0x200000,
coredump, data, coredump,,         0x010000,
```
And for the reboot command use the partition name `boot`, `app0`, `app1`

Issue #1665, opened 2024-03-17
Refresh UI - moving settings to one location


proddy:
Moving some of the web screens around, so focus is on dashboards and settings are all behind one section.

proddy:
![image](https://github.com/emsesp/EMS-ESP32/assets/1230712/77d8ddf3-cb8d-4277-b895-2fbf5f4ded7c)

proddy:
![image](https://github.com/emsesp/EMS-ESP32/assets/1230712/a1a1ca1c-263b-41a7-b84e-5c1cc9b824de)

bbqkees:
Looks great. Now that I think of it it makes more sense to do it this way.
Did you 'dock' the Devices tabs back?

proddy:
That's a good idea, @MichaelDvP—having all the Statuses in one place. I'll think of an elegant way to do that.
yes @bbqkees the Devices and Sensors are no longer tabs, so just one click away

proddy:
Latest is in a PR - https://github.com/emsesp/EMS-ESP32/pull/1667

PR #1667, opened 2024-03-22
changes to web layout (status and settings)

proddy:
This is #1665 

lots of files changes, but not a lot of changes, just refactoring and renaming

@MichaelDvP can you take a look and see what you think. Easiest way is via `yarn standalone`


MichaelDvP:
Works good, looks nice. A few minor remarks:
Status page: 
NTP icon should be a clock, network and AP status missing on this page.
System log is (also in current version) a bit strange, the page is guest accessible, but events are admin only. I thiink this should be consistent.
Do you want to merge this to current dev? We have very little issues in 3.6.5-dev Maybe release the dev as is as 3.6.5 
and put the new look in 3.7.0-dev (and merge test (water/pool/remote thermostats/a lot of new entities) to the new dev.

proddy:
Thanks for the feedback. Agree to release 3.6.5 as is and save this for 3.7

I'll make the Log admin only and fix the NTP icon. 

I purposely didn't add AP as I didn't think it important enough. But can add. 

And also Network as I figured if it wasn't connected they wouldn't be able to view the web page 😀

proddy:
I added NTP and fixed the Log (admin only)

Issue #1673, opened 2024-03-26
3.7.0-test.0 (ESP32) - Error Menu -> System 


SoftVienka:
### PROBLEM DESCRIPTION

After clicking Menu -> System, an error message is displayed (only SK language):
```
Unexpected Application Error!

t.reduce is not a function

TypeError: t.reduce is not a function
    at http://192.168.1.240/assets/vendor-OahXKFOG.js:163:84379
    at http://192.168.1.240/assets/vendor-OahXKFOG.js:163:84640
    at Array.map (<anonymous>)
    at http://192.168.1.240/assets/vendor-OahXKFOG.js:163:84001
    at http://192.168.1.240/assets/vendor-OahXKFOG.js:163:84676
    at http://192.168.1.240/assets/vendor-OahXKFOG.js:163:84705
    at b.to [as constructor] (http://192.168.1.240/assets/index-CG2_D2J_.js:7:108407)
    at b.I [as render] (http://192.168.1.240/assets/vendor-OahXKFOG.js:1:9023)
    at z (http://192.168.1.240/assets/vendor-OahXKFOG.js:1:6532)
    at w (http://192.168.1.240/assets/vendor-OahXKFOG.js:1:1476)
```
**3.7.0-dev.0 (ESP32) - no problem.**

### REQUESTED INFORMATION

_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [ ] Searched the problem in [issues](https://github.com/emsesp/EMS-ESP32/issues)
- [ ] Searched the problem in [discussions](https://github.com/emsesp/EMS-ESP32/discussions)
- [ ] Searched the problem in the [docs](https://emsesp.github.io/docs/Troubleshooting/)
- [ ] Searched the problem in the [chat](https://discord.gg/3J3GgnzpyT)
- [ ] Provide the output of http://ems-esp.local/api/system :

```lua
  System information output here:

{
  "System Info": {
    "version": "3.7.0-test.0",
    "uptime": "000+00:28:10.246",
    "uptime (seconds)": 1690,
    "platform": "ESP32",
    "arduino": "Tasmota Arduino v2.0.14",
    "sdk": "4.4.6.240105",
    "free mem": 153,
    "max alloc": 87,
    "free caps": 102,
    "used app": 1606,
    "free app": 378,
    "partition": "app0",
    "reset reason": "Software reset CPU / Software reset CPU"
  },
  "Network Info": {
    "network": "WiFi",
[...]

PR #1674, opened 2024-03-27
fix sk-language #1673, update pkg

MichaelDvP:


proddy:
@MichaelDvP when shall we merge dev2 into dev to share the love?

MichaelDvP:
I think we can merge now.

Issue #1634, opened 2024-02-23
2 warm water circuits 


richii93:
### PROBLEM DESCRIPTION

I do have a Buderus GB162 + RC310. There are 3 heating circuits and 2 warm water circuits connected.
The ems-esp gateway can see all the heating circuits (HC1, HC2, HC3) but only one warm water circuit shows up, expect for "dhw circuit 1 extra" and "dhw circuit 2 extra".
The RC310 does show both warm water circuits and the settings for each one. 

### REQUESTED INFORMATION

```lua
[...]

PR #1698, opened 2024-04-17
dhw circuits, fixes for remote, new entities

MichaelDvP:
This is the change to dhw nest.
@tp1de, for ioBroker you need to add dhw, see https://github.com/MichaelDvP/ioBroker.ems-esp/commit/ff0698fbfee15bbe9e170c6de5ff7e24a1701c25
Also some improvements to the remote emulation


interface/src/types/system.ts
```diff
@@ -67,5 +67,5 @@ export interface LogEntry {
 export interface LogSettings {
   level: number;
   max_messages: number;
-  compact: false;
+  compact: boolean;
```
proddy, line 70: wow, that must have been there for a long time. good find!

src/emsesp.cpp
```diff
@@ -775,23 +775,39 @@ std::string EMSESP::pretty_telegram(std::shared_ptr<const Telegram> telegram) {
...
+            type_name = "?";
+        }
     }
 
     std::string str;
     str.reserve(200);
     if (telegram->operation == Telegram::Operation::RX_READ) {
         str = src_name + "(" + Helpers::hextoa(src) + ") -R-> " + dest_name + "(" + Helpers::hextoa(dest) + "), " + type_name + "("
-              + Helpers::hextoa(telegram->type_id) + "), length: " + Helpers::hextoa(telegram->message_data[0]);
+              + Helpers::hextoa(telegram->type_id) + "), length: " + Helpers::hextoa(telegram->message_data[0])
```
proddy, line 809: i was thinking of changing this to show the length in decimals and not Hex. What do you think?

tp1de:
@MichaelDvP  Thanks for your info. I updated the ioBroker adapter, but I can't release the new version since the test-and-release scripts don't work anymore on Github. I need to wait until this is solved.

@proddy  Could you wait before releasing the new ems-esp version?

proddy:
> @MichaelDvP Thanks for your info. I updated the ioBroker adapter, but I can't release the new version since the test-and-release scripts don't work anymore on Github. I need to wait until this is solved.
> 
> @proddy Could you wait before releasing the new ems-esp version?

yes, just let me know when.

tp1de:
@proddy  I learned that there was a new ioBroker deployment/testing script since some hours. 

The new adapter version is now finally deployed after 6 tries.
It will take some hours until it is visible for end-users.

If you will build the new firmware version this evening it should fit for those users who will install immediately.

MichaelDvP:
@proddy I've fixed the `command::call` for shower coldshot. While testing i see that the coldshot loops and direct after one coldshot the next is started. In this way it makes no sense to define a duration, it will stay cold. I changed the code to ave the second coldshot after a second shower time.  e.g. 7 min shower - 10s coldshot - 7 min shower - coldshot - etc. 
Is this ok? (My system have dhw buffer and can not make a real coldshot, i have to simulate it.)

proddy:
> @proddy I've fixed the `command::call` for shower coldshot. While testing i see that the coldshot loops and direct after one coldshot the next is started. In this way it makes no sense to define a duration, it will stay cold. I changed the code to ave the second coldshot after a second shower time. e.g. 7 min shower - 10s coldshot - 7 min shower - coldshot - etc. Is this ok? (My system have dhw buffer and can not make a real coldshot, i have to simulate it.)

I saw the change, never knew it didn't work. Thanks for the fix. To be honest I never really used the coldshot function. Funny story, I built the first version of EMS-ESP (called boiler.cpp) 5 years ago just to be able to do this and stop my teenage kids from taking 20 minute long showers :-)

tp1de:
I installed dev3 and from my point of view it does not work correctly.
Within thermostat I now see 2 dhw circuits but only one is activated. This is very confusing.

Home assistant entities are now changed as well and lovellace ui need to be adjusted.
And my ioBroker adapter needs adjustments as well. 

Any chance to withdraw this version ?

MichaelDvP:
@tp1de  Read the changelog! This is not v3.6.6dev, it is bigger version change to 3.7 with **breaking changes**. 
The definition of breaking changes is, that you have to do some adjustments on your side.

tp1de:
@proddy @MichaelDvP   Just tell me if you want feedback. Otherwise I will stay quiet.

Do I have missed a discussion with key-users and interface developers about reasons to change the api/mqtt interface and the consequences? 

Having some thousand gateways installed and most users are using home automation software like HA, ioBroker or similar, it is important evaluate the pro and cons of structural and name changes together.  Anyhow this kind of interface changed should be minimized unless there are good reasons.

Together with quite a lot of new entities without any content and dhw2 which do not exist on most systems, this adds even more confusion. How to keep end users informed. It's an issue for @bbqkees  as well. 

In the meantime I will introduce warnings within my ioBroker adapter to use no productive firmware versions on own risk.

proddy:
Fair point. On the other hand EMS-ESP wouldn't be where it is today without some innovative structural changes and hard choices we had to make during the years as we learned more about EMS and the variation of devices out there.

For backward compatibility, we could always add a choice in the WebUI like we did for MQTT format for versions < EMS-ESP 3.6.4.

With the latest dev I'm not seeing any breaking changes in HA, it's still the same. The friendly name changed the WW to DHW only but the sensor name is still the same, e.g. `sensor.boiler_wwsetpumppower`. Can you give an example of what is broken?

tp1de:
> Fair point. On the other hand EMS-ESP wouldn't be where it is today without some innovative structural changes and hard choices we had to make during the years as we learned more about EMS and the variation of devices out there.

Totally agreed. But in respect to dhw I still struggle to follow your approach. 

EMS-ESP data structure is base on devices seen on the bus. But dhw is distributed across  boiler, thermostat and mixer as seen from a device view. For the original km200 gateway api this is seen as an own "device" entry point "dhwCuircuits" with substructures dhw1, dhw2 (if available). I recommended some time ago to consider to do the same.

tp1de:
> With the latest dev I'm not seeing any breaking changes in HA, it's still the same. The friendly name changed the WW to DHW only but the sensor name is still the same, e.g. `sensor.boiler_wwsetpumppower`. Can you give an example of what is broken?

I checked again between 3.6.5 and 3.7.0-dev4. Most mqtt entities are still active but some have changed. Examples:

select.thermostat_wwmode  -> select.thermostat_dhw_mode
switch.boiler_wwonetime -> switch.boiler_dhw_onetime

MichaelDvP:
These changes are discussed since nearly 2 month in #1634
The entity structure is much clearer, any devices can have circuits like hc or dhw, or hs, etc. All circuits are nested and tagged.

tp1de:
So you have done changes for one "exotic" user with 2 dhw circuits which now creates a phantom dhw2 for all users with ems+?

My RC310 could manage 2 ww circuits but I have only one. But still dhw2 is created by the actual firmware (with values - why?) and I need to disable dhw2 with mqq/api by Customizations.  
I believe this needs to be corrected.

MichaelDvP:
> So you have done changes for one "exotic" user

It's always **one** exotic user who helps to decipher the telegrams. 
If values are invalid we don't show them, but if a telegram has a valid value we show it.

I have a unmodulated single stage burner, but burnmodulation and stage 2 is always shown. Are the users of modulated burners or 2-stage burners "extotic" or me? I don't care and add the values.

We have added features to ignore values (customization) and to add values (custom entities), feel free to use this.

tp1de:
From my understanding dhw2 needs a mixer device to be loaded - isn't it?
So without a mixer linked to dhw the dhw2 substructure should not be created on thermostat.
Or do you have a different understanding? 

> We have added features to ignore values (customization) and to add values (custom entities), feel free to use this.
I know how to handle, but are you sure that all other users know as well? Keep it as simple as possible for them.

tp1de:
I feel not to add any value with my feedback.
I will stop now sharing my observations and I went back to 3.6.5 on my 2 installations.

MichaelDvP:
The telegrams to the dhw circuits we have are:
```
RC300WWmode(0x2F5), data: 01 FF 04 00 00 00 08 05 00 08 04 00 00 00 00 00 00 00 00 00 01
RC300WWmode(0x2F6), data: 02 FF 04 00 00 00 08 05 00 08 04 00 00 00 00 00 00 00 00 00 01
```` 
First byte is unknown. Could this be the circuit? If it is zero on a diabled circuit, we can skip the telegram, results in not showing the entities. Can the RC310 also use higher circuit numbers? Do we need to to adapt the tag if the telegram starts with 03,..?

proddy:
Would be really useful if we had example data from @tp1de which we can pre-load into the test.cpp and see the API, MQTT, HA values that Thomas is experiencing.

tp1de:
2F5 is send every minute. 2F6 not
, RC300WWmode(0x02F5), data: 01 FF 04 00 00 00 04 06 00 08 01 00 00 00 00 00 00 00 00 00 01

tp1de:
2F6 pollled with read command
, ?(0x02F6), data: 00 00 04 03 00 00 08 01 00 08 04 00 00 00 00 00 00 00 00 00 01

tp1de:
RC310 supports 2 dhw circuits

MichaelDvP:
> 2F5 is send every minute. 2F6 not (back on. 3.6.5)

I think 2F5 is fetched, not sent out, 2F6 is fetched only in 3.7.
> , ?(0x02F6), data: 00 00 04 03 00 00 08 01 00 08 04 00 00 00 00 00 00 00 00 00 01

As i hoped, the first byte is zero for inactive circuit, So we can skip the reading of the telegram. 
Does this https://github.com/MichaelDvP/EMS-ESP32/releases/tag/latest work for you?

If the first byte is the circuit number and also dhw3,... are possible, we should refactor the code to register dhw circuits like heating circuits as vector and publish with right tag.

tp1de:
> Does this https://github.com/MichaelDvP/EMS-ESP32/releases/tag/latest work for you?

Yes api output is now correct. 

But I observed other strange things:

Dashboard overview does not show all values (boiler / thermostat). 
Opening / refreshing browser or using a different browser does not help.  E.g. thermostat.hc1.dhwprio  and others. 
By selecting these entries the values are shown.
As far as saw, this was already within 3.6.5 - I never recognized.

Within customizations non-active entities are shown without values. dhw2 elements are still present.
API / MQTT and Dashboard are without non-active elements. Why do you show them within customizations? 
    
Have you had the chance to review the mqtt renaming issue for the examples I gave?

MichaelDvP:
> Dashboard overview does not show all values (boiler / thermostat).

That's a bug from the refactoring of the web page in dev4., I'll fix.
> Within customizations non-active entities are shown without values. dhw2 elements are still present.
> API / MQTT and Dashboard are without non-active elements. Why do you show them within customizations?

Customization show all registered entities, You can remove from mem if you like.
For the dhw2 entities i plan to refactor thermostat code and add the circuit dynamic, so it will not show if not present (like unused hc) But this is some more work and takes time.
> Have you had the chance to review the mqtt renaming issue for the examples I gave?

What examples do you mean? This one?: `select.thermostat_wwmode -> select.thermostat_dhw_mode` 
That is the breaking change mentioned in changelog and is intended. 
All dhw entities for all devcies are now prefixed with `dhw`, `dhw2`, ... and the `ww` from mqtt names is skipped.

tp1de:
> All dhw entities for all devcies are now prefixed with `dhw`, `dhw2`, ... and the `ww` from mqtt names is skipped.

Ok, I understood different from @proddy . 

Please do not change nrgww into a new prefixed name since otherwise energy statistics might be destroyed

tp1de:
I setup a fresh test gateway connected to test mqtt server. I am running the last binary from @MichaelDvP 
There seem to be double entries. Old ones with ww-prefix and new ones. Looks strange:

image

tp1de:
All mentioned topics are resolved within latest dev version.
For mqtt dhw name changes made a new issue:  #1714

PR #1711, opened 2024-04-24
small fixes, check second dhw circuit 

MichaelDvP:
see commits


interface/src/project/deviceValue.ts
```diff
@@ -41,10 +41,7 @@ export function formatValue(
     case DeviceValueUOM.SECONDS:
       return LL.NUM_SECONDS({ num: value });
     case DeviceValueUOM.NONE:
-      if (typeof value === 'number') {
-        return new Intl.NumberFormat().format(value);
-      }
-      return value;
+      return new Intl.NumberFormat().format(value);
```
proddy, line 44: thanks @MichaelDvP forgot to change that part

PR #1718, opened 2024-04-26
Polish translation update

pswid:


proddy:
I removed the word "Account" for a reason

```
 ADMIN: 'Admin Account', 
 GUEST: 'Hosť Account'
```

Can you think of a better way that works for Polish? Perhaps put the type in brackets after the name?

pswid:
What reason? In the current version the word "account" was not translated, now (at least in Polish) it makes sense (_admin account_ = _konto administratora_). PR does not change anything in other languages, although it allows you to translate the entire phrase _admin/guest account_ (if needed).

proddy:
Keep the ADMIN and GUEST translations to show the real translations for the words Admin and Guest as before.

Then remove showing the type of the User in the popbox box. That's the easiest fix for everyone.

proddy:
I'll make the changes

PR #1719, opened 2024-04-26
Thermostat dhw, mqtt, ..

MichaelDvP:
The changes of the last days:
- thermostat dhw as vector and dynamic register
- remove some redundant code for hc (use id and check hc_num in heating_circuit)
- mqtt HA settings for v3.6 compatibility
- add missing temperaturesensor dev_cla
- change loglevel for scheduler command fails


proddy:
big changes, looks great. I like the DhwCircuit class and the code is much cleaner now.

PR #1733, opened 2024-05-02
fix show commands, add hp input states

MichaelDvP:
and other small changes, see commits.

BTW: i'm always getting confused by command id and tag-numbers. 
I like to sync this: count TAGs as:
```
enum DeviceValueTAG : int8_t {
     TAG_NONE        = -1, // wild card
     TAG_DEVICE_DATA = 0,
     TAG_HC1, 
```
(skips TAG_HEARTBREAK, this only used one time in mqtt and can be set there).
Should i make the change?


proddy:
> BTW: i'm always getting confused by command id and tag-numbers. I like to sync this: count TAGs as:
> 
> ```
> enum DeviceValueTAG : int8_t {
>      TAG_NONE        = -1, // wild card
>      TAG_DEVICE_DATA = 0,
>      TAG_HC1, 
> ```
> 
> (skips TAG_HEARTBREAK, this only used one time in mqtt and can be set there). Should i make the change?

Ok with changing that. May need some changes to EMSdevice::tag_to_string(), DeviceValue::has_tag() and other places because of the -1 value.

One thing I was thinking, too, is changing the return of all commands from bool to use the enum CommandRet (FAIL, OK, ERROR, INVALID, etc). Sometimes we get users reporting commands don't work and the logs just show an error code '3' for error, so it's hard to find out why. Maybe it was an out-of-bounds error, or the heating circuit doesn't exist, etc.. For example instead of `bool Thermostat::set_temperature()` it would return a `CommandRet_t` if we made it a type. just an idea.

proddy:
looks good, don't think anything was missed. If it was we'll find out soon!

src/command.cpp
```diff
@@ -430,28 +430,44 @@ Command::CmdFunction * Command::find_command(const uint8_t device_type, const ui
...
+        if (Helpers::toLower(cmd) == Helpers::toLower(cf.cmd_) && (cf.device_type_ == device_type) && ((flag & 0x3F) == (cf.flags_ & 0x3F))) {
             cmdfunctions_.erase(it);
             return;
         }
         it++;
     }
 }
 
+// get the tagged command
+std::string Command::tagged_cmd(std::string cmd, const uint8_t flag) {
```
proddy, line 456: it's better to make this a `const std::string & cmd` so the reference is passed to the function and not the whole object class. I'll change it as I need to make some changes to get standalone compiling

Issue #1738, opened 2024-05-04
Remove OTA feature


proddy:
I'm planning on removing the OTA feature. You can upload firmware versions over-the-air via the WebUI upload feature or via a Python script in platformIO or standalone via the CLI.

This will also free up some memory in EMS-ESP for other features.

@MichaelDvP agree with this? I'll make a PR

MichaelDvP:
OK,
 i've used it very few times when i broke the webinterface and don't want to walk stairs, but i think normal users never need it. 
Also remove the OTA description from docs.

proddy:
Ok, I'll do the change now. I've added a script so you can still upload a .bin file from a Linux or DOS prompt

proddy:
with OTA removed there is 8kb flash free'd up and I'm my ESP32-S3 about 2kb more heap and alloc buffer space.

To upload from platformIO add this to your `pio_local.ini`

```
[env]
; upload settings
upload_protocol = custom
custom_emsesp_ip = 10.10.10.173
custom_username = admin
custom_password = admin
upload_port = /dev/ttyUSB*
```

and include the `upload.py` script like

```
extra_scripts =
  pre:scripts/build_interface.py ; comment out if you don't want to re-build the WebUI each time
  scripts/rename_fw.py
  scripts/upload.py
```

To upload from the command line use 

```sh
python3 scripts/upload_cli.py -i 10.10.10.173 -f ./build/firmware/EMS-ESP-3_7_0-dev_8-ESP32_4M.bin
```

PR #1739, opened 2024-05-04
Remove OTA feature #1738

proddy:
https://github.com/emsesp/EMS-ESP32/issues/1738


MichaelDvP (approved):
Looks good and works for me. 
Tested also the upload script: after some windows complains i've tried with wsl and direct working good.

proddy:
thanks for checking, that'll make your PR dev.10

PR #1749, opened 2024-05-08
fix web settings

MichaelDvP:
Have fixed the web enable/disable/config to check the memory difference. 
But i don't have anything to test modbus communication,

Same for the KNX request in discussions, would be nice to support, but don't know how to support the request without any KNX device and no idea how KNX works.


proddy:
thanks Michael. I'll do a compile and another memory check.

At some point we also need to update `settings` in mock-api/rest_server.ts

Issue #1443, opened 2023-11-20
missleading label name


Bingo2023:
rename "Heizgrenze Heizen" -> "Lstg.Zuheizer begrenzen" see pics below.

![2023-11-20_21h33_04](https://github.com/emsesp/EMS-ESP32/assets/61052575/edccaee1-5684-4ec5-aade-217fbe0a294c)
![IMG_20231120_212955_MP](https://github.com/emsesp/EMS-ESP32/assets/61052575/df453e5c-8eac-43c5-8374-2a1260b88d60)
![2023-11-20_21h42_40](https://github.com/emsesp/EMS-ESP32/assets/61052575/b29dd9a6-6fa3-4f87-949b-830b8b3d8aff)

I have a BC400 with latest HW ans SW Version installer by Buderus some month ago. They have cleaned up & regrouped menus -> better overview - I like it :)

I addition:
"Außentemperatur Parallelmodus" -> "Bival. pkt. Parallelbetr." is also changed.
Like you can see on top all of these entries are in a menu called "Elektrischer Zuheizer".
maybe we could add something like "Elek. Zuheizer...." in front of all 4 entries? e.g. "Elek. Zuheizer: Lstg.Zuheizer begrenzen" then all of them would appear in one place in out list of 158 entries.

Bingo2023:
I'm with you Michael, to much renaming is not worth the effort.

But first point is valid, "Heizgrenze Heizen" = is used to limit the power of the electric heater (Heizstab/Zusatzheizer) - that is very difficult to understand, especialy with the other two limits for HP-heating and HP-WW.

proddy:
@Bingo2023 I'd like to close this out. Anything we still need to do on top of the latest dev (3.6.5-dev-8) ?

Bingo2023:
sorry could not find what has changed (without installing), can you point it out?

proddy:
We can change the DE translation if you tell us which are the new wordings (or do a PR on device_translations)

Bingo2023:
proposed:
Heizstab: Limit mit Kompressor
Heizstab: Limit Leistung
Heizstab: Limit für WW
Heizstab: Parallelbetrieb

proddy:
@MichaelDvP ?

MichaelDvP:
Ok for me. 
@Bingo2023 Please make a PR.

proddy:
could you make the PR against the `dev` branch please

PR #1761, opened 2024-05-10
#1443

Bingo2023:
#1443


proddy:
yes, these are the automated tags. It's by design. Michael can better explain.

PR #1793, opened 2024-06-09
Modules

proddy:
Merged with feat_modules
To test you need to add -DEMSESP_TEST
Update AsyncTCP, ESPWebServer, MQTT libraries
Use Arduino 3.0.1
Also updated other small things in the code when I found them


proddy:
@MichaelDvP if you find time could you test it?

MichaelDvP (approved):
A lot of changes, i have not read all the code. 
For espWebServer the  formatting has changed, hard to see the changes. 
I've just compiled and loaded in a s3-esp, works.

I still don't see the usecase for the modules, The examples can be enabled, are working and send a mqtt message. 
What i understand: They have a init on startup, a loop() in the main loop and can access emsesp-functions.

proddy:
> A lot of changes, i have not read all the code. For espWebServer the formatting has changed, hard to see the changes. I've just compiled and loaded in a s3-esp, works.
> 
> I still don't see the usecase for the modules, The examples can be enabled, are working and send a mqtt message. What i understand: They have a init on startup, a loop() in the main loop and can access emsesp-functions.

yes - I'll send you a Discord PR on the use case, which is part of the other GitHub repo

src/system.h
```diff
@@ -82,8 +82,6 @@ class System {
     void send_heartbeat();
     void send_info_mqtt();
 
-    std::string getMacAddress() const;
```
MichaelDvP: This causes link error,because it is refered by 
lib_deps = https://github.com/emsesp/EMS-ESP-Modules.git

proddy: i was working on the code, but didn't update the repo. Just did now, so should work.

I'm do a workaround for the breaking CONFIG_IDF_TARGET_* and then merge

proddy:
@MichaelDvP with the latest dev.13 I just pushed there's a bug in MQTT Discovery which I'll need to fix... I've removed the firmware binaries from the GitHub releases

Issue #1801, opened 2024-06-15
set the duration trigger of a shower


proddy:
Hi, is there any possibility in ESP config to set the duration trigger of a shower higher than 2min? Many thanks in advance...

_Originally posted by @NongJoWo in https://github.com/emsesp/EMS-ESP32/discussions/1293#discussioncomment-9776370_

proddy:
Todo: make SHOWER_MIN_DURATION a setting

proddy:
in dev14 - can you test @NongJoWo

proddy:
> I don't have any experience with GitHub projects. How can I test it?

Download the latest dev release from https://github.com/emsesp/EMS-ESP32/releases/tag/latest, or use the EMS-ESP Web user interface to download this for you. It'll be version 3.7.0 so if you're coming from 3.6.x there will be a few surprises!

NongJoWo:
OK, the update was really easy! I've adjusted the shower timer. Now I just have to take a shower. I'll let you know when everything is OK. Thanks for the super-fast support!!

NongJoWo:
So just tested. The shower trigger works perfectly. Unfortunately, the shower duration no longer works. Only the value 0 is returned here. There is also a new MQTT-device: ems-esp Shower. However, this does not contain any entities.

proddy:
thanks for reporting. I porbably broke something. Will take a look and test soon

NongJoWo:
Should I download and test the new version?

NongJoWo:
Just tested. Now everything is as expected. Many thanks again!

PR #1807, opened 2024-06-19
#1801 shower_duration_min

proddy:
Also fallback to Ardunio 2.0.17


src/shower.h
```diff
@@ -49,15 +46,17 @@ class Shower {
     bool     shower_alert_;          // true if we want the alert of cold water
     uint32_t shower_alert_trigger_;  // default 7 minutes, before trigger a shot of cold water
     uint32_t shower_alert_coldshot_; // default 10 seconds for cold water before turning back hot water
+    uint32_t shower_min_duration_;   // default 2 minutes (1200 seconds), before recognizing its a shower
```
MichaelDvP, line 49: 2 minutes => 120 seconds

PR #1812, opened 2024-06-20
scheduler conditions

MichaelDvP:
This is the testbuild merged.

One thing you can look for in all builds: 
when editing custom entities the number of changes always shows the numbers of entities,
If you have 3 custom entities and andmake a change on one, it shows 3 changes to commit.
But it's cosmetic, no need to open an issue.


proddy:
> This is the testbuild merged.
> 
> One thing you can look for in all builds: when editing custom entities the number of changes always shows the numbers of entities, If you have 3 custom entities and andmake a change on one, it shows 3 changes to commit. But it's cosmetic, no need to open an issue.

do you want me to look into fixing this?

src/system.cpp
```diff
@@ -1448,13 +1490,13 @@ bool System::command_info(const char * value, const int8_t id, JsonObject output
     // Settings
     node = output["Settings"].to<JsonObject>();
     EMSESP::webSettingsService.read([&](WebSettings & settings) {
-        node["board profile"]      = settings.board_profile;
-        node["locale"]             = settings.locale;
-        node["tx mode"]            = settings.tx_mode;
-        node["ems bus id"]         = settings.ems_bus_id;
-        node["shower timer"]       = settings.shower_timer;
-        node["shower alert"]       = settings.shower_alert;
-        node["shpwe_min_duration"] = settings.shower_min_duration; // seconds
```
proddy, line 1457: haha! too much beer that night ;-)

src/web/WebCustomEntityService.h
```diff
@@ -79,8 +78,7 @@ class WebCustomEntityService : public StatefulService<WebCustomEntity> {
     HttpEndpoint<WebCustomEntity>  _httpEndpoint;
     FSPersistence<WebCustomEntity> _fsPersistence;
 
-    std::vector<CustomEntityItem> * customEntityItems; // pointer to the list of entity items
-    // std::list<CustomEntityItem> * customEntityItems; // pointer to the list of entity items
+    std::list<CustomEntityItem> * customEntityItems_; // pointer to the list of entity items
```
proddy, line 81: I typically use <name>_ for private objects

src/web/WebSchedulerService.cpp
```diff
@@ -367,26 +364,66 @@ bool WebSchedulerService::command(const char * cmd, const char * data) {
...
 
+#include "shuntingYard.hpp"
+
+bool WebSchedulerService::onChange(const char * cmd) {
+    for (const ScheduleItem & scheduleItem : *scheduleItems_) {
+        if (scheduleItem.active && scheduleItem.flags == SCHEDULEFLAG_SCHEDULE_ONCHANGE && Helpers::toLower(scheduleItem.time) == Helpers::toLower(cmd)) {
+#ifdef EMESESP_DEBUG
+            // emsesp::EMSESP::logger().debug(scheduleItem.cmd.c_str());
+#endif
+            return command(scheduleItem.cmd.c_str(), compute(scheduleItem.value.c_str()).c_str());
```
proddy, line 375: compute() takes std::string so no need to convert to char *

src/web/WebSchedulerService.cpp
```diff
@@ -367,26 +364,66 @@ bool WebSchedulerService::command(const char * cmd, const char * data) {
...
+        }
+    }
+    return false;
+}
+
+void WebSchedulerService::condition() {
+    for (ScheduleItem & scheduleItem : *scheduleItems_) {
+        if (scheduleItem.active && scheduleItem.flags == SCHEDULEFLAG_SCHEDULE_CONDITION) {
+            auto match = compute(scheduleItem.time.c_str());
```
proddy, line 384: compute() takes std::string so no need to convert to char *

src/web/WebSchedulerService.cpp
```diff
@@ -367,26 +364,66 @@ bool WebSchedulerService::command(const char * cmd, const char * data) {
...
+#ifdef EMESESP_DEBUG
+            // emsesp::EMSESP::logger().debug("condition match: %s", match.c_str());
+#endif
+            if (!match.empty() && match[0] == '1') {
+                if (scheduleItem.retry_cnt == 0xFF) { // default unswitched
+                    scheduleItem.retry_cnt = command(scheduleItem.cmd.c_str(), compute(scheduleItem.value.c_str()).c_str()) ? 1 : 0xFF;
```
proddy, line 390: compute() takes std::string so no need to convert to char *

src/web/WebSchedulerService.cpp
```diff
@@ -367,26 +364,66 @@ bool WebSchedulerService::command(const char * cmd, const char * data) {
...
+        last_uptime_sec = uptime_sec;
+    }
+
     // check startup commands
-    if (last_tm_min == -1) {
+    if (last_tm_min == -2) {
         for (ScheduleItem & scheduleItem : *scheduleItems_) {
             if (scheduleItem.active && scheduleItem.flags == SCHEDULEFLAG_SCHEDULE_TIMER && scheduleItem.elapsed_min == 0) {
-                scheduleItem.retry_cnt = command(scheduleItem.cmd.c_str(), scheduleItem.value.c_str()) ? 0xFF : 0;
+                scheduleItem.retry_cnt = command(scheduleItem.cmd.c_str(), compute(scheduleItem.value.c_str()).c_str()) ? 0xFF : 0;
```
proddy, line 423: compute() takes std::string so no need to convert to char *

src/web/WebSchedulerService.cpp
```diff
@@ -401,7 +438,7 @@ void WebSchedulerService::loop() {
             // scheduled timer commands
             if (scheduleItem.active && scheduleItem.flags == SCHEDULEFLAG_SCHEDULE_TIMER && scheduleItem.elapsed_min > 0
                 && (uptime_min % scheduleItem.elapsed_min == 0)) {
-                command(scheduleItem.cmd.c_str(), scheduleItem.value.c_str());
+                command(scheduleItem.cmd.c_str(), compute(scheduleItem.value.c_str()).c_str());
```
proddy, line 441: compute() takes std::string so no need to convert to char *

src/web/WebSchedulerService.cpp
```diff
@@ -416,8 +453,9 @@ void WebSchedulerService::loop() {
         uint16_t real_min = tm->tm_hour * 60 + tm->tm_min;
 
         for (const ScheduleItem & scheduleItem : *scheduleItems_) {
-            if (scheduleItem.active && (real_dow & scheduleItem.flags) && real_min == scheduleItem.elapsed_min) {
-                command(scheduleItem.cmd.c_str(), scheduleItem.value.c_str());
+            uint8_t dow = scheduleItem.flags & SCHEDULEFLAG_SCHEDULE_TIMER ? 0 : scheduleItem.flags;
+            if (scheduleItem.active && (real_dow & dow) && real_min == scheduleItem.elapsed_min) {
+                command(scheduleItem.cmd.c_str(), compute(scheduleItem.value.c_str()).c_str());
```
proddy, line 458: compute() takes std::string so no need to convert to char *

src/web/WebSchedulerService.cpp
MichaelDvP: relict from first implementation, will change.

src/web/WebCustomEntityService.h
MichaelDvP: the pointer is private, _ is also better to differ the list and the pointer to list, like in schedule.

proddy:
@MichaelDvP I would also add
 - sort the Web table by the schedule column (it's still Time). When you add a new onChange or Conditions item, it looks funny.
 - add an example to rest_server for the mockup and online demo
 - make sure it compiles standalone/native, as this is used in Sonar static analysis checks. I can also do this.
 - I would rename "OnChange" to "On Change" in the table (Schedule column), so it's consistent with the buttons in the web dialog
 - A default value is set for the Schedule in the web so when you add a new Schedule Item the Time box looks strange. I can also look at fixing that. It's not urgent.

I haven't had time to try this out (I'm away) but was thinking whether we need a fail-safe in case the Scheduler breaks something (like a circular reference). So when EMS-ESP reboots it automatically disables the Scheduler with a warning message

MichaelDvP:
Ok, but don't know why sonar breaks. the error is 
`src/web/shuntingYard.hpp:394:16: error: enumeration value ‘Unknown’ not handled in switch [-Werror=switch-enum]`
but the switch have a default: to handle this. But i add the other cases.

proddy:
i don't know either until I can try myself on a Linux WSL. Maybe easiest is adding a -Wno-switch-enum in the `Makefile`

proddy:
> The Run-Build works now, but on PR the scanner stops at `ERROR: Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator to check the permissions of the user the token belongs to`

that's fine - it works for EMS-ESP dev as the project key's are stored as GitHub secrets in the main repo. I couldn't find a way to prevent the GitHub action running on the PR's repo.

PR #1832, opened 2024-06-30
feat: add Modbus support

mheyse:


mheyse:
I just pushed a fix for an indentation error in update_modbus_registers.py.

mheyse:
Added `generate-modbus-register-doc.*` scripts

proddy:
I have one PR of mine pending, when then goes, you can refresh and we'll push this one.

proddy:
@mheyse could you refresh with the latest dev, fix those merge issues and then we can merge the PR and roll it out to the world. It would be good to add some documentation to the wiki as well https://emsesp.github.io/docs/Integrations/

mheyse:
OK I rebase to the latest dev, have to test it tonight.

Regarding the Wiki: isn't https://github.com/emsesp/docs/pull/36 what 
you're looking for? If not, can you be more specific what is needed?

mheyse:
Rebased on latest dev.

proddy:
Thanks, it's merged. There are some small lint warnings that need fixing. I can do most of then, but one if you could look at is in `Modbus::check_parameter_order()`. Printing `prev->short_name` when prev is a nullptr (at first) will crash the ESP32

mheyse:
> Thanks, it's merged. There are some small lint warnings that need fixing. I can do most of then, but one if you could look at is in `Modbus::check_parameter_order()`. Printing `prev->short_name` when prev is a nullptr (at first) will crash the ESP32. Can you take a look?

https://github.com/emsesp/EMS-ESP32/pull/1840

> The other which I changed is `Helpers::numericoperator2scalefactor(int8_t numeric_operator)`. Since the value can be negative I changed the declaration from unit to int

OK that's correct

PR #1833, opened 2024-07-01
Modbus doc

mheyse:
This PR adds scripts to generate Modbus documentation for all entities. It is currently based on the "feat: add Modbus support" PR.


mheyse:
OK, I'll modify the original PR. (My intention was to first merge the 
Modbus PR, then rebase the Modbus-doc PR, then it's only 2 files.)

Issue #1174, opened 2023-05-13
Feature: Allow device name to be customized


proddy:
> A user wants to look at details from their device so the name is more important that the type

For users with multiple devices, like 3 mixers for hcs and dhw name and type are not clear. Maybe think about custom device names, or add, depending on device-id, the circuit to the name (e.g. "MM100 (hc1)"). But i think optional custom names are better.

_Originally posted by @MichaelDvP in https://github.com/emsesp/EMS-ESP32/issues/1169#issuecomment-1546606909_

proddy:
This feature is to extend the Settings->Customization page to allow the EMS Device Name to be renamed. I suggest adding a RENAME button to that page after a device is selected. And the name will be stored in the customizations `masked_entities` alongside the product_id and device_id which makes it easier and less code. If the custom name is empty ('') it will default to the one stored in Flash from device_library.

It sounds simple but there is some backend code to write.

proddy:
As our DB is growing and we're seeing multiple brands and models using the same ProductID I think this is important to add as a next feature. I'll add a Rename option to the Customizations WebUI.

PR #1834, opened 2024-07-02
Allow device name to be customized 

proddy:
This implements #1174 - Feature: Allow device name to be customized

I've tested it, and it seems to work okay.


MichaelDvP:
Works good, but without prefixed Devicetype the selection field is getting very small. Maybe make a grid with 6-width? And add devcietype back, i think it helps to identify the right ems-device.
grafik

MichaelDvP:
Aha, i see, you have skipt the devicename because of the editfield. 
What do you think of this with ttn as translated type name, only affects the selector:
```
          <TextField
            sx={{ width: 1/2 }}
            name="device"
            label={LL.EMS_DEVICE()}
            variant="outlined"
            value={selectedDevice}
            disabled={numChanges !== 0}
            onChange={(e) => setSelectedDevice(parseInt(e.target.value))}
            margin="normal"
            select
          >
            <MenuItem disabled key={-1} value={-1}>
              {LL.SELECT_DEVICE()}...
            </MenuItem>
            {devices.devices.map((device: DeviceShort) => (
              <MenuItem key={device.i} value={device.i}>
                {device.ttn ? device.ttn + ' ' + device.s : device.s}
              </MenuItem>
            ))}
          </TextField>
```

proddy:
I've made the change, just adding the Device Type name as an extra parameter - which is translated. PR'ing now....

proddy:
merging, so modbus can be added next

PR #1840, opened 2024-07-04
fix: nullptr linter error

mheyse:
This can never happen (that branch is only executed if isFirst is false, and when isFirst is set to false prev is also assigned). But the linter complains without nullptr checks, so here is a fix.

### DIFF
--- a/src/modbus.cpp
+++ b/src/modbus.cpp
@@ -65,8 +65,10 @@ bool Modbus::check_parameter_order() {
     for (const auto & mi : modbus_register_mappings) {
         if (isFirst) {
             isFirst = false;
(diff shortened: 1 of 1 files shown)


proddy:
thanks

mheyse:
According to sonarcloud there still is an issue?

https://sonarcloud.io/project/issues?id=emsesp_EMS-ESP32&open=AZB8xhHs3lw9YRNNmgCq

Sure, it is technically correct that `mi` could contain a reference to nullptr, but I don't see how this could possibly happen? Could you enlighten me? Do I really have to check for `&mi != nullptr` within the loop?

proddy:
ok, I've marked it as Safe.

PR #1843, opened 2024-07-04
remotetemp CMD, temperaturesensor id, controlsettings BC400, Boiler/HP

MichaelDvP:


proddy:
nice, thanks for adding the Boiler/HP too

mattreim:
Unfortunately `CMD` is not supported on the ioBroker adapter and therefore I can no longer use `remotetemp` as of v22. There is no other option for me because my temperature sensor is a Zigbee device.

Can you please change `remotetemp` back to `INT16`?

MichaelDvP:
> Can you please change remotetemp back to INT16?

No, i think the better way is that CMD show up in ioBroker,  so you can also use `boiler/reset` or `boiler/hpshutdown`.
Check https://github.com/MichaelDvP/EMS-ESP32/releases

tp1de:
@MichaelDvP @proddy 
Sorry guys I do not understand at all what has changed in respect to remotetemp and the API interface used by ioBroker.
Can you please explain?

tp1de:
And "type": "command" can be now "numeric" or "enum". Anything else?
Do I need a special command to change value by API?

tp1de:
May I add another observation:

When entering a remotetemp value within dashboard another thermostat is created.
Therefore within .../api/system 2 thermostat entries are listed - same type-name. 

To retrieve the values per device the api command is used with type-name. 
It would be much better to create a unique type-name retrieve the values per device. (e.g. thermostat_2)

MichaelDvP:
As mentioned it is for testing in my repo only. 
The double entities are already fixed in official. Before the cmd change it does not matter, because hc1/remotetemp from master thermostat is the same value as from remote. But now the master is command only without value, so there is a conflict. The temperature from remote thermostat is now shown as `RFtemp`. 
In my repo:
`api/thermostat/values` lists also commands but value is blank. 
`api/thermostat/remotetemp` gives the normal structure with `type:number`,writeable and no value
'api/boiler/reset' gives a `type:enum` with options and no value
Works for me with the actual version of ioBroker adapter, there is no need to change something in the adapter.

tp1de:
Thanks @MichaelDvP for clarification. Yes it works. Just one proposal for improvement:

remotetemp is type number but value is string "". This gives a warning in ioBroker. Better to leave value empty (null)

MichaelDvP:
I think it's not possible to leave a json value empty. I try now with `NAN`, shows up in ioBroker and as plain text as `null`, just test.

mattreim:
If I delete the object structure in ioBroker, `remotetemp` is not recreated. With the old structure the following error appears:

2024-07-08 18:59:51.566 ERROR 33: [emsesp] API failed with error unknown device (Error)
2024-07-08 19:00:51.880 ERROR 34: [emsesp] API failed with error unknown device (Error)
2024-07-08 19:01:51.684 ERROR 35: [emsesp] API failed with error unknown device (Error)
2024-07-08 19:02:51.590 ERROR 36: [emsesp] API failed with error unknown device (Error)
2024-07-08 19:03:51.600 ERROR 37: [emsesp] API failed with error unknown device (Error)
2024-07-08 19:04:51.706 ERROR 38: [emsesp] API failed with error unknown device (Error)
2024-07-08 19:05:51.618 ERROR 39: [emsesp] API failed with error unknown device (Error)

tp1de:
> I think it's not possible to leave a json value empty. I try now with `NAN`, shows up in ioBroker and as plain text as `null`, just test.

Works fine now for me

tp1de:
> If I delete the object structure in ioBroker, `remotetemp` is not recreated. With the old structure the following error appears:

@mattreim are you using the build from @MichaelDvP ? https://github.com/MichaelDvP/EMS-ESP32/releases/tag/latest

mattreim:
It works again and `reset` is there too. Thank you

PR #1866, opened 2024-07-15
fix value info for second thermostat

MichaelDvP:


proddy:
I'll update my tests to include a second thermostat

PR #1873, opened 2024-07-19
Scheduler web access and others

MichaelDvP:
See commits. 

The scheduler web access can be used for commands and for condition/value. Has to be a json.
Commands: 
`{"url":"<url>", "header":{"content-type":"some type", "aut":".."}, "value":".."}`
url is needed, the other are optional. Any header pair in the header-object will be set. value is optional and will set the value in a body of post, equivalent to setting the value in the value fielt of the scheduler.
The command is send out and does not read any reply data.

Shunting yard:
Used for condition and value field. Same sytax as in command, but reads the data from web, could be critical timing.
You can add a json `"key":"xyz"`, then the web reply is parsed and only key is set. 
For a tasmota switch these are same result:
`{"url":"http://192.168.0.101/cm?cmnd=power1", "key":"power1"} == on` 
`{"url":"http://192.168.0.101/cm?cmnd=power1"} == {"power1":"on"}`

I've made some test, but all in home network with low response time. 
Please check carefully before merging.


proddy:
nice work! I'll check out your code and run some tests locally. I can also simulate a slow network and see what happens when there is a timeout.

PR #1876, opened 2024-07-20
scheduler async, command use string, lowercase

MichaelDvP:
I've changed the command paramters to string, makes it easier to lowercase convert and saves a lot of .c_str()
For the command we don't need to check for '{' as first character, the deserialize parses good, accept also blanks before '{'

Please check.


proddy:
looks good. I like what you did with queuing the cmd_changed

I still need to try some use cases with HTTP timeout and large returns

PR #1883, opened 2024-07-22
refactor web file structure and seperate settings from status

proddy:
These are mainly cosmetic changes. Icleaned up the file structure and made a cleaner separation between Status (which was called System before) and Settings.

let me know if you see any issues.

There's a live demo on https://ems-esp.derbyshire.nl


MichaelDvP:
Looks good. 
Do we really need the "Active Devices & Sensors", also the scan devices is imho useless. 
I would prefere the version on top.

proddy:
agree, I'll remove the Active Devices & Sensors and also the scan endpoint and code. And move version to the top.

proddy:
Merged the web changes, and changes for module so it compiles again on C3.

PR #1888, opened 2024-07-23
fix system command

proddy:
add more tests, all seems to work. I'll keep adding tests for the unknown calls


MichaelDvP:
This removes the `return_not_found` from system, but the others in analogsensor, temperaturesensor, customEntity, scheduler will give the same message if there is a cf found. I still think it's better to clear the output after cf is found, 
There still the check:
https://github.com/emsesp/EMS-ESP32/blob/dfe85b9ba7b9044dc4c1658935142337fed1b53d/src/command.cpp#L349-L353
that keeps the messages from system, analog, temperatur, etc.  If you remove all the  `return_not_found` this check is also not needed.
This is safe, much less error-prone and shows all messages from the value_info if there is no cf.
https://github.com/MichaelDvP/EMS-ESP32/blob/bc7d848b9b64c968f451b1a42b531c3547e7a982/src/command.cpp#L356

proddy:
I see your logic. Let me make some changes and add some more tests. I found some missing checks with missing attributes in the sensors too.

proddy:
I added some tests - they all succeed. Please take another look. If I add the root.clear() earlier after a successful cf, it will make unknown commands always return 'info'.

src/command.cpp
```diff
@@ -506,10 +510,11 @@ std::string Command::tagged_cmd(const std::string & cmd, const uint8_t flag) {
 
 // list all commands for a specific device, output as json
 bool Command::list(const uint8_t device_type, JsonObject output) {
-    // force add info and commands for those non-EMS devices
+    // check of it a 'commands' command
     if (device_type == EMSdevice::DeviceType::TEMPERATURESENSOR || device_type == EMSdevice::DeviceType::ANALOGSENSOR) {
         output[F_(info)]     = Helpers::translated_word(FL_(info_cmd));
         output[F_(commands)] = Helpers::translated_word(FL_(commands_cmd));
+        output[F_(values)]   = Helpers::translated_word(FL_(values_cmd));
```
MichaelDvP, line 517: For ems devices the values command is hidden. Please show it for all devices or for none, but don't mix.

src/console.cpp
```diff
@@ -575,11 +575,12 @@ static void setup_commands(std::shared_ptr<Commands> & commands) {
                     return;
                 } else {
                     // show message if no data returned (e.g. for analogsensor, temperaturesensor, custom)
-                    shell.println("No data.");
+                    shell.println("No data returned.");
```
MichaelDvP, line 578: This `no data` is confusing. It show for every command that have no json, 
E.g. For a `call system send ""` i don't expect a `no data` message, i only expect thet the command is executet.

src/command.cpp
proddy: I saw that too and was thinking of making `values` not HIDDEN. It's the only command that is. What do you think?

src/web/WebSchedulerService.cpp
```diff
@@ -136,6 +136,7 @@ bool WebSchedulerService::get_value_info(JsonObject output, const char * cmd) {
     if (Helpers::toLower(cmd) == F_(commands)) {
         output[F_(info)]     = Helpers::translated_word(FL_(info_cmd));
         output[F_(commands)] = Helpers::translated_word(FL_(commands_cmd));
+        output[F_(values)]   = Helpers::translated_word(FL_(values_cmd));
```
MichaelDvP, line 139: same as before hidden or not?

src/command.cpp
MichaelDvP: Ok for me, show it everywhere.

src/web/WebSchedulerService.cpp
proddy: this is a special case as `info` shows the same data as `values` in the scheduler right now.

proddy:
I'm going to move some checks around too. I still think it's a waste to create device_info and then when there is an invalid entity or command wipe the json output and replace with an error. It's expensive and a waste. I'll move the deviceID check earlier in the code and do some more cleaning.

src/console.cpp
MichaelDvP: `No data` looks like a error message and is confusing. Maybe show in terminal a confirmation `command successful`, but i expect the command is executed if there is no error-message.

src/web/WebSchedulerService.cpp
MichaelDvP: yes, also for custom. If needed we can add more attributes to info.

MichaelDvP:
> I still think it's a waste to create device_info and then when there is an invalid entity or command wipe the json output and replace with an error. 

That's mainly the `return_not_found` you have placed at the end of each `get_value_info` and then check it before checking the cf functions. It's always the same output that has to be cleared and can be replaced by single output in the `if (!cf)` section. The attiributs error can stay, it will never conflict with cf.

src/command.cpp
proddy: ok. done.

src/console.cpp
proddy: done.

proddy:
attempt #4 - think it works, at least the tests pass. And I didn't need to flush root.clear()

src/command.cpp
```diff
@@ -400,12 +400,15 @@ uint8_t Command::call(const uint8_t device_type, const char * cmd, const char *
...
+            // TODO for now using debug (regarles if compiled with EMSESP_DEBUG)
+            // TODO mvdp ?
+            logger_.debug(("%sCalled command %s"), ro.c_str(), info_s);
         } else {
             if (id > 0) {
-                LOG_INFO(("%sCalling command %s with value %s and id %d on device 0x%02X"), ro.c_str(), info_s, value, id, device_id);
+                logger_.debug(("%sCalled command %s with value %s and id %d on device 0x%02X"), ro.c_str(), info_s, value, id, device_id);
             } else {
-                LOG_INFO(("%sCalling command %s with value %s"), ro.c_str(), info_s, value);
+                logger_.debug(("%sCalled command %s with value %s"), ro.c_str(), info_s, value);
```
MichaelDvP, line 411: single command will always reply with data, so debug level is right.
Calling setting commands with value should have info log level. We can see in the log if users flood the bus by sending a lot of commands.

MichaelDvP (approved):
Yes, looks all ok now,
Only thing is that i prefere log_info for the commands with values.

PR #1899, opened 2024-07-26
scheduler single and fixes

MichaelDvP:
I've split in small commits.


src/command.cpp
```diff
@@ -353,6 +353,8 @@ uint8_t Command::call(const uint8_t device_type, const char * cmd, const char *
         return CommandRet::ERROR;
     }
 
+    output.clear(); // we have a command function, clear messages from device_value_info
```
proddy: don't think its needed, but does no harm

MichaelDvP: Oh, this was from my first solution for Thomas, not needed anymore after merging your changes.

proddy: no problem, we can change later

MichaelDvP: I see that i've already removed it while merging your changes in https://github.com/emsesp/EMS-ESP32/pull/1899/commits/a8307bddbec4b3bfbc165653fd03d6a8c0c83ef4

Issue #1893, opened 2024-07-25
Feature: call schedule entries via command


proddy:
From an earlier discussion:

I'm thinking, @MichaelDvP, shall we extend the Scheduler so you can execute the event? Even if its disabled. That way you could create quite complex commands and conditions, keep it disabled and invoke it manually via the API.

Michael said:

Something like a new type of schedule: single action: inputs just activate, name, command and value: executes one time after activation and sets active to false. Activating in the dialogue means execution on the ok update button; in the list, it is always deactivate. Or we add a button for execution and skip the checkbox.

But you can also do this now: use trigger onChange with a custom value. By changing the custom value by command, you'll trigger the schedule. Or use condition 1,but then you have to send on and off within 10 seconds to have a one-time execution.

_Originally posted by @proddy in https://github.com/emsesp/EMS-ESP32/discussions/1835#discussioncomment-10148907_

MichaelDvP:
Some more thoughts about that:
- A single action command does not have a schedule trigger, so only show command, value, and name  if no trigger is selected:
![grafik](https://github.com/user-attachments/assets/d0511ac3-b2e6-410b-84cc-c8b594440744)
- The Execute button make it active. In the list the schedule and tme field is empty.
![grafik](https://github.com/user-attachments/assets/5f0cd47b-4082-4772-824b-1c5bb87ea686)
![grafik](https://github.com/user-attachments/assets/ea3153db-9bc7-41d6-a3b5-3055b0a9f3c0)
- You can call a command from mqtt/api/terminal to activate and the schedule-command is executed and deactivated again:
```
ems-esp-s3:# call scheduler info
{
  "webtest": 0,
  "singleshot": 0
}
ems-esp-s3:# call scheduler singleshot
{
  "name": "singleshot",
  "type": "string",
  "active": 0,
  "command": "custom/test",
  "value": "custom/test + 3",
  "readable": true,
  "writeable": true,
  "visible": true
}
ems-esp-s3:# call scheduler singleshot on
Command executed. Check log for messages.
000+00:01:11.485 I 1: [command] Called command scheduler/singleshot (enable schedule item) with value on
000+00:01:11.485 I 2: [command] Called command custom/test (set custom value on ems) with value 8
ems-esp-s3:# call scheduler singleshot
{
  "name": "singleshot",
  "type": "string",
  "active": 0,
  "command": "custom/test",
  "value": "custom/test + 3",
  "readable": true,
  "writeable": true,
[...]

PR #1900, opened 2024-07-27
updates to scheduler web ui for #1893

proddy:
still work in progress to make sure everything works, and then add the LL. translations


proddy:
going to merge this to fix the bootloop, its the quickest way

MichaelDvP:
Scheduler looks nice, but i can't change the scheduler type. Only when adding a new scheduler.

MichaelDvP:
Also on newschedule i can toggle the buttons and  set to undefined state, showing odd options.

proddy:
I'll take a look. You can't change the scheduler type when editing, I did this by design. We can change it back, but needs some more code to reset fields to their default (as time is used for a time and also a cmd).

proddy:
> Also on newschedule i can toggle the buttons and set to undefined state, showing odd options.

can't reproduce - can you give an example?

MichaelDvP:
grafik

MichaelDvP:
I think the scheduler type buttons should react as radiobutton, not as togglebutton.

proddy:
ok, I made https://github.com/emsesp/EMS-ESP32/pull/1901 to track

PR #1907, opened 2024-07-30
refactor common commands 

MichaelDvP:
Please check. 
- `commands` is completly processed in commands.ccp
- all other common commands without value are processed in a specific `get_value_info()`
-  there is a `get_value_json`  for a single entity called from get_value_info
- `entities` command lists all values with all info.
- system is nearly identical, but a bit tricky. e.g. you don't get uom for a value.
- still the `strcmp(.., info)` etc are in all get_value_info and each test have a loop. (only one loop executed on command)  We can make the stringcompare in commands, set a enum, send it get_value_info and do a switch/case there.

I've tested for different devices:
- api/device
- api/device/info
- api/device/value
- api/device/entities
- api/device/commands
- api/thermostat/hc2/values (info/entities) to get single circuit info
- api/system/network/values (info/entities) to get single circuit info same as for ems devices
- api/device/name
- api/device/name/attribute
- api/device/circuit/name
- api/device/circuit/name/attirbute

I think it's all working now. But maybe a forgot a test.


proddy:
I'll take a look.

proddy:
I looked through the code, really nice and much more tidy. I'll test tomorrow and add some to test.cpp with the other combinations so we can view the output.

proddy:
I'm working on some unit test scripts for this which will follow this PR, but its good to merge.

platformio.ini
```diff
@@ -39,7 +39,7 @@ unbuild_flags =
 
 [espressi32_base]
 ; 6.7.0 = Arduino v2.0.16 (based on IDF v4.4.7). See https://github.com/platformio/platform-espressif32/releases/tag/v6.7.0
-platform = espressif32@6.7.0
+platform = espressif32@6.8.0
```
proddy, line 42: 6.8.1 is already out

proddy:
The standalone make is broken as it's missing some String functions. Either I'll add them, or change the code to use std::string and not the horrible Arduino String()

MichaelDvP:
I'll change to std::string. mixing string types in the same function looks odd. I've used the String to test the http with the native parameter type, but c_str() is also accepted (and working).

proddy:
ok, that's better. Then I'll stop the changes I was working on and wait. And then I'll push my PR which includes quite some changes so it compiles and runs the test on native Windows.

proddy:
@MichaelDvP I'm porting over the tests from test.cpp to the Unity test_api.cpp and found one call that fails which needs fixing.


```
web output: [/api/boiler/dhw/chargetype] (400) {"message":"no dhw in boiler"}
web output: [/api/boiler/dhw.chargetype/writeable] (400) {"message":"no dhw.chargetype in boiler"}
```

the value is there:

```
ems-esp:# call boiler values
{
  "reset": "",
  "heatingoff": false,
  "heatingactive": false,
  "tapwateractive": true,
  "selflowtemp": 0,
  "curflowtemp": 60.2,
  "rettemp": 48.1,
  "syspress": 1.4,
  "burngas": true,
  "burngas2": false,
  "flamecurr": 37.4,
  "fanwork": true,
  "ignwork": false,
  "oilpreheat": false,
  "heatingpump": true,
  "selburnpow": 115,
  "curburnpow": 61,
  "ubauptime": 3940268,
  "servicecode": "=H",
  "servicecodenumber": 201,
  "nompower": 0,
  "nrgtotal": 0.0,
  "nrgheat": 0.0,
  "dhw": {
    "seltemp": 52,
    "comfort": "hot",
    "flowtempoffset": 40,
    "circpump": false,
    "chargetype": "3-way valve",
    "hyston": -5,
    "hystoff": 0,
    "disinfectiontemp": 70,
    "circmode": "off",
    "circ": false,
    "storagetemp1": 53.8,
    "activated": true,
    "3wayvalve": true,
    "nrg": 0.0
  }
}
```

I'll continue to add the tests so you can run them too.

MichaelDvP:
Typo, I've made a fix: https://github.com/emsesp/EMS-ESP32/commit/732949fdb80880e34d12892fcf13bf7d1b230f72

proddy:
nice. I've added it my PR.

PR #1912, opened 2024-08-02
scheduler: change to std::string

MichaelDvP:


src/web/WebSchedulerService.cpp
```diff
@@ -334,31 +334,32 @@ bool WebSchedulerService::command(const char * name, const std::string & command
...
+        // don't search the complete url, it may contain a devicename in path
+        auto q = url.find_first_of('?');
+        if (q != std::string::npos) {
+            auto s = url.substr(q + 1); // copy only parameters
+            auto l = s.length();
+            commands(s, false);
+            url.replace(q + 1, l, s);
         }
-        if (url.startsWith("http") && http.begin(url)) {
+        if (!url.find("http") && http.begin(url.c_str())) {
```
proddy, line 349: Codacy comes back with a suggestion 

"Inefficient usage of string::find() in condition; string::starts_with() could be faster."

we can change those later.

MichaelDvP: the esp32 basic_string does not have '.starts_with()', we can use `!url.compare(0, 4, "http")`, mainly identical to `url.substr(0,4) == "http"`, but i don't know if it more or less efficient.

proddy: ah ok, I'll make it a false positive then in Sonar and Codacy

MichaelDvP: We can also skip it. I added it because i had crashes with command parsing in URL, but at last it was a stack overflow when using the `compute` to URL. Only checking the get parameters for command works fine in actual stack.

proddy: its harmless, let's make a comment or a TODO and leave it

PR #1913, opened 2024-08-02
fix sonar complains

MichaelDvP:


proddy:
ha! I just made the same change. all good :-)

PR #1914, opened 2024-08-02
various changes, work in progress

proddy:
- add Unit Tests using the Unity framework (https://github.com/ThrowTheSwitch/Unity). I still need to port over the API ones from test.cpp. To run, on any platform, use `pio run -e native-test -t exec`
- Support running native tests on Windows. For the console you need something that supports winsock. I've added notes in the  platformio.ini file and also https://emsesp.github.io/docs/Building/#simulating-offline-without-an-esp32-microncontroller. So now people building on Windows can modify the test.cpp and check it compiles without using the Makefile before checkin.
- the display prompt looks strange on Windows (there's an extra newline), it's not a huge problem but I'll see if I can fix it later

TODO

- [ ] Add console commands like `call boiler dhw/chargetype`


MichaelDvP:
Tried to compile the native build in windows. Seems the gcc does not know the %F%T options of strftime, After replacing it works.

proddy:
strange, didn't see that. I'm on Windows 11 Pro x64

Also it uses g++ when I do a `pio run -e native`.

proddy:
I've added all the API tests now, 56 so far automatically testing each function. Run with `pio run -e native-test`. 

I'll merge your PR and then fix the merge issues and merge this one in.

proddy:
Still to add are the console commands and perhaps adding a GitHub action into the CI pipeline to auto check

PR #1921, opened 2024-08-04
switchProgMode offset 13 to 19, increase length of raw custom entities

MichaelDvP:


mattreim:
Here the translation is missing and it is the only value with unity.

EMS

EMS2

Issue #1920, opened 2024-08-04
Updates to webUI


proddy:
I'm planning to make some changes to the WebUI, and noting them down here so I don't forget

- [x] upgrade Alova from version 2 to 3. This is done, but there's a bug in Alova (rendering with preact) that is blocking the PR
- [x] refactor all the updateState() to use methods
- [x] use AutoRefreshRequest() in devices, sensors, custom entities, status (activity, network, hardware, mqtt, ntp, ap) and remove the Refresh button where appropriate.
- [x] uppercase "Never" "Always" etc in the AP, to make it consistent
- [x] Rename Customization to Customizations in .tsx filenames to be consistent with naming in UI
- [x] Remove the RemoveTrailingSlashes code, which router-dom does now
- [x] Add translations for Immediate, used in the Scheduler
- [x] Remove history package
- [x] Replace Functional Components (FC) with normal JSX functions if we're not passing down children props
- [x] Set initial data for `psram` in ApplicationSettings.tsx to prevent error when page is loaded and data not fetched from hardarestatus endpoint
- [x] Fix download URLs (e.g. S3 would try and find EMS-ESP-3_7_0-dev_30-ESP32_S3_16M.bin)
- [x] Prevent signIn username from forcing capslock on for first input character (on ios)
- [x] replace range() function used in APSettings to get rid of lodash library and save .9KB
- [x] prevent changing board profile if pre-loaded with BBQKees (in NVS)
- [x] Tidy up Application settings, moving to sections like Hardware, Formatting, Services

mattreim:
And maybe a few more translations.

proddy:
> And maybe a few more translations.

that would be nice. We really leave it up to the community to do the translations (I'm not fluent in 10 languages, unfortunately!). 

I'll do a broadcast and ask for people to help.

proddy:
Looks like Alova 3 is working again so I'll make a PR. Only remaining issue seems to be the Upload progress bar.

proddy:
For the Application page I was thinking 

```
Services
 - API
   - Bypass Access Token
 - Telnet (on/off)
 - Modbus (on/off)
   - Max Clients, Port, Timeout
 - Syslog (on/off)
   - Host, Port, Log level, Mark interval
 - Sensors
	- enable Analog Sensors
	- enable Temperature parasite power
 - Shower Timer
	- Wait time
	- enable shower alert
	  - Trigger Time, Cold shot duration
 
Formatting Options
[...]

PR #1931, opened 2024-08-10
Web changes #1920

proddy:
Many small web updates and optimizations as listed in #1920 

plus:

- fix WiFi scan with AP
- added Unit tests for console commands


MichaelDvP:
custom entities adding does notwork. Dialog is ok, but in the list the new entity is instantly removed (i guess from refresh).

proddy:
oops. will check.

MichaelDvP:
from #1920:
> Only remaining issue seems to be the Upload progress bar.

Seems the progress.total/loaded always stays zero. 
A fast workaround is check `isUploading` and just use text LL.UPLOADING and set progressbar to "indeterminate"
better than no feedback at all.

proddy:
> from #1920:
> 
> > Only remaining issue seems to be the Upload progress bar.
> 
> Seems the progress.total/loaded always stays zero. A fast workaround is check `isUploading` and just use text LL.UPLOADING and set progressbar to "indeterminate" better than no feedback at all.

I think that's a good and only workaround for now. I've informed the Alova team about it not working in v3 and then created some demo code to show it breaking (https://codesandbox.io/p/devbox/alova-test-upload-w46x26?file=%2Fsrc%2Fapi.ts) but it seems to be working fine, which is embarrassing. So something else is messing the react hooks & children propagation so next step is to change the demo code to also use a new component like our SingleUpload.tsx and see if the state is passed through correctly.

MichaelDvP:
The upload  button is imo a step backward. Drag&Drop was nice.
The progressbar still shows 0 and does not move.

proddy:
progressbar is fixed (was an alova bug). Drag&Drop was nice but not useful on phones or tablets. I also found it dangerous to use. And it added 12kb to the web bundle (6kb gzipped).

proddy:
ok, wrote my own Drag & Drop. Needs some more testing and validation but its nicer and easier

proddy:
Think it's finished. ready to commit. If you want @MichaelDvP to take one last check.

MichaelDvP:
The autorefresh is great, but still in custom entities it reverts all adds/changes made in the dialog.

Upload works good with drag/drop. The needs-reboot banner stay while the reboot circle is shows, that's confusing.

proddy:
Thanks for reviewing. I'll fix both those.

MichaelDvP:
In settings you use headline "general options" for "force heatingoff", " remote" and "shower", in the docs of emsesp.org the headline is "special funnctions". Maybe sync both headlines.

proddy:
> In settings you use headline "general options" for "force heatingoff", " remote" and "shower", in the docs of emsesp.org the headline is "special funnctions". Maybe sync both headlines.

I prefer "Special Functions". Shall I change it?

MichaelDvP:
> I prefer "Special Functions". Shall I change it?

yes, please, i also prefer it.

proddy:
done. If there are any other issues I can fix as we go along.

MichaelDvP:
I'm still not very happy with the new upload. 
Before it was 
- click or drop file: progress until 100% then restarts with only circle shown

Now its:
- scroll down to upload
- click or drop file: filename is shown
- scroll down to see the upload button
- press button upload: progress to 100%, nothing seems to happen after that
- scoll down to see the banner
- click restart: wait, than there is only a toast-message "network disconnected", than the page is reloaded.

Most users does not update often, so it's ok. But for me it's annoying to make 6 interaction instead of 1.

proddy:
I'll have to think about how to make it more intuitive. When the progress hits 100% it should be replaced with a Restart button or a banner. What do you suggest? The Network Disconnected is something I need to fix, related to the timeout I thnk. 

A few other points:

- Upload is not a highly used feature, as you said. I never use it. For developers it's faster to click on the upload button in VSCode or execute a `pio run -t upload`.
- This is why Upload is at the bottom of the page and Downloads are shown first
- Showing the filename is important. Before you could drag and drop a hundred files and not know which one will be loaded. Selecting the wrong one could brick the system.
- Showing the upload button adds another safety check in case users accidentally does select the wrong file. It's common on systems (esphome, tasmota, amazon alexa, sonos, unifi all do this)
- Another idea is to add an automatic update like "Click to upgrade to the latest version" which does an automatic download and then upload & restart?
- It saves 6kb in flash this way. We need it for the new Web code that's coming (favourite dashboard, PV dashboard, timer/switch mapping). I also spent two whole days doing it!

proddy:
made some minor improvements. Don't get the Network Timeout though. If you still see this (after a browser cache refresh or restart) let me know. Timeout is 1 minute (api/system.ts).

MichaelDvP:
> Showing the filename is important. Before you could drag and drop a hundred files and not know which one will be loaded. Selecting the wrong one could brick the system.

Normally not, the bin is checked for right processor binary and size. maybe you can upload a tasmota esp32 binary, but i think this will run and is not  bricking.

proddy:
I was referring to json setting files or schedule configurations

PR #1939, opened 2024-08-16
add hint for manual workaround to preserve nrg values to CHANGELOG_LATEST.md

paludi:
added hint to manually preserve value of dhw energy (nrgww)

### DIFF
--- a/CHANGELOG_LATEST.md
+++ b/CHANGELOG_LATEST.md
@@ -5,7 +5,7 @@
 ## **IMPORTANT! BREAKING CHANGES with 3.6.5**
 
 - new device WATER shows dhw entities from MM100 and SM100 in dhw setting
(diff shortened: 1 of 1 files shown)


MichaelDvP:
I think we add a conversion. The value is stored in nvs with the shortname, that has changed due to dhw-prefix form nrgww to nrg.
In boiler 
```
        nrgWwF_   = EMSESP::nvs_.getDouble(FL_(nrgWw)[0], 0);
        if (nrgWwF_ == 0) { // convert values form 3.6.5
            nrgWwF_ = EMSESP::nvs_.getDouble("nrgww", 0);
        }
```
using additional nvs_.remove("nrgww") will prevent downgrading, but would save a little nv storagespace.
Or we use a fixed name for nvs, not same as mqtt shortname?

Issue #1933, opened 2024-08-11
correct firmware naming


proddy:
As Michael reported in #1930

The S32 gateways (without S3) have 16M flash without PSRAM, It can't handle the 16M file. 
The 16M file is compiled with [-DEMSESP_DEFAULT_BOARD_PROFILE="E32V2](https://github.com/emsesp/EMS-ESP32/blob/5b60ec1836d2745b53e8a63b7af8b819ce07454d/platformio.ini#L116) and defaults on factory reset to the wrong board-pins when flashed to S32.

_Originally posted by @MichaelDvP in https://github.com/emsesp/EMS-ESP32/issues/1930#issuecomment-2282749387_

Today it looks like:

![image](https://github.com/user-attachments/assets/16d010d9-95d9-44e5-bbc3-e7558ddf68b8)

proddy:
I suggest we change the platformio env names, the GitHub actions and the Web firmware download code to build and check for these variants:

`EMS-ESP-<version>-<chip type>-<flash mem>.bin`

`chip type` is S, S3, perhaps C3
`flash mem` is 4, 16, 32

do we need to explicitly check for PSRAM too?

@MichaelDvP I am missing anything?

MichaelDvP:
I think we should check the common chips and the BBQKees gateways, form old to new:
1. gateway premium - ETH-live-Mini, ESP32, 4M
2. gateway premium - LiligoS3: ESP32S3, 16M, PSRAM
3. gateway premium Lolin Mini C3 or Mini S2
4. gateway S32 - nodeMCU ESP32, 4M
5. gateway E32 - ESP32, 4M
6. gateway S32 - soldered chip, ESP32, 16M
7. gateway S3 - ESP32S3, 16M, PSRAM
8. gateway E32V2 - ESP32, 16M, PSRAM
9. ems-interface with common modules, nodeMCU, ETH-Live, Liligo, etc.

- ESP32S3 firmware works with all S3 cips (2. and 7.), defaults to board gateway S3, on the Liligo chip LED is not working. need configuring, but boots upwith AP.
- ESP32-16M frmware only fits to the E32V2, bootloop due to low ram on the ESP32 without PSRAM (tested with the gateway 6.)
- ESP32 firmware is updateable without issue to v3.6.5 and works for all ESP32 versions.

For C3 and S2 we don't have prcompiled firmare. 
Also for 32M chips, there was only a S3-32M with OPI-flash that needs a special compile env.

[...]

PR #1941, opened 2024-08-16
correct firmware naming #1933

proddy:
See https://github.com/emsesp/EMS-ESP32/issues/1933 for the description.

New naming is

`EMS-ESP-<version>-<chip type>-<flash mem>[+].bin`

`chip type` is as defined in esp_platform from pio. `ESP32`, `ESP32S3`, `ESP32C3` etc. All uppercase.
`flash mem` is either `4MB` or `16MB`.
`+` is added at the end if there is PSRAM


platformio.ini
```diff
@@ -81,31 +81,33 @@ lib_ldf_mode = chain+
...
-board_build.extra_flags = -DBOARD_HAS_PSRAM
+build_flags =
+  ${espressi32_base_tasmota.build_flags}
+  '-DEMSESP_DEFAULT_BOARD_PROFILE="S32"'
 
-[env:ci_s3]
-; 16MB ESP32-S3 - like BBQKees S3
+[env:ci_s_16M]
+; 16MB ESP32 - no PSRAM - like some of the older BBQKees S32
 extends = espressi32_base
```
MichaelDvP, line 101: without psram it should use the tasmota base to save ram.

proddy: good idea. I modified it.

platformio.ini
```diff
@@ -81,31 +81,33 @@ lib_ldf_mode = chain+
...
+[env:ci_s_4M]
+; 4MB ESP32 - using Tasmota - no SSL, no PSRAM - like the BBQKees older S32 models
 extends = espressi32_base_tasmota
 extra_scripts = scripts/rename_fw.py
 board = esp32dev
 board_build.partitions = esp32_partition_4M.csv
-board_build.extra_flags = -DBOARD_HAS_PSRAM
+build_flags =
+  ${espressi32_base_tasmota.build_flags}
+  '-DEMSESP_DEFAULT_BOARD_PROFILE="S32"'
```
MichaelDvP, line 97: ci_s_4M could be S32 or E32 gateway, there should not be a defined DEFAULT_BOARD_PROFILE to trgger the automatic detection of E32 ethernet.

proddy: good spot. I'll change!

Issue #1938, opened 2024-08-16
nrgtotal, nrgheat, dhw.nrg are not reported correctly in 3.7.0-dev.30 compared to 3.6.5


paludi:
### PROBLEM DESCRIPTION

After updating to 3.7.0-dev.30, the reported values of nrgtotal, nrgheat, dhw.nrg from the boiler (Buderus GB172 in my case) change compared to 3.6.5:
reported values in 3.7.0-dev.30:
- "nrgtotal": 793.71,
-  "nrgheat": 793.71,
- "dhw.nrg": 0 

3.6.5:
-  "total energy (nrgtotal)": 9647.08,
-   "energy heating (nrgheat)": 793.71,
-   "dhw energy (nrgww)": 8853.37

It looks like 
- nrgheat is still reported correctly
- nrgtotal now reports the value of nrgheat instead of the total energy consumed
- energy for hot water (dhw.nrg) is not reported any longer. 

### REQUESTED INFORMATION

_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [ x] Searched the problem in [issues](https://github.com/emsesp/EMS-ESP32/issues)
- [x ] Searched the problem in [discussions](https://github.com/emsesp/EMS-ESP32/discussions)
- [x ] Searched the problem in the [docs](https://emsesp.github.io/docs/Troubleshooting/)
- [x ] Searched the problem in the [chat](https://discord.gg/3J3GgnzpyT)
- [ x] Provide the output of http://ems-esp.local/api/system : see attached files at bottom for both releases

```lua
  System information output here:
```

```
logs of dev.30


2024-08-16 07:20:52.547 W 671: [telegram] Incomplete Rx: 10 0B FF 00 01 40 41 00 03 00 00 00 00 00 FF 01 F4 06 FF FF 00 00 00 00 00 00 AA
2024-08-16 07:21:48.799 W 672: [telegram] Incomplete Rx: 10 0B FF 00 01 69 11 01 01 11 89
2024-08-16 07:22:54.945 W 673: [telegram] Incomplete Rx: 2A 0B FF 00 06 E0 00 00 00 00 00 00 00 00 03 00 00 00 00 00 00 03 03 03 03 90
2024-08-16 07:23:47.909 W 674: [telegram] Incomplete Rx: 10 0B FF 00 01 94 00 00 00 00 FF 00 E4 FE 90
[...]

PR #1942, opened 2024-08-16
some small fixes

MichaelDvP:
- for #1938 i set the nvs names of 3.6.5 fixed. There is no need to sync with mqtt shortnames.
- The reboot to loader/partiton was broken, i added the option to use boot or other ota partition
- fixes for the raw custom entities tested with Thomas
- The scheduler conditions log a lot of errors after reboot because reading bus entities (e.g. mixer/pumpstatus) reports error until device is detected and telegrams received. I just added a deadteime of 1 minute after reboot.
- For soc without PSRAM the scheduler is now sync, for psrm versions i've made the mqtt async.


proddy:
The UI dialog is really confusing to a non-techie that just wants to restart EMS-ESP

image

I suggest either removing partiton/boot or consider putting them in a drop-down with an explanation on what they are what happens when you select them.

MichaelDvP:
We can remove. The boot/loader was there before, but not working anymore.
The restart with optional partiton name is in telnet more usefull (if telnet is reachable).

PR #1951, opened 2024-08-18
upload and install firmware automatically

proddy:
I thought I'd sneak this change in before my vacation. It adds the option to download and install the firmware automatically. UI looks like:

image

It works well when calling the function from the console with `test upload`. But calling it from within the Web instance by clicking on the button causes a WDT

```
$ E (253289) task_wdt: Task watchdog got triggered. The following tasks did not reset the watchdog in time:
E (253289) task_wdt:  - async_tcp (CPU 0/1)
E (253289) task_wdt: Tasks currently running:
E (253289) task_wdt: CPU 0: ipc0
E (253289) task_wdt: CPU 1: loopTask
E (253289) task_wdt: Aborting.
```

I think we need to close the connection (I tried this) but something is still blocking it. And I'm out of ideas. If you have time @MichaelDvP can you take a look?

See the code in `UploadFileService::uploadURL()` which calls `System::uploadFirmwareURL()`. Compile with -DEMSESP_TEST and then run `test update`. 

Note I hardcoded the calls to the GitHub API in `DownloadUpload.tsx` (lines 110-123) as GH was blocking them since I was testing too much!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,9 +47,6 @@ interface/src/i18n/i18n-types.ts
 interface/src/i18n/i18n-util.ts
 interface/src/i18n/i18n-util.sync.ts
 interface/src/i18n/i18n-util.async.ts
-# mock-api uploads
-*/uploads/*
-!uploads/README.md
 
 # scripts
 test.sh
--- a/interface/package.json
+++ b/interface/package.json
@@ -21,7 +21,7 @@
     "lint": "eslint . --fix"
   },
   "dependencies": {
-    "@alova/adapter-xhr": "2.0.4",
+    "@alova/adapter-xhr": "2.0.5",
     "@emotion/react": "^11.13.0",
     "@emotion/styled": "^11.13.0",
     "@mui/icons-material": "^5.16.7",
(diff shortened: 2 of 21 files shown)


MichaelDvP:
> I think we need to close the connection (I tried this) but something is still blocking it. And I'm out of ideas. 

You call emsesp::EMSESP::system_.uploadFirmwareURL(url.c_str()) within the webserver callback. 
Try to set a system_.upload() flag and check in the main loop if it is set to execute the download in this context.

proddy:
doh! that's it. I'll make a quick change. Thanks Michael

proddy:
Works now. Thanks. Please test and see if it needs any more work.

MichaelDvP (approved):
Code looks fine and working test with a spare esp32S3 successful.

proddy:
thanks for checking. I'll merge. Any more changes you think we need for the firmware download to reduce your 6 additional steps? Perhaps skip the 'upload' step and automatically upload the .bin file when selected/dragged?

MichaelDvP:
Just a reminder here: The "install" button hardcoded and needs translations.

PR #1959, opened 2024-08-23
Update German translation

mattreim:


proddy:
Thanks, I'm making changes to the translations so to avoid a merge conflict I changed these manually in my next PR.

PR #1991, opened 2024-09-08
some minor fixes

proddy:
- GPIO check didn't work on Ethernet as it's tested before the ethernet is active. I had uploaded an old customization file (with GPIO21) and it put my E32V2 into an annoying boot loop
- Check to see if /config exists on LittleFS wasn't working. So a factory reset of the MQTT settings was happening each time after a `call system format`
- NVS boot overrides any Board Profile set, on BBQKees boards
- Make LOG WebUI look pretty
- limit LOG for ESP32s with no psram
- exit command from Serial not working on Serial connections (ptty, minicom etc..)
- new `log show` command


proddy:
Update the serial console. Now when a USB is connected you have to ctrl-C to create the Shell instance. exit will remove it. This saves a few KBs heap

MichaelDvP:
> Update the serial console. 

Nice, I would prefer to autostart the console on debug builds (not only standalone).

proddy:
> > Update the serial console.
> 
> Nice, I would prefer to autostart the console on debug builds (not only standalone).

sure, will add. You ok with the other changes?

platformio.ini
```diff
@@ -128,28 +128,28 @@ board_build.extra_flags = -DBOARD_HAS_PSRAM '-DEMSESP_DEFAULT_BOARD_PROFILE="S32
 ;
 
 [env:s_4M]
-extends = espressi32_base_tasmota
+extends = espressif32_base_tasmota
 board = esp32dev
 board_upload.flash_size = 4MB
 board_build.partitions = esp32_partition_4M.csv
 board_build.extra_flags = '-DEMSESP_DEFAULT_BOARD_PROFILE="S32"'
```
MichaelDvP, line 135: platformIO is a beast, putting the boardprofile here instead of build_flags disables the platform menu in vscode. Can not erase a chip.

src/emsesp.cpp
```diff
@@ -1541,52 +1541,42 @@ void EMSESP::start() {
...
-    shell_->maximum_log_messages(100);
-    shell_->start();
-#if defined(EMSESP_DEBUG)
-    shell_->log_level(uuid::log::Level::DEBUG);
-#else
-    shell_->log_level(uuid::log::Level::TRACE);
-#endif
+
+// always start a serial console if we're running standalone, except if we're running unit tests
 #if defined(EMSESP_STANDALONE)
```
MichaelDvP, line 1548: As written i would prefer `#if defined(EMSESP_STANDALONE) || defined(EMSESP_DEBUG)`

src/web/WebSettingsService.cpp
```diff
@@ -88,90 +87,109 @@ void WebSettings::read(WebSettings & settings, JsonObject root) {
...
-                    (int8_t)(root["tx_gpio"] | 5),
-                    (int8_t)(root["pbutton_gpio"] | 0),
-                    (int8_t)(root["phy_type"] | PHY_type::PHY_TYPE_NONE),
-                    (int8_t)(root["eth_power"] | 0),
-                    (int8_t)(root["eth_phy_addr"] | 0),
-                    (int8_t)(root["eth_clock_mode"] | 0)};
+#ifdef EMSESP_DEBUG
+    EMSESP::logger().debug("NVS boot value is %s, board profile is %s, EMSESP_DEFAULT_BOARD_PROFILE is %s",
+                           EMSESP::nvs_.getString("boot").c_str(),
+                           root["board_profile"].as<const char *>(),
```
MichaelDvP, line 102: This crashes on empty FS, const char* returns a nullptr. Better `root["board_profile"].as<String>().c_str(),`

src/web/WebSettingsService.cpp
```diff
@@ -88,90 +87,109 @@ void WebSettings::read(WebSettings & settings, JsonObject root) {
...
-            System::load_board_profile(data, settings.board_profile.c_str());
-        }
-        EMSESP::logger().info("No board profile found. Re-setting to %s", settings.board_profile.c_str());
+
+        // apply the new board profile setting
+        System::load_board_profile(data, settings.board_profile.c_str());
+        EMSESP::logger().warning("No Board profile setup - using %s", settings.board_profile.c_str());
     } else {
-        EMSESP::logger().info("Loading board profile %s", settings.board_profile.c_str());
+        if (old_board_profile != settings.board_profile) {
```
MichaelDvP, line 188: Here we need also a ` System::load_board_profile` to initialite the data.
Otherwise crashes with empty FS.

MichaelDvP (changes requested):
I've tried the init with serial console `call system format` and found some crashes mared in code.

src/emsesp.cpp
proddy: done

platformio.ini
proddy: I still see the erase option in the menu

alt text

src/web/WebSettingsService.cpp
proddy: I had it as `.as<String>().c_str()` first but it breaks standalone

proddy: I'll find another way

MichaelDvP: will this also break: `as<std::string>().c_str()`?

src/web/WebSettingsService.cpp
proddy: don't think its needed, the else is from the other if. its hard to follow with all the #ifdefs.

src/web/WebSettingsService.cpp
proddy: fixed.

proddy:
> I've tried the init with serial console `call system format` and found some crashes mared in code.

That's what I've been doing all day. `call system format` - but I don't get any crashes. I'm testing on an S32?

src/web/WebSettingsService.cpp
proddy: I can't get it to crash - if you could please re-test to make sure I'm not missing anything

proddy:
Add a new cmd `log show`, useful for seeing missing logs. like:

image

platformio.ini
```diff
@@ -95,7 +96,7 @@ board_build.extra_flags = '-DEMSESP_DEFAULT_BOARD_PROFILE="S32"'
```
MichaelDvP, line 96: The 4M version could be S32 or E32, for this we need the E32 test.

src/web/WebSettingsService.cpp
MichaelDvP: Youre right.

platformio.ini
MichaelDvP: Restarting vscode brings my Platform folder back, Don't now why it was not shown after first loading your changes.

src/web/WebSettingsService.cpp
```diff
@@ -88,90 +87,111 @@ void WebSettings::read(WebSettings & settings, JsonObject root) {
...
+#ifdef EMSESP_DEBUG
+        EMSESP::logger().debug("Trying to detect board and set board profile...");
+#endif
+
 #if defined(EMSESP_STANDALONE)
-            settings.board_profile = "S32";
+        settings.board_profile = "S32";
 #elif CONFIG_IDF_TARGET_ESP32
-            if (settings.board_profile == "" && !psram) { // empty: new test for E32
+        // check for no PSRAM, could be a E32 or S32
```
MichaelDvP, line 155: Ok, i think now i found where i get bootloops. If we have no -DEMSESP_BOARD_PROFILE in platformio.ini it goes to "default" and starts here the test. But the test sets only E32 or E32V2, not S32 otherwise, it leaves on "default". 
Set it here to "S32" and overwrite to E32/E32V2 if ETH found.

proddy: good find. I thought I defaulted to S32, but will adjust. thanks for testing

platformio.ini
MichaelDvP: No, the EMSESP_DEFAULT_BOARD_PROFILE is stored in code. If an E32 user updates the 4M version and make factory reset it will **now** default to Wifi-S32 instead of searching the ethernet. That's why the test code is included and 4M should have board_profile "default".

proddy: so is the best fix here is to not set a EMSESP_DEFAULT_BOARD_PROFILE in pio for `ci_s_4M` and let the code run the 'default' logic?

src/web/WebSettingsService.cpp
proddy: Hopefully made a fix. i don;t have a E32 so can't test

mattreim:
Update German translation: 

```lua
RESTARTING_POST: 'Vorbereiten'
```

platformio.ini
MichaelDvP: > so is the best fix here is to not set a EMSESP_DEFAULT_BOARD_PROFILE in pio for ci_s_4M and let the code run the 'default' logic?

Yes.

src/web/WebSettingsService.cpp
MichaelDvP: Yes, this also works.

src/web/WebSettingsService.cpp
```diff
@@ -88,90 +87,115 @@ void WebSettings::read(WebSettings & settings, JsonObject root) {
...
+            if (ETH.begin(ETH_PHY_LAN8720, 0, 23, 18, 15, ETH_CLOCK_GPIO0_OUT)) {
 #endif
-                    EMSESP::nvs_.putString("boot", "E32V2");
-                } else {
-                    EMSESP::nvs_.putString("boot", "S32");
-                }
+                settings.board_profile = "E32V2";
             } else {
-                EMSESP::nvs_.putString("boot", "S32");
+                settings.board_profile = "S3";
```
MichaelDvP, line 175: This is section TARGET_ESP32, t could never be a S3-chip (what is named `S32S3`), a S3 board profile does not exist. This should also default to S32 or CUSTOM with S32 identical gpios.

lib/framework/NetworkSettingsService.cpp
```diff
@@ -354,31 +354,39 @@ void NetworkSettingsService::WiFiEvent(WiFiEvent_t event, WiFiEventInfo_t info)
...
-        }
+        emsesp::EMSESP::logger().info("Local IPv6 (WiFi)=%s", WiFi.localIPv6().toString().c_str());
 #else
         emsesp::EMSESP::logger().info("Local IPv6=%s", IPAddress(IPv6, (uint8_t *)info.got_ip6.ip6_info.ip.addr, 0).toString().c_str());
 #endif
         emsesp::EMSESP::system_.has_ipv6(true);
         break;
 
+        // IPv6 specific
+        // This a bug in arduino where this is triggered twice, so we prevent it
```
MichaelDvP, line 378: This is not a bug, it is triggered on every IP6-address, normally twice for local and global. I have set an ULA address in router and get three addresses. (local, global, ula). Also every night the router gets a new global prefix and this is triggert once with the new global address.
With TASMOT_SDK it shows the address that is new (the logging `Local` is wrong here.
From my log:
```
2024-09-10 10:19:56.123 I 15: [emsesp] Local IPv6=fe80::ea31:cdff:fefd:4308
2024-09-10 10:19:59.123 I 16: [emsesp] Local IPv6=2003:c2:4728:3f00:ea31:cdff:fefd:4308
2024-09-10 10:19:59.123 I 17: [emsesp] Local IPv6=fd00::ea31:cdff:fefd:4308
```

proddy: wow, didn't know you could have multiple IP addresses. Shall we keep the change I made and only show the first one? On my system it shows twice with the same IP.

MichaelDvP: The espressif arduino 2.xx only can show Local_IP6, but also gets the others. For IDF5 and tasmota-arduino we can show the ip that is transmitted to the GOT_IP6. You always have a local (FE80::) and a global address. In my case 
`::ea31:cdff:fefd:4308` is the device part, `fe00::` the local address prfix, `2003:c2:4728:3f00:` the global prefix and `fd00::` the ULA (Unique Local Address, fd:: is reseved for ULA).

proddy: thanks for explaining that

proddy:
This is been in review for a few days and the changes are pilling up, so I'll merge and we can fix in the next dev release.

SoftVienka:
Update Slovak translation:

RESTARTING_POST: 'Príprava'

proddy:
> Update Slovak translation:
> 
> RESTARTING_POST: 'Príprava'

thanks. added in my latest PR (dev-36)

Issue #1652, opened 2024-03-09
Fix missing log entries in web


proddy:
There is a chance of log messages being skipped when displayed on the Web specially on slow networks.

MichaelDvP suggested the following:

"Maybe set a longer timeout (300-500ms) in EMS-ESP and post back the message-id on an event received. If the post is received in EMS-ESP, the next message-event can be sent immediately; otherwise, you can just repeat the message after a timeout. It could also help to pack all queued messages into one event as json-array, which reduces the number of events."

which is a perfect solution.

_Originally posted by @MichaelDvP in https://github.com/emsesp/EMS-ESP32/issues/1611#issuecomment-1986835773_

proddy:
There was a bug in the React code, which I fixed in the 3.7.0-dev.1, where the EventStream could be initiated more than once when the web page is mounted, which could account for some messages getting lost. I'll see if this helps with this issue by slowing down the WiFi on a real ESP32 next,.

proddy:
`WebLogService.h` had REFRESH_SYNC as 50, I changed to 30 and it may have caused #1611 missing logs, then it was changed to 80. I'll change it back to 30 and see if the fix in the web solves the sync issue.

If you compile with `-DEMSESP_PINGTEST` it will send a log message every 1/2 second, which i used for testing.

MichaelDvP:
Wow, no missing logs with 30ms on my old notebook with compiler running in background. I think you can close the issue. Also no doubletts when removing the id checks.

proddy:
> Wow, no missing logs with 30ms on my old notebook with compiler running in background. I think you can close the issue. Also no doubletts when removing the id checks.

Great, I close this one then. Was a silly thing in how React renders the DOM on page rendering. Before it was easier with onMount().

MichaelDvP:
I've also tested with wait-delay completly removed and buffer 100. **Works good.** Next i send all buffer messages at once (in one loop), works also, but maybe this blocks the main loop too long and it takes much memory for TCP buffering. Better leave it by one message per loop.

proddy:
@MichaelDvP in 3.7.0-dev.36 I've removed the buffer delay after updating AsyncEventSource with this: https://github.com/mathieucarbou/ESPAsyncWebServer/commit/0163e043747d3ad8cbe70da8e78b88cc652ec12b

with Ping test enabled, and setting the browser to slow 3G network and high CPU throttle I'm not missing any incoming messages.

PR #2006, opened 2024-09-11
fix typos, optimize web log

proddy:
- removed delay in weblog
- applied fix to AsyncEventSource (https://github.com/emsesp/EMS-ESP32/issues/1652#issuecomment-2344643625)


interface/src/app/status/SystemLog.tsx
```diff
@@ -87,6 +87,7 @@ const SystemLog = () => {
 
   const [logEntries, setLogEntries] = useState<LogEntry[]>([]);
   const [lastIndex, setLastIndex] = useState<number>(0);
```
MichaelDvP, line 89: You can remove the last_index check, it was to prevent double messages, but this is solved.

interface/src/app/status/SystemLog.tsx
```diff
@@ -87,6 +87,7 @@ const SystemLog = () => {
...
+  const [autoscroll, setAutoscroll] = useState(true);
```
MichaelDvP, line 90: I like this autoscroll checkbox.

interface/src/app/status/SystemLog.tsx
```diff
@@ -112,7 +113,7 @@ const SystemLog = () => {
   });
 
   onError(() => {
-    toast.error('No connection to Log server');
+    toast.error('No connection to Log service');
   });
 
   // called on page load to reset pointer and fetch all log entries
```
MichaelDvP, line 119: You can remove the fetch request and reset the pointer on getValue, this is also called on page open.

interface/src/app/status/SystemLog.tsx
```diff
@@ -222,6 +223,16 @@ const SystemLog = () => {
...
             />
+            <BlockFormControlLabel
+              control={
+                <Checkbox
+                  checked={autoscroll}
+                  onChange={() => setAutoscroll(!autoscroll)}
+                  name="autoscroll"
+                />
+              }
+              label="Auto scroll"
```
MichaelDvP, line 234: Mark it with "TODO Translate"

src/web/WebLogService.cpp
```diff
@@ -183,11 +183,13 @@ void WebLogService::loop() {
...
     }
 
-    // put a small delay in
+    /*
+    // put a small delay in - https://github.com/emsesp/EMS-ESP32/issues/1652
     if (uuid::get_uptime_ms() - last_transmit_ < REFRESH_SYNC) {
         return;
     }
     last_transmit_ = uuid::get_uptime_ms();
+    */
```
MichaelDvP, line 192: Confirm that it working on my slow laptop, you can completly remove the delay.

MichaelDvP (approved):

interface/src/app/status/SystemLog.tsx
proddy: not sure I understand. the pointer needs to be reset when the page loads

interface/src/app/status/SystemLog.tsx
proddy: added the translation

interface/src/app/status/SystemLog.tsx
proddy: I understand now. Will fix.

mattreim:
Update German translation: 

```lua
AUTO_SCROLL: 'Automatisches Scrollen' 
```

Issue #1767, opened 2024-05-16
Worcester-Bosch Greenstar Summer Mode temp value upper limit too low


V4ler1an:
### PROBLEM DESCRIPTION

The 'Summer Mode' temperature setting in the Worcester Bosch Greenstar Weather Compensation settings currently has an upper limit of only 30 degrees C. When the Weather Compensation feature is enabled on the boiler and the 'Boiler Outside Temperature' thermostat value meets or exceeds this 'Summer Mode' temperature value, the boiler will disable any response to central heating demand - as designed.

The problem (at least for the way my boiler is installed), is that activation of Summer Mode will also prevent the boiler from heating my hot water tank. This appears to be because the boiler has not been connected up to use its proper designed DHW features. These are all unused in the boiler settings. Intead if has a setup which relies on the central heating features and (presumably) some sort of separate hot water tank valve which opens when the Hive Home smart controls demand hot water.

Long story short, the upper summer mode temp limit will mean that for certain periods in the summer, if the outside temp reaches or exceeds 30 degrees C, I won't be able to heat any hot water. The workaround required would be to disable the weather compensation curve feature entirely, which disables Summer Mode.

Could I please have the 'boiler summertime temperature' value limit increased to 45 degrees C?

### TO REPRODUCE

Attempt to increase the 'boiler summertime temperature' value above 30 degrees C.

### EXPECTED BEHAVIOUR

Current behaviour is as expected but undesirable.

### SCREENSHOTS

Home Assistant automation failure due to Summertime Mode temperatire value limit exceeded:
![image](https://github.com/emsesp/EMS-ESP32/assets/134781183/8c0709a7-21e8-497e-9bce-4935077ffb0c)

### ADDITIONAL CONTEXT
I appreciate that had the central heating installers installed my boiler to use its built-in water heating (DHW) features to the full, hot water would be heated separately from central heating water. Tbh, I don't know the reasons for installing it this way. Nevertheless, as I see it, increasing the Summer Mode value limit shouldn't really cause any problems.

Note: this Summer Mode feature was introduced to the EMS-Bus Gateway firmware several months ago when I requested that you add support for the official Worcester Bosch weather compensation add-on in this issue I raised: 

https://github.com/emsesp/EMS-ESP32/issues/1642

proddy:
that's like a 2 second change. I'll do it ASAP

proddy:
in dev11. Max is 45

V4ler1an:
[...]

PR #2010, opened 2024-09-13
update doc in prep for 3.7.0 release

proddy:


mattreim:
Small question according to locale_translations.h there is only one summertemp and now I find two with different values. Is that correct?

Boiler:
Summertemp_boiler
Thermostat:
Summertemp_hc1

proddy:
> Small question according to locale_translations.h there is only one summertemp and now I find two with different values. Is that correct?
> 
> Boiler: Summertemp_boiler Thermostat: Summertemp_hc1

yes that's correct. Some boilers have the Summer Time Mode. I'm not sure which one has precedence. Example at #1767

mattreim:
Thank you, I was just wondering.

PR #2018, opened 2024-09-16
network

proddy:
- Show all IPv6 addresses
- Create copy of firmware bins so older 3.6.4/3.6.5 can find the binaries
- Experimental fix for Eth not starting

### DIFF
--- a/lib/framework/NetworkSettingsService.cpp
+++ b/lib/framework/NetworkSettingsService.cpp
@@ -312,7 +312,7 @@ void NetworkSettingsService::WiFiEvent(WiFiEvent_t event, WiFiEventInfo_t info) 
 
     case ARDUINO_EVENT_WIFI_STA_GOT_IP:
         char result[10];
-        emsesp::EMSESP::logger().info("WiFi connected (IP=%s, hostname=%s, TxPower=%s dBm)",
+        emsesp::EMSESP::logger().info("WiFi connected (Local IP=%s, hostname=%s, TxPower=%s dBm)",
                                       WiFi.localIP().toString().c_str(),
                                       WiFi.getHostname(),
                                       emsesp::Helpers::render_value(result, ((double)(WiFi.getTxPower()) / 4), 1));
@@ -330,7 +330,7 @@ void NetworkSettingsService::WiFiEvent(WiFiEvent_t event, WiFiEventInfo_t info) 
     case ARDUINO_EVENT_ETH_GOT_IP:
         // prevent double calls to mDNS
         if (!emsesp::EMSESP::system_.ethernet_connected()) {
-            emsesp::EMSESP::logger().info("Ethernet connected (IP=%s, speed %d Mbps)", ETH.localIP().toString().c_str(), ETH.linkSpeed());
+            emsesp::EMSESP::logger().info("Ethernet connected (Local IP=%s, speed %d Mbps)", ETH.localIP().toString().c_str(), ETH.linkSpeed());
             emsesp::EMSESP::system_.ethernet_connected(true);
             mDNS_start();
         }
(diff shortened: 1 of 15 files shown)


src/system.cpp
```diff
@@ -788,6 +788,13 @@ void System::network_init(bool refresh) {
...
     auto clock_mode = (eth_clock_mode_t)eth_clock_mode_;
 
+    // reset power and add a delay as ETH doesn't not always start up correctly after a warm boot
+    // TODO still experimental
+    pinMode(eth_power_, OUTPUT);
+    digitalWrite(eth_power_, LOW);
+    delay(1000);
+    digitalWrite(eth_power_, HIGH);
+
 #if ESP_IDF_VERSION_MAJOR < 5
```
MichaelDvP, line 798: Does this work? 
The comment says that eth_power can also be -1, this should be checked

lib/framework/NetworkSettingsService.cpp
```diff
@@ -364,27 +366,24 @@ void NetworkSettingsService::WiFiEvent(WiFiEvent_t event, WiFiEventInfo_t info)
...
-            emsesp::EMSESP::logger().info("Local IPv6 (Eth)=%s", ETH.localIPv6().toString().c_str());
+        emsesp::EMSESP::logger().info("Local IPv6 (Eth)=%s", ETH.localIPv6().toString().c_str());
 #else
-            emsesp::EMSESP::logger().info("Local IPv6=%s", IPAddress(IPv6, (uint8_t *)info.got_ip6.ip6_info.ip.addr, 0).toString().c_str());
+        emsesp::EMSESP::logger().info("IPv6 (Eth)=%s", IPAddress(IPv6, (uint8_t *)info.got_ip6.ip6_info.ip.addr, 0).toString().c_str());
 #endif
-            emsesp::EMSESP::system_.has_ipv6(true);
-        }
+        emsesp::EMSESP::system_.has_ipv6(true);
         break;
```
MichaelDvP, line 387: I've tested how to show the different types of ip6 addresses also for arduino 2.xx, try this:
```
    case ARDUINO_EVENT_ETH_GOT_IP6: {
#if !TASMOTA_SDK && ESP_IDF_VERSION_MAJOR < 5
        auto ip6 = IPv6Address((uint8_t *)info.got_ip6.ip6_info.ip.addr).toString();
#else
        auto ip6 = IPAddress(IPv6, (uint8_t *)info.got_ip6.ip6_info.ip.addr, 0).toString();
#endif
        if (ip6.startsWith("fe80")) {
            emsesp::EMSESP::logger().info("IPv6 local: %s", ip6.c_str());
        } else if (ip6.startsWith("fd")) {
            emsesp::EMSESP::logger().info("IPv6 ULA: %s", ip6.c_str());
        } else {
            emsesp::EMSESP::logger().info("IPv6 global: %s", ip6.c_str());
        }
        emsesp::EMSESP::system_.has_ipv6(true);
    } break;
```

scripts/rename_fw.py
```diff
@@ -99,6 +96,28 @@ def bin_copy(source, target, env):
...
         file1.close()
         
+    # make a copy using the old 3.6.x filename format for backwards compatibility with the WebUI version check, e.g.
+    #  EMS-ESP-3_6_5-ESP32_S3.bin (16MB) with target ci_s3_16M_P
+    #  EMS-ESP-3_6_5-ESP32.bin (4MB) with target ci_s_4M
+    extra_variant = ""
+    if env.get('PIOENV') == "ci_s3_16M_P":
+        extra_variant = "EMS-ESP-" + app_version.replace(".", "_") + "-ESP32_S3"
+    elif env.get('PIOENV') == "ci_s_4M":
+        extra_variant = "EMS-ESP-" + app_version.replace(".", "_") + "-ESP32"
```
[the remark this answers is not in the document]
proddy: This is tricky one. I would mean creating a new ci build target for ci_s_4M_P (4MB with PSRAM). We didn't automatically build firmware for 3.6.5/main branch - the target `ci_16M` wasn't included in the GH action. So I think Kees used 3.7.0 - I'll need to ask him.

src/system.cpp
proddy: don't know! I must have done 30 uploads and hit the restart button at least 50 times and I can't reproduce it, even without this code.

lib/framework/NetworkSettingsService.cpp
proddy: nice, i'll add.

PR #2027, opened 2024-09-19
upgrade to ArduinoJson 7.2

proddy:
replacing the redundant `containsKey()`

According to https://arduinojson.org/news/2024/09/18/arduinojson-7-2/ it should use less memory, especially with Arrays, but I don't see this. I noticed it uses about 1k more heap


src/command.cpp
```diff
@@ -298,8 +298,8 @@ bool Command::set_attribute(JsonObject output, const char * cmd, const char * at
     if (attribute == nullptr) {
         return true;
     }
-    if (output.containsKey(attribute)) {
-        std::string data = output[attribute].as<std::string>();
+    if (output[attribute].is<std::string>()) {
```
MichaelDvP, line 301: Does this work? The attribute "value" can also be float or int, "writable" is bool, etc. We force it to read as string for output to "api_data".

src/console.cpp
```diff
@@ -552,8 +552,8 @@ static void setup_commands(std::shared_ptr<Commands> & commands) {
 
             if (return_code == CommandRet::OK) {
                 if (json.size()) {
-                    if (json.containsKey("api_data")) {
-                        String data = json["api_data"].as<String>();
+                    if (json["api_data"].is<String>()) {
```
MichaelDvP, line 555: better std::string

src/web/WebSchedulerService.cpp
```diff
@@ -245,7 +245,7 @@ void WebSchedulerService::publish(const bool force) {
     JsonDocument doc;
     bool         ha_created = ha_registered_;
     for (const ScheduleItem & scheduleItem : *scheduleItems_) {
-        if (!scheduleItem.name.empty() && !doc.containsKey(scheduleItem.name)) {
+        if (!scheduleItem.name.empty() && !doc[scheduleItem.name].is<const char *>()) {
```
MichaelDvP, line 248: depending on bool format this can be bool, int, const char*, better use JsonVariantConst

src/command.cpp
proddy: I change to JsonVariantConst. Thanks

proddy: all the unit tests pass so I think its ok

src/console.cpp
proddy: thanks - changed to std::string to keep consistent

PR #2040, opened 2024-09-22
Update device library

xarbit:
- Initially noticed a minor issue where my Bosch CS5800i was being recognized as CS6800i. Since both models belong to the same family, I updated the `device_library.h` to correctly reflect this.
- Additionally, identified that the Buderus MX300 and BOSCH Connect-Key K30 RF are likely the same devices. Since I have the K30RF, I included adjustments for this as well.

[emsesp_system_info-2.txt](https://github.com/user-attachments/files/17089058/emsesp_system_info-2.txt)


proddy:
nice, thanks!

Issue #2054, opened 2024-09-26
Feature: Custom Support info


proddy:
The current Help page shows
 - links to the wiki, discord and how to submit a request or report a bug in GitHub.
 - download links for support information and all values
 - a footer asking everyone to like the project in GitHub

The change

1. For users that have a "guest" role, they don't have access to many of the Settings so this information above should only be shown for "admin" users.
2. Allow for customized support information to be shown at the top of the Help page. You can upload a "support.json" file that contains some keys including "support_text" and "image_url" that are rendered in a 2-column table. This data is permanently stored in EMS-ESP and loaded at boot. If the file is not present, or badly formatted it will not be shown.
3. Remove the footer. I think people know it's open-source.

proddy:
#1 in 3.7.0
#2 to customize it via a JSON we'll add to 3.7.1

proddy:
![image](https://github.com/user-attachments/assets/3c03d235-4bc2-4d5a-8337-6115040a4b7c)

proddy:
Add 2 and 3 to 3.7.0.

You can create a JSON file like:

```json
{
  "type": "customSupport",
  "Support": {
  "html": [
    "This product is installed and managed by:",
    "",
    "<b>Bosch Installer Example</b>",
    "",
    "Nefit Road 12",
    "1234 AB Amsterdam",
    "Phone: +31 123 456 789",
    "email: support@boschinstaller.nl",
    "",
    "For help and questions please <a target='_blank' href='https://emsesp.org'>contact</a> your installer."
  ],
    "img_url": "https://emsesp.org/media/images/designer.png"
  }
}
```

and upload it via the WebUI. It will be displayed in the Help page for both Admin and Guest users

PR #2066, opened 2024-09-29
custom support page, action endpoint

proddy:
- implements #2054
- created a new POST endpoint for those random non-API commands allvalues, export, uploadURL, checkUpgrade and customSupport. This removes 3 other endpoints and gives us 2-3kb more available heap
- allvalues was an odd system command. It's now moved to the export commands, like the rest of the data downloads


proddy:
do you want to take a quick look @MichaelDvP - don't think anything is broken

MichaelDvP:
I can access the image from link in source code, but on the emsesp-page it is shown broken. 
Any idea what this is?
grafik

proddy:
Seems like a Firefox issue with caching. In Chrome/Edge I get a 304 return code and the image is displayed. Can you disable the cache in the browser and see what comes back? The fix maybe would be with either to move the sample image file off the cloudflare server (which is proxied) or see if I can add some additional headers. https://stackoverflow.com/questions/704561/ns-binding-aborted-shown-in-firefox-with-httpfox#:~:text=In%20one%20post%20a%20Firefox%20developer%20confirms

src/web/WebStatusService.cpp
```diff
@@ -161,16 +203,41 @@ void WebStatusService::checkUpgrade(AsyncWebServerRequest * request, JsonVariant
...
 }
 
-// returns data for a specific feature/settings as a json object
-void WebStatusService::exportData(AsyncWebServerRequest * request) {
-    auto *     response = new AsyncJsonResponse();
-    JsonObject root     = response->getRoot();
+// action = allvalues
+// output all the devices and the values
+void WebStatusService::allvalues(JsonObject output) {
+    JsonDocument doc;
```
MichaelDvP, line 212: Tis doc is never used

src/web/WebStatusService.cpp
```diff
@@ -188,13 +255,43 @@ void WebStatusService::exportData(AsyncWebServerRequest * request) {
         System::extractSettings(EMSESP_CUSTOMIZATION_FILE, "Customizations", root);
     } else if (type == "entities") {
         System::extractSettings(EMSESP_CUSTOMENTITY_FILE, "Entities", root);
+    } else if (type == "allvalues") {
+        allvalues(root);
```
MichaelDvP, line 259: This is not a setting, the `type:allvalues` in the beginning of the json file looks a bit odd. Maybe `root.clear()` before this or in line 241 if (type != "allvalues") {

proddy: fixed, thanks

proddy:
> Seems like a Firefox issue with caching. In Chrome/Edge I get a 304 return code and the image is displayed. Can you disable the cache in the browser and see what comes back? The fix maybe would be with either to move the sample image file off the cloudflare server (which is proxied) or see if I can add some additional headers. https://stackoverflow.com/questions/704561/ns-binding-aborted-shown-in-firefox-with-httpfox#:~:text=In%20one%20post%20a%20Firefox%20developer%20confirms

I tried with Firefox and can't reproduce. Maybe something else blocking it? PiHole?

MichaelDvP:
Same with cache deactivated, With Edge:
grafik

MichaelDvP:
> Maybe something else blocking it? PiHole?

My pihole is disabled, actually using Adguard-home, but it is not the cause. Seems to be a policy in my windows, any browser shows the same `strict-origin-when-cross-origin` error. From any android device/browser in the same network/gateway it works.

proddy:
I switched off the Browser Integrity Check in Cloudflare (where emsesp.org is hosted). See if that helps?

[Browser Integrity Check](https://developers.cloudflare.com/waf/tools/browser-integrity-check/#disable-browser-integrity-check)

proddy:
And if that doesn't work, it may be due to resizing the image so I can try with not using a <Box> component in Help.tsx but a plain old .

Does it also break if you run it `yarn standalone` ?

MichaelDvP:
> Does it also break if you run it yarn standalone ?

Yes. Firefox and Edge. both images from emesp.org with strict-origin-when-cross-origin
grafik

MichaelDvP:
Afair it's new to september updates of firefox and chromium engines. In Firefox setting about:config: `network.http.referer.defaultPolicy` to 0 or 1 (default 2) enables the image load. 
But i think many people will run into this issue sooner or later. Better find another way to embed the image.

proddy:
that would explain it. What bugs me is that I can't reproduce it, on either Windows or Mac with Firefox 128 and also 130.0.1

MichaelDvP:
I think because you have loaded it once and it's in cache. You have to set `[x] cache deactivated` in developer tools.
(I use Firefox 130.0.1)

proddy:
I tried that on both Firefox & Edge. Cleared cache, disabled, and it still fetches correctly with a 200 status code. It'll be difficult for me to fix without being able to reproduce it.

This EMS-ESP feature includes being able to set a URL for image for Installers/Maintenance people. So we need to make it work somehow. An alternative solution is to able users to upload an image and we store it in LittleFS, and convert it which will be more programming.

I'm still not 100% sure it isn't Cloudflare doing something when proxing the inbound requests. So to rule that out could you please try this: 

change the `img_url` in `function custom_support()` in the file `rest_server.ts` to
```js
  img_url: 'https://picsum.photos/200/300'
```
and then run `yarn standalone` to see if you get random images shown in the Help page. This works for me on all browsers.

MichaelDvP:
Yes, this is working. 
grafik

proddy:
> Yes, this is working. grafik

thanks for checking. Its on cloudflare then. I'll find a new dummy image to use

proddy:
I managed to reproduce it, on all browsers now when I use a WiFi hotspot with my phone's 5G. And i think I fixed it by just adding `referrerPolicy="no-referrer"` to the `` component. Can you test again @MichaelDvP ?

Issue #1958, opened 2024-08-19
Feature: Dashboard showing all data (favourites, sensors, custom)


proddy:
> personally, I would call it Dashboard and have it above the menu options Devices & Sensors. It would only show favourite entities and all sensors that are enabled and have values. No edits and no click-throughs. Just keeping it simple for now. And this could be the default page when the app is loaded instead of /devices

Cant describe it any better...that way around would be optimal to my eyes.

_Originally posted by @dleding in https://github.com/emsesp/EMS-ESP32/discussions/1911#discussioncomment-10223260_

proddy:
Moving this idea fro m the Discussions to Issues so we refine the requirements and build it. 

As described above it
 - a single table with columns Name and Value
 - showing all EMS device entities that are tagged as favourite
 - showing all Sensors (I'm assuming these are important)
 - showing all Custom Entities (I'm assuming these are also important)
 - if there is no value, leave the Value blank
 - show an icon on the left to show the Device type (boiler, thermostat, sensor etc)
 - clicking on a row will take you to the /devices, /sensors or /customentities page. But not focus on the entity.
 - table is sortable
 - perhaps add a Search like we do with Customizations
 - perhaps add an export to CSV (nice to have)

proddy:
<img width="685" alt="image" src="https://github.com/user-attachments/assets/fda0460b-48b5-41db-af67-ceaf13daf87e">

proddy:
<img width="497" alt="image" src="https://github.com/user-attachments/assets/8908d490-cab7-4433-83ac-6f3aa47c1c36">

dleding:
cant find an other word for it than just a single "wow!"

thank you two! (and maybe others?)

proddy:
This is all you. Thanks for the great idea!

PR #2083, opened 2024-10-10
Ft dashboard - #1958

proddy:
Implements #1958 - The dashboard
plus other fixes, changes.

- data fetches when dialog boxes open
- use useMemo to reduce re-render with page titles
- Analog Sensor - show error in WebUI when trying to using an invalid GPIO
- Replace CTRL-C with CTRL-D, as CTRL-C stops PIO Monitor
- Analog Sensor: fix polarity for Digital Out


proddy:
@MichaelDvP another big one, with many small changes and the backend implementation of the dashboard. Adding the Scheduler to the dashboard was "hard* but I think it works. Its not easy to review all the changes, I would just glance through it and see it it works, and we can fix things as we go along

mattreim:
Update German translation:
```lua
  DASHBOARD: 'Dashboard',
  NO_DATA: 'Keine Daten verfügbar',
  DASHBOARD_1: 'Passen Sie Ihr Dashboard an, indem Sie EMS-Entitäten mithilfe des Moduls „Anpassungen“ als Favorit markieren.'
```

proddy:
Thanks Michael. Those are great additions. I'll like the idea of adjusting write_device_value(), its a lot cleaner.

I could make the changes, or I just create a new branch called ft_dashboard and you put in a PR against that since you have the code working now. What do you prefer?

MichaelDvP:
No preferences, I have for testing a local ft_dashboard branch from your GH, i think i can also do a PR to your proddy/EMS-ESP32/ft_dashboard if you like.

proddy:
Ok, let’s do that.

proddy:
@MichaelDvP ok to merge?

MichaelDvP:
Yes.

MichaelDvP:
Because the devices page has headline with black background i've tested a black box for dashboard: What do you think?
grafik

proddy:
I like the Blackground. I can add it

Do you like the highlight being solid blue instead of the border? I personally do.

Also, notice the slightly different spacing of the devices between the Dashboard and Devices pages. I compacted the Dashboard to show all the data in as little white space as possible. What do you think?

image

MichaelDvP:
For the black background i've used left padding 2 to allign with table ( padding 1 and frame 1):
```
        <Box
          sx={{
            backgroundColor: 'black',
            pt: 1,
            pl: 2
          }}
        >
```

The spacing is imho good. Dashboard will mostly have more lines. 
Also coloring is imho ok.

mattreim:
Is there a way to remove the scheduler entries in the dashboard?

mattreim:
And these entries cannot be opened and closed reliably.

Dashboard

proddy:
> Is there a way to remove the scheduler entries in the dashboard?

if you don't give the schedule entry a name, it will be excluded. 

Is there a reason you would like to see this?

mattreim:
My current scheduler entries are only for creating the remote thermostat and are therefore quite uninteresting for the dashboard.

proddy:
We could add an option to the Scheduler like 'include in Dashboard' @MichaelDvP ?

or @mattreim do all your Scheduler entries need a name? Do you call them via the API?

mattreim:
The entries don't actually need a name and I don't want to call them either.

mattreim:
Is that normal?

Dashboard

proddy:
no, was it always like that or did you make a change to the Scheduler entries?

proddy:
> The spacing is imho good. Dashboard will mostly have more lines. Also coloring is imho ok.

@MichaelDvP do you you think we should use the solid blue hover-over color in the rest of the app, like in Scheduler, Custom Entities, Configurations.... ?

MichaelDvP:
> Is that normal?

Sometimes, temporary,  while scanning the bus, sadly yes. 
I think this is due to resorting devices if a new device is found on the bus and adding the new device sometimes. Devices are stored as vector and vector is not constant in memory and reallocated on adding/sorting a new device. 
I tested to change device storage to deque and it seems to work (empty lines are gone).

proddy:
we can also adjust the web code to only show the parent tree node (e.g. the device) only if it has children (the data entities). This only happens at startup when EMS-ESP is discovering new devices. I'm not at home so can't test, I've been coding this using dummy data

mattreim:
> no, was it always like that or did you make a change to the Scheduler entries?

I have now removed the name and now it looks like this.

proddy:
Then it's something else causing it, either an EMS device that doesn't have values yet or Custom Entities or Analog or Temp Sensors. Can you spot what is missing?

MichaelDvP:
@mattreim 

Please check if it is fixed with https://github.com/MichaelDvP/EMS-ESP32/releases/tag/latest

MichaelDvP:
@proddy 
> We could add an option to the Scheduler like 'include in Dashboard' @MichaelDvP ?

Only for scheduler? Global for all scheduler entities? Seems not right to me. 
Do we need it for sensors and custom entities too?
I think leave is as is or we have to add the flags to all entities (a/t-sensor/custom/scheduler), in consequence with favorite and exclude api/mqtt flags But this is a lot of work for the UI..

MichaelDvP:
BTW: We should change the text here:
grafik
`hide from Dashboard` is wrong now, Rename to: `hide from Devices`? Or remove this flag?
And for `mark as favorite? Should we add `show on Dashboard` Or is it clear that only favorites are shown on Dashboard?

proddy:
> think leave is as is or we have to add the flags to all entities

agree...leave as is.

proddy:
> BTW: We should change the text here: grafik `hide from Dashboard` is wrong now, Rename to: `hide from Devices`? Or remove this flag? And for `mark as favorite? Should we add `show on Dashboard` Or is it clear that only favorites are shown on Dashboard?

good catch. I say leave the word Favorite as its also integrated with the Star * in Devices
But 'hide from Dashboard' is old.  I think we added it because there was a limit of the number of entries you could update on the Customization page. Perhaps we can link 'remove from dashboard' with 'remove from memory'. I'll create a new issue

mattreim:
> @mattreim
> 
> Please check if it is fixed with https://github.com/MichaelDvP/EMS-ESP32/releases/tag/latest

This is what it looks like after 30 minutes of running time.

Dashboard


And I only have the two scheduler entries and nothing else.

proddy:
> > @mattreim
> > Please check if it is fixed with https://github.com/MichaelDvP/EMS-ESP32/releases/tag/latest
> 
> This is what it looks like after 30 minutes of running time.
> 
> Dashboard
> 
> And I only have the two scheduler entries and nothing else.

and if you have time check with the latest .45?

mattreim:
Yes, of course

mattreim:
Now I've waited 10 minutes and it still looks exactly the same.

proddy:
ok, thanks, I'll do some more testing. if I can't find it I'll send you a debug version which will output what the dashboard is using.

MichaelDvP:
Oh, this is scheduler with no named entities, this is a typo, i'll fix in my dev.

proddy:
thanks Michael - I fixed that yesterday but it was still stuck in my local branch

PR #2091, opened 2024-10-14
S3-Temperature, Thermostat dhw circuits

MichaelDvP:
- next fix for digital_out active setting
- add pvMaxCompressor
- check active ems-bus-device (don't fetch from device if not listed in 07 telegram) - don't know if it helps, there was no change for the unstable bus in #1953 maybe revert this and only add to 3.7.1dev?


proddy:
the check active is a nice touch, lets keep it in 3.7.0

PR #2094, opened 2024-10-14
fix empty scheduler on dashboard

MichaelDvP:


src/web/WebDataService.cpp
```diff
@@ -432,7 +432,7 @@ void WebDataService::dashboard_data(AsyncWebServerRequest * request) {
     }
 
     // show scheduler, with name, on/off
-    if (EMSESP::webSchedulerService.count_entities()) {
+    if (EMSESP::webSchedulerService.count_entities(true)) {
```
proddy, line 435: that's it, thought I had checked that fix in yesterday

mattreim:
Now it's fixed, thank you both

Issue #2093, opened 2024-10-14
hide from Dashboard vs remove from memory


proddy:
![grafik](https://github.com/user-attachments/assets/51312f83-cf0b-413c-accf-b357e581c5ff)

`hide from Dashboard` is wrong now, since the introduction of the real Dashboard.  Rename to: `hide from Devices`? Or remove this flag and combine it with "remove from memory"

_Originally posted by @MichaelDvP in https://github.com/emsesp/EMS-ESP32/issues/2083#issuecomment-2411654757_

mattreim:
I'm in favor of renaming `hide from devices` because I use `remove from memory` for the entries without values. When combining, I would always have to include the unimportant ones in the device list.

mattreim:
https://github.com/mattreim/EMS-ESP32/blob/1b70b55989a94dac1d46d73894feb46e0a3f9778/interface/src/i18n/en/index.ts#L159

mattreim:
https://github.com/emsesp/EMS-ESP32/blob/0f877b67de09dae5b0d20e07c827eb9f5a8499f4/interface/src/i18n/en/index.ts#L150

proddy:
I think we go with the rename and re-think Remove from Devices/Remove Memory later. It needs some careful thinking.

PR #2097, opened 2024-10-14
Rename from "CUSTOMIZATIONS_HELP_5" and update slovak translation

mattreim:
If you choose to rename (#2093).


proddy:
wow, that saved me some time, thanks!

PR #2113, opened 2024-10-20
 Punctuation corrected and formatting

mattreim:


proddy:
It's probably easier to add the full-stop to the Typescript code one time, rather than adding it 11 times in all the translation files? Less Flash memory is used

mattreim:
I changed it again.

mattreim:
Should I close it or change something?

mattreim:
Ok

Issue #2116, opened 2024-10-20
Call read commands from Web


proddy:
This has been discussed before and not implemented in favour of the `send` command, but I still think it's useful for novice users.

I've added a new setting in Application Settings called `Developer Mode` which we can use to expose more advanced/debug/developer type features. When enabled you'll get a new command icon in the Web Log where you can send write commands like:

![image](https://github.com/user-attachments/assets/59ad591d-9eed-49ac-aa21-2bca7ec87ff9)

Delta1977:
great feature.
i have another Idea:
It would be great if we could do this also via api or mqtt or is this also implemented?

proddy:
> great feature. i have another Idea: It would be great if we could do this also via api or mqtt or is this also implemented?

Yes, possible now. It's a new system command called `read` so you can call it via API and MQTT.  Like

```sh
curl -X POST \
    -H "Authorization: Bearer ${emsesp_token}" \
    -H "Content-Type: application/json" \
    -d '{"value":"8 2"}' \
    ${emsesp_url}/api/system/read
```

Delta1977:
Very cool. Thank you

Delta1977:
@proddy  i tested the MQTT Variant and the Output is logged in EMS-ESP Log, but no return via mqtt.
is a Mqtt-feedback planned? 
Maybe we can publish the complete Logs via MQTT in an logging Topic.

proddy:
> @proddy i tested the MQTT Variant and the Output is logged in EMS-ESP Log, but no return via mqtt. is a Mqtt-feedback planned? Maybe we can publish the complete Logs via MQTT in an logging Topic.

<IP>/settings/version

proddy:
@Delta1977  log output isn't sent over MQTT, it wouldn't be very useful as each log line would be replaced every second and unless you're quick enough to catch it you'll miss half the logs. We use the SysLog protocol for that as it's designed for that. Alternatively, you could tap into the SSE event source, which is over HTTP.  

The idea of this feature is for novice users to type in a simple read command, then watch the single line output, copy & paste into GitHub for the developers to work on.

VlastiBroucek:
I am using 3.7.1-dev.2 and it is not working. Tried it in Edge, Chrome and Firefox on both Windows and in KALI Linux

proddy:
> I am using 3.7.1-dev.2 and it is not working. Tried it in Edge, Chrome and Firefox on both Windows and in KALI Linux

[...]

PR #2117, opened 2024-10-20
Developer Mode and send EMS Read Commands from WebUI - #2116

proddy:
- Developer Mode and send EMS Read Commands from WebUI
- Move Version check from Status to Settings and add a dedicated page
- Banded rows in WebUI dashboard
- Updated Unity tests
- Remove 'All Values' button from WebUI (for now)


mattreim:
Update German translation:

```lua
DEVELOPER_MODE: 'Entwicklermodus'
```

MichaelDvP:
What should the developer mode do? I can only see a checkbox in Web and bools in system, but nothing that checks this flag?

proddy:
> What should the developer mode do? I can only see a checkbox in Web and bools in system, but nothing that checks this flag?

oops, forgot to save. When setting developer mode it'll show a button on the Web Log page where you can send a read command

MichaelDvP:
Wow, this build is 200k bigger than before, i guess the stringstream class?

proddy:
oh, that is not good. I saw stream was used in the Modbus code, so I assumed it was already loaded somewhere. I'll re-write that function.

src/system.cpp
```diff
@@ -2012,4 +2015,63 @@ bool System::uploadFirmwareURL(const char * url) {
...
+    // extract <deviceID> <type ID> [offset] [length] from string
+    char * p;
+    char   value[10] = {0}; // null just in case
+
+    // make a copy so we can iterate
+    char data_args[EMS_MAX_TELEGRAM_LENGTH];
+    for (uint8_t i = 0; i < strlen(data); i++) {
+        data_args[i] = data[i];
+    }
+    data_args[strlen(data)] = '\0'; // make sure its terminated
```
MichaelDvP, line 2029: EMS_MAX_TELEGRAM_LENGTH seems not the logical., it's not a telegram.
the for.. copy can overflow if data is longer, why not using strlcpy?

proddy: that is better indeed

PR #2121, opened 2024-10-22
Update German translation

mattreim:


proddy:
thanks, I had already changed this in the next PR so will merge your changes now

Issue #2115, opened 2024-10-20
feature: extending Scheduler with functions


proddy:
### Discussed in https://github.com/emsesp/EMS-ESP32/discussions/2111

<div type='discussions-op-text'>

<sup>Originally posted by **SoftVienka** October 19, 2024</sup>
Hello
Is it possible to use the round() function in the scheduler?

Write entity RAM float (17.25) to entity RAM as int (17)

example:
Command: custom/interval
Value: (x > 2 && x < 10) ? custom/interval_1 : custom/interval

interval_1 is a float and I want it to be stored as an int into interval

when I write round(custom/interval1) then npr: round17.25 is written into the interval</div>

proddy:
Michael added the functions int, round, abs, log, exp, sqrt and pow in 3.7.0-dev.48

Issue #2109, opened 2024-10-18
Wrong Uptime compressor pool


Sbried:
### DESCRIPTION

Pool Compressor uptime is wrong. But this is not a problem of EMS-ESP.
I saw some problems in the issues/discussion around the interpretation of 4 byte values.

The information sent from the boiler is:
`002+12:18:17.249 N 15: [emsesp] boiler(0x08) -W-> me(0x0B), UBAInformation(0x0495), data: 00 0E 59 CA 00 01 0A D6 00 0D 4E F4 FF FF FF 00 00 00 00 00 0`
`FF FF FF 00`  calculates to those ~49000 days, see screenshot.
Assuming it´s a signed 4 byte integer, it will calculate to -256 (two´s complement). This is also nonsens as an uptime can´t be negative in my opinion.
Maybe a solution is to show other than 0 only in case if the value of the signed 4 byte integer is positive?
Or if compressor pool starts is >0? Or if a PMxx is installed?
Or other ideas?

### REQUESTED INFORMATION

_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [X] Searched the issue in [issues](https://github.com/emsesp/EMS-ESP32/issues)
- [X] Searched the issue in [discussions](https://github.com/emsesp/EMS-ESP32/discussions)
- [ ] Searched the issue in the [docs](https://emsesp.org/Troubleshooting/)
- [ ] Searched the issue in the [chat](https://discord.gg/3J3GgnzpyT)
- [X] Provide the System information in the area below, taken from `http://<IP>/api/system`

```json
{
    "system": {
        "version": "3.7.0-dev.43",
        "uptime": "002+12:22:28.559",
        "uptimeSec": 217348,
        "platform": "ESP32",
        "cpuType": "ESP32-D0WD-V3",
        "arduino": "ESP32 Arduino v2.0.17",
        "sdk": "v4.4.7-dirty",
        "freeMem": 157,
        "maxAlloc": 107,
        "freeCaps": 4174,
        "usedApp": 2333,
        "freeApp": 3363,
        "partition": "app0",
        "resetReason": "Software reset CPU / Software reset CPU",
        "psram": true,
        "psramSize": 4093,
        "freePsram": 4051,
        "model": "BBQKees Gateway Model EMS Gateway E32 V2 v2.0/2024071"
    },
    "network": {
        "network": "Ethernet",
        "hostname": "ems-esp",
        "TxPowerSetting": 0,
[...]

Issue #2073, opened 2024-10-06
improve EMS brand detection


proddy:
EMS-ESP tries to determine the manufacturer/brand (Buderus, Sieger, Junkers, Bosch, Nefit etc..) from each connected EMS device by reading the last byte from the Version 0x02 telegram (in `emsesp.cpp::process_version()`).

But it does a poor job and only detects a few brands correctly. Let's improve this, ask the community for make/model and the results of a `read <deviceID> 2` to correctly build the mapping.

MichaelDvP:
Good idea, i think the 2nd and 3rd identifier WILL gives some more info, also for the thermostats with same 1st ID. I'll start with my system:
Type Brand Name DeviceID data
```
Boiler		Buderus	GB125	(0x08): data: 48 03 00 4B 04 1C 00 00 00 00
Controller	Buderus	BC10 	(0x09): data: 44 02 03
Thermostat	Buderus	RC35 	(0x10): data: 56 01 12
Mixer		Buderus	MM10 	(0x21): data: 45 02 01
Solar		Buderus	SM10 	(0x30): data: 49 02 00
```

proddy:
Mine:
```
Boiler       Nefit Trendline (0x08):     data 7B 06 01 00 00 00 00 00 00 04
Controller   BC10 (0x09):                data BE 01 03
Thermostat   Nefit Moduline 300 (0x17):  data 4D 03 03
```
Notes
- the controller is recognised as a BC10 but it's built into the boiler, since its a combi-boiler
- the boiler's brand is identified correctly as Nefit because of the `04` in the last byte

VlastiBroucek:
Mine (I am including more data FYI, since the device 0x08 and 0x09 return same string)...
```
ems-esp:$ show devices
These EMS devices are currently active:

boiler: Condens 2500/Logamax/Logomatic/Cerapur Top/Greenstar/Generic HT3 (DeviceID:0x08, ProductID:95, Version:12.16)
 Received telegram type IDs: 0x10 0x11 0x15 0x1C 0x18 0x19 0x34 0x2A 0x04
 Fetched telegram type IDs: 0x14 0x16 0x33
 Pending telegram type IDs: 0xBF 0xC2 0x1A 0x35 0x26
 Ignored telegram type IDs: 0x23

thermostat: RC200/CW100 (DeviceID:0x18, ProductID:157, Version:43.02)
 Received telegram type IDs: 0x06 0xA2 0x31D
[...]

PR #2129, opened 2024-10-22
scheduler functions #2115, uptime notset-value #2109, device custom name #2073

MichaelDvP:
my devcies does not show telegram 01 offset 27. This build reads the name after registering 02 version and sets the custom name if this is not set by user, so it can always be overwritten by user. But not tested.

The change in telegram sets for telegram-reads the correct  length. This was a  test for the faulty mix communication but does not help. But it's more clean.


proddy:
Great to see those scheduler functions. Would you mind adding a line to https://emsesp.org/Configuring/#conditions so we record it. Also mention it in the ChangeLog.

For the name (offset 27), let's see what people think. It may be a little strange for some e.g. a Logatherm WLW heat pump gets renamed to "XCU_THH".

mattreim:
@proddy Is the sonar message (`Make sure use of "strlen" is safe here.`) really a problem?

MichaelDvP:
It's a warning and strlen is safe in this case: the name is generated with strlcpy, so it's always null terminated within size limit.

proddy:
yes, it's ok. I've updated Sonar

mattreim:
Thanks for the information

PR #2154, opened 2024-10-29
updates to 3.7.1

proddy:
minor updates to comments and doc.

I created a data folder. So when you add a file called `pre_load.json` , doing a `pio run -e s3_16M_P -t uploadfs` (example) will pre-install some user settings.


src/telegram.h
```diff
@@ -133,7 +133,7 @@ class Telegram {
         auto msg_size = (index - this->offset + num_bytes - 1);
 
         if ((index < this->offset) || (msg_size >= this->message_length) || (msg_size > EMS_MAX_TELEGRAM_MESSAGE_LENGTH)) {
-            return false;
+            return false; // TODO we should set value to a default value here
```
MichaelDvP, line 136: No!, we often get only parts of a telegram. If a value is not in this part (offset/length) we don't want to change it.

proddy: okay!

PR #2158, opened 2024-10-30
small fixes, webinterface and log messages

MichaelDvP:
see commits


proddy:
thanks, nice updates. To think, 3 years ago none of us knew react. And now we're pros!

PR #2183, opened 2024-11-06
fix standalone

MichaelDvP:


proddy:
already done in my PR, so we'll go with that one

Issue #2177, opened 2024-11-05
Support MQTT Discovery (AD) with Domoticz


proddy:
There's been a few comments in various issues around correct support for Domoticz, so thought best to create it's own GitHub Issue for tracking progress.

@bbqkees pointed me to https://github.com/domoticz/domoticz/blob/master/hardware/MQTTAutoDiscover.cpp so now we can see what is and what isn't supported.

This is related to https://github.com/emsesp/EMS-ESP32/discussions/2138 and https://github.com/emsesp/EMS-ESP32/discussions/2156

Domoberry:
Just got an (immediate :-) )  reply on the same from Domoticz in this post [https://www.domoticz.com/forum/viewtopic.php?p=321086#p321086](url). It refers to the development branch of the same: [https://github.com/domoticz/domoticz/blob/development/hardware/MQTTAutoDiscover.cpp](url).

proddy:
Thanks. I'll take a look. Interesting that this file is 5200 lines without a single comment and there isn't any documentation either. I guess most people have moved over to Home Assistant

proddy:
Looks like Domoticz does support select, switch and climate so we can add that back in. Except for 'text'.

And we need to make sure every entity has a uom ("unit_of_meas") set

proddy:
> Remember [#1360 (comment)](https://github.com/emsesp/EMS-ESP32/issues/1360#issuecomment-1876689332) Quote:
> 
> > > However, I would leave the Domoticz choice in there and mark it as specific for pre 2024.1 Domoticz versions. Reason is that the 2024.1 Domoticz lost support for ArmV6 Raspberry's, so a lot off people will be on the older Domoticz for some time.
> 
> I think this is still valid.

yes, we just replace 'text' and also fix the analog & temp sensors (I made TODOs). I'll do the final changes after your PR is merged.

proddy:
made some changes in dev-4 that need testing (when the PR is merged). I'm not sure how to handle the uom for sensors without a uom (like a number). It's not clear from the Domoitcz c++ code how they handle this.

Domoberry:
with the risk of posting 'old news'... 
Had V3.7.1.-test.3 running for a while. These are some notes/suggestions:
[...]

PR #2185, opened 2024-11-06
mainly MQTT fixes for Domoticz

proddy:


src/devices/thermostat.cpp
```diff
@@ -4912,7 +4912,7 @@ void Thermostat::register_device_values_dhw(std::shared_ptr<Thermostat::DhwCircu
                               DeviceValueUOM::MINUTES,
                               MAKE_CF_CB(set_wwchargeduration));
         register_device_value(tag, &dhw->wwCharge_, DeviceValueType::BOOL, FL_(wwCharge), DeviceValueUOM::NONE, MAKE_CF_CB(set_wwcharge));
-        register_device_value(tag, &dhw->wwExtra_, DeviceValueType::UINT8, FL_(wwExtra), DeviceValueUOM::DEGREES);
+        register_device_value(tag, &dhw->wwExtra_, DeviceValueType::UINT8, FL_(wwExtra), DeviceValueUOM::NONE);
```
MichaelDvP, line 4915: No complain in all these years ;-) 
But the comment in https://github.com/emsesp/EMS-ESP32/blob/9689f0a7da258d5cc27cb9e8a588f98a736a3141/src/devices/thermostat.cpp#L1249 says it is bool.

MichaelDvP (approved):
Looks all good to me.

src/devices/thermostat.cpp
proddy: that's what I thought, but if you scroll up you see it also can have a value of 2. So?? not sure

proddy:
made some more edits. While we wait for the Domoticz to be tested - https://github.com/emsesp/EMS-ESP32/issues/2177

src/devices/thermostat.cpp
MichaelDvP: Did not find this. The comment above describes the telegram offsets.
I think it's bool.

proddy: i think you're right. will change

PR #2202, opened 2024-11-14
 fix analog input pullup and counter on gpio 25/26, dac values on web #2201 

MichaelDvP:
I think the documentation is still ok. 
One of the pin links seems broken and the valid pins only listed for esp32, not for the s3 chip.
S2 have a DAC (S3 not) on pin 23/24, but these pin are gpio 17/18. 
Seems to be much work to display S2 dac on web correctly, skipped it as we have no official S2 support.


proddy:
thnaks Michael for picking this one up so quickly

proddy:
> A small note: the installer image is gone.
> 
> ESP-installer

I'll look into that now

mattreim:
And this translation could be used for the new firmware page.
https://github.com/emsesp/EMS-ESP32/blob/bf210977452979c8e6390cda9e1a6faf903b64e3/interface/src/i18n/en/index.ts#L178-L179

proddy:
fixed.

Issue #2207, opened 2024-11-15
Track number of WiFi reconnects


proddy:
There were reports of v3.7.0 losing connection to the WiFi network, so we added a new system entity to the MQTT heartbeat, counting the number of WiFi reconnects. It's called `WiFi reconnects` (wifireconnects) to keep consistent with the MQTT mqttconnects. The value is also shown in the WebUI. 

![image](https://github.com/user-attachments/assets/aa3f563a-e5ca-470e-8fd2-d1429fd06964)

For HA users if you want to get alerted when EMS-ESP drops from the WiFi network and reconnects, use a trigger like:

```yaml
alias: EMS-ESP Wifi Reconnected
description: Notify when EMS-ESP reconnects
triggers:
  - entity_id:
      - sensor.system_wifireconnects
    trigger: state
conditions: []
actions:
  - data:
      message: EMS-ESP has reconnected!
      title: EMS-ESP
    action: notify.mobile_app_my_iphone
mode: single
```

in 3.7.1-dev.8

proddy:
dev-8

PR #2208, opened 2024-11-15
Wifi Reconnect #2207

proddy:
- #2207 - show Wifi reconnect in MQTT heartbeat + HA
- Don't go to Admin/SU mode when connected to a serial/USB
- Remove NTP Status from heartbeat


MichaelDvP (approved):
One thing: You show `wificonnects: 0` in heartbeat. It should be named `reconnects` or count the first connect too. The `mqttconnects` starts with first connect 1.
So in my heartbeat it shows mqttconnects=1 and wificonnects=0.

proddy:
> One thing: You show `wificonnects: 0` in heartbeat. It should be named `reconnects` or count the first connect too. The `mqttconnects` starts with first connect 1. So in my heartbeat it shows mqttconnects=1 and wificonnects=0.

yes, you're right. I'll update.

PR #2222, opened 2024-11-19
fix #2216, add #2212, #2217

### LINKED ISSUE #2216: Free Memory decreased by HTTP API requests
### DESCRIPTION

With frequent HTTP API requests every minute (for example with "/api?device=boiler&cmd=info") , the free memory continuously decreases until the module freezes after 45 minutes.
No other services active (MODBUS off; MQTT off, etc.). This occured with latest 3.7.0. and 3.7.1. developer version with ESP32-S3.

### TO REPRODUCE

frequent HTTP API requests

### EXPECTED BEHAVIOUR

That the free memory not decreases and no freeze

MichaelDvP:
The new solar values needs to be added to modbus. Don't know how this work. There are some scripts  in the script folder, but the commands suggested in the comments do not exist and calling the script in wsl only throw half a million of errors.


proddy:
Merged, but standalone is broken

PR #2235, opened 2024-11-21
modbus fixes

MichaelDvP:
Please generate the modbus files, i think this should gve now the water devices.
Can yo give me advice how to run the generate_csv_and_headers.sh script? I guess something is missing in wsl, but don't know what.

### DIFF
--- a/src/emsesp.cpp
+++ b/src/emsesp.cpp
@@ -343,17 +343,14 @@ void EMSESP::dump_all_entities(uuid::console::Shell & shell) {
         for (const auto & device : device_library_) {
             if (device_class.first == device.device_type) {
                 uint8_t device_id = 0;
-                // Mixer class looks at device_id to determine type and the tag
-                // so fixing to 0x28 which will give all the settings except flowSetTemp
-                if (device.device_type == DeviceType::MIXER) {
+                // Water class looks at device_id to determine type and the tag
+                if (device.device_type == DeviceType::WATER) {
                     if (device.flags == EMSdevice::EMS_DEVICE_FLAG_MMPLUS) {
-                        if (device.product_id == 160) { // MM100
-                            device_id = 0x28;           // dhw
-                        } else {
-                            device_id = 0x20; // hc
-                        }
-                    } else {
-                        device_id = 0x20; // should cover all the other device types
+                        device_id = 0x28; // dhw 1/2
+                    } else if (device.flags == EMSdevice::EMS_DEVICE_FLAG_SM100) {
+                        device_id = 0x2A; // dhw 3
+                    } else if (device.flags == EMSdevice::EMS_DEVICE_FLAG_IPM) {
+                        device_id = 0x40; // dhw 1
                     }
                 }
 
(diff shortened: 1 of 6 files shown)


proddy:
I'll update the doc. I did this this morning before seeing your PR so will run it again.

To generate the files I just run `sh ./scripts/generate_csv_and_headers.sh` from the root folder. You will need python installed (type `python3`) and also GNUMake. Instructions are in https://docs.emsesp.org/Building/#wsl-building-a-linux-environment-on-windows

proddy:
there's a problem with mixer tags getting a value of -31 which breaks the modbus gen scripts, I'm looking into that now

src/emsesp.cpp
```diff
@@ -343,17 +343,14 @@ void EMSESP::dump_all_entities(uuid::console::Shell & shell) {
         for (const auto & device : device_library_) {
             if (device_class.first == device.device_type) {
                 uint8_t device_id = 0;
-                // Mixer class looks at device_id to determine type and the tag
-                // so fixing to 0x28 which will give all the settings except flowSetTemp
-                if (device.device_type == DeviceType::MIXER) {
```
proddy, line 348: the entities dump now has a dv.tag value of -31 (not sure why yet), so it breaks the modbus gen files. Any idea how to fix this @MichaelDvP ?

MichaelDvP:
Oh, we needvto set device-id for Mixer to 0x20 to have hc1.

Am 21. November 2024 10:25:57 MEZ schrieb Proddy ***@***.***>:

proddy:
> Oh, we needvto set device-id for Mixer to 0x20 to have hc1. Am 21. November 2024 10:25:57 MEZ schrieb Proddy ***@***.***>:
> […](#)
> @proddy commented on this pull request. > @@ -343,17 +343,14 @@ void EMSESP::dump_all_entities(uuid::console::Shell & shell) { for (const auto & device : device_library_) { if (device_class.first == device.device_type) { uint8_t device_id = 0; - // Mixer class looks at device_id to determine type and the tag - // so fixing to 0x28 which will give all the settings except flowSetTemp - if (device.device_type == DeviceType::MIXER) { the entities dump now has a dv.tag value of -31 (not sure why yet), so it breaks the modbus gen files. Any idea how to fix this @MichaelDvP ? -- Reply to this email directly or view it on GitHub: [#2235 (review)](https://github.com/emsesp/EMS-ESP32/pull/2235#pullrequestreview-2450739924) You are receiving this because you were mentioned. Message ID: ***@***.***>

I'll add it

proddy:
done

MichaelDvP:
> I'll update the doc. 

Hmm, the scripts mentioned there are not in the dev branch, i took them from main. 
But is seems i have an issue with native build in wsl, Something missing? This is the output:
[error.txt](https://github.com/user-attachments/files/17867504/error.txt)

Also trying wsl native build deletes the build/firmware folder, all previous builds are lost :-(

>> need to set device-id for Mixer to 0x20 to have hc1.

> done

We also need to add the same to telegram-dump, the mixer telegram-type-ids are calculated with device-id.

proddy:
which scripts are not in the dev branch?

the compile error, I remember seeing this before. its to do with versions of gcc and the toolchain. I'll need to refresh my memory. Sometimes always good to remove the .pio folder and start fresh. Maybe something left over from pioarduino?

MichaelDvP:
> which scripts are not in the dev branch?

`scripts/update_modbus_registers.sh`
`scripts/generate-modbus-register-doc.sh`

proddy:
those scripts have been removed. I'll update the comments where they are still mentioned.

proddy:
guess I'll need to fix that as well then

For reference, a good guide on how to build (at least the steps I do) are in https://github.com/emsesp/.github-private/blob/main/profile/README.md

Issue #2245, opened 2024-11-23
Unable to enable Modbus 3.7.1-dev.10


Zembol76:
### DESCRIPTION

After uploading 3.7.1-dev.10 such errors 
2024-11-23 19:56:06.000 ERROR 7: [modbus] Error in modbus parameters: temp must be listed before requiredtemp.
2024-11-23 19:56:06.000 ERROR 8: [modbus] Unable to enable Modbus - the parameter list order is corrupt. This is a firmware bug.
In 3.7.1-dev.9 everything is ok

MichaelDvP:
I've fixed in https://github.com/MichaelDvP/EMS-ESP32/releases/tag/latest and made a PR.

We have added the water devices in dev10. SM100 can only act as DHW3 circuit, but modbus script generation needs all mapped to DHW1, so it has generated doulettes.

Zembol76:
Ok works, tested with plc s7 1200 Siemens.

Issue #2220, opened 2024-11-18
HM200 0x53 missing entities, HP heating starts /HP dhw starts  (ems-esp 3.7.0)


TobiRabid:
The HM200 provides (like the boiler _heatstarts_ and _dhw/starts_ ) statistic values for the heatpump starts (heating/dhw).
Attached a screenshoot from mybuderus app with this values (3626 for heating starts, 184 for dhw starts)

I tried to provoke to increase the counter during the watch command via sg-ready stop ( the hp stoped,but not sure that the values just in time changed. The app shows the old values, but there is a cloud backend between...)

Atached also the watch file

It would be very nice to have this entities in one of the next versions

regards

![IMG_2621](https://github.com/user-attachments/assets/44584f4c-5213-4d9f-8535-fcd0d4c9296a)
[ems-esp watch on 0x53-hp-stop-sg-ready.txt](https://github.com/user-attachments/files/17804182/ems-esp.watch.on.0x53-hp-stop-sg-ready.txt)

atached a second very long  watch file.. i released the blocking of the hp via sgready signal, the boiler stops immediately and the hp starts after 11 minutes.... the counter for hp heating start increased +1 to 3627 in the app

[hp-start-after-blocking.txt](https://github.com/user-attachments/files/17805463/hp-start-after-blocking.txt)

proddy:
@TobiRabid can you check the latest 3.7.1 dev release?

TobiRabid:
The starts counter values are visible, but the discribtion is swapped

<img width="1603" alt="2024-11-22_10h02_17" src="https://github.com/user-attachments/assets/3f91c786-c8b5-44e1-b910-c2a457be2fbf">

TobiRabid:
thx

PR #2246, opened 2024-11-24
fix modbus #2245 and reorder #2220

MichaelDvP:
I have only removed the SM100 water from modbus. Has to be generated again, but still can't compile in wsl, mybe my laptop has too low resources.


proddy:
I can do the generation thing. Maybe it should be part of the CI scripts to automatically rebuild the files and check-them in.

Or if you can get pio native to work on Windows (not WSL) we can adjust the .sh shell script to work with PowerShell.

WSL2 is a monster - it uses up to 7GB memory on my desktop PC so I can imagine it's awful on a windows laptop. I know BBQKees had a lot of issues too running on chrome laptops. In the end it was just easier to install Linux as a separate OS on the laptop and switch between the two at boot

MichaelDvP:
Yes, windows native compile works good, To generate the modbus in powershell and/or github would be fine.

BTW: i renamed one of he ISM modules to ISM2, the csv put both ISM under one headline, looks a bit odd.

Issue #2247, opened 2024-11-24
Last Will (LWT) not set on MQTT Connect


andatche:
### DESCRIPTION

I recently upgraded from 3.6.2 to 3.7.0 and since, EMS-ESP is not setting the Last Will properties on the MQTT Connect command. As a result, Home Assistant continues to perceive it as online after disconnecting/crashing etc.

I tested downgrading to 3.6.5 and that appears to be affected too.

Example PCAP from a connection attempt:
```
Frame 9: 118 bytes on wire (944 bits), 118 bytes captured (944 bits)
Ethernet II, Src: <snip> (<snip>), Dst: <snip> (<snip>)
Internet Protocol Version 4, Src: 192.168.5.2, Dst: 192.168.1.10
Transmission Control Protocol, Src Port: 57354, Dst Port: 1883, Seq: 1, Ack: 1, Len: 64
MQ Telemetry Transport Protocol, Connect Command
    Header Flags: 0x10, Message Type: Connect Command
        0001 .... = Message Type: Connect Command (1)
        .... 0000 = Reserved: 0
    Msg Len: 62
    Protocol Name Length: 4
    Protocol Name: MQTT
    Version: MQTT v3.1.1 (4)
    Connect Flags: 0xc2, User Name Flag, Password Flag, QoS Level: At most once delivery (Fire and Forget), Clean Session Flag
        1... .... = User Name Flag: Set
        .1.. .... = Password Flag: Set
        ..0. .... = Will Retain: Not set
        ...0 0... = QoS Level: At most once delivery (Fire and Forget) (0)
        .... .0.. = Will Flag: Not set
        .... ..1. = Clean Session Flag: Set
        .... ...0 = (Reserved): Not set
    Keep Alive: 60
    Client ID Length: 7
    Client ID: ems-esp
    User Name Length: 7
    User Name: ems-esp
    Password Length: 32
    Password: <snip>
```

### REQUESTED INFORMATION

_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [x] Searched the issue in [issues](https://github.com/emsesp/EMS-ESP32/issues)
- [x] Searched the issue in [discussions](https://github.com/emsesp/EMS-ESP32/discussions)
- [x] Searched the issue in the [docs](https://docs.emsesp.org/Troubleshooting/)
- [x] Searched the issue in the [chat](https://discord.gg/3J3GgnzpyT)
[...]

PR #2248, opened 2024-11-24
fix Last Will (LWT) not set on MQTT Connect #2247

proddy:


lib/framework/MqttSettingsService.cpp
```diff
@@ -207,6 +207,17 @@ bool MqttSettingsService::configureMqtt() {
...
+        // create last will topic with the base prefixed. It has to be static because the client destroys the reference
+        static char will_topic[FACTORY_MQTT_MAX_TOPIC_LENGTH];
+        if (_state.base.isEmpty()) {
+            snprintf(will_topic, sizeof(will_topic), "status");
+        } else {
+            snprintf(will_topic, sizeof(will_topic), "%s/status", _state.base.c_str());
+        }
+        setWill(will_topic);
+
+
```
MichaelDvP, line 220: This only effects the unsecure client. 
Better remove `setWill()`function, set the will_topic before line 187 (#ifndef TASMOTA) and use 
`static_cast<espMqttClientSecure *>(_mqttClient)->setWill(will_topic, 1, true, "offline");`
and
`static_cast<espMqttClient *>(_mqttClient)->setWill(will_topic, 1, true, "offline");`
in the secure/unsecure sections (before line 200/210)

MichaelDvP (changes requested):
I think this is the right way and works on the unsecure client, setting for secureclient  is missing.

src/mqtt.cpp
```diff
@@ -524,6 +514,16 @@ void Mqtt::on_connect() {
...
+    // create last will topic with the base prefixed. It has to be static because the client destroys the reference
+    static char will_topic[MQTT_TOPIC_MAX_SIZE];
+    if (!Mqtt::base().empty()) {
+        snprintf(will_topic, MQTT_TOPIC_MAX_SIZE, "%s/status", Mqtt::base().c_str());
+    } else {
+        snprintf(will_topic, MQTT_TOPIC_MAX_SIZE, "status");
+    }
+    // EMSESP::esp8266React.setWill(will_topic); // with qos 1, retain true
+
+
```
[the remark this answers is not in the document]
proddy: oops, it was left in. I was testing whether you had to set the will before or after the connect

lib/framework/MqttSettingsService.cpp
proddy: setWill does before secure and unsecure, but indeed this approach is cleaner. I'll update

PR #2251, opened 2024-11-25
show download button instead of dialog

proddy:


MichaelDvP:
Now the download button points is placed on dev entry, but points to 3.7.0 release version.
grafik

proddy:
thanks, I'll fix now

PR #2280, opened 2024-12-05
small fixes and changes

MichaelDvP:
see commits and changelog,it addresses some issues and discussions
still not all is confirmed by the users


interface/package.json
```diff
@@ -24,18 +24,18 @@
...
     "alova": "3.2.6",
     "async-validator": "^4.2.5",
     "jwt-decode": "^4.0.0",
     "mime-types": "^2.1.35",
     "preact": "^10.25.1",
-    "react": "^18.3.1",
-    "react-dom": "^18.3.1",
-    "react-icons": "^5.3.0",
-    "react-router": "^7.0.1",
+    "react": "^19.0.0",
```
proddy, line 35: react 19 is a big change, we'll need to keep an eye out that nothing is broken downstream

MichaelDvP: Yes, i quickly navigated all pages and it looks ok. 
Only thing is recognized, but this was also before, the router tabs in security and network are not highlighted/underlined as in older versions.
Acccording to upgrade guide https://react.dev/blog/2024/04/25/react-19-upgrade-guide most breaking changes should also give a warning.
If i see something strange,i will also test with older react versions.

PR #2311, opened 2024-12-29
Updated unknown compressor stati "enum_hpactivity"

Sbried:
Updated the compressor status
- 0 from "none" to "off". 0 is defined as "off"
- 5 from "unknown" to "pool_heating". 5 is defined as pool_heating, but I have no idea what is the difference between 4 (pool) and 5 (pool_heating).
- 7 is added as "compressor_alarm".

I did a local build with the environment S_16M_P which runs without problems.

Any hints regarding this pull request? It´s my first at all ;-)


proddy:
looks ok to me, Michael?

MichaelDvP (approved):

PR #2318, opened 2025-01-01
Core update 2108

proddy:
Looks like a lot of changes, but its really just minor refactor

- move src files into a new folder called 'core' 
- move ESP8266React from lib into its own source folder under src, as it's not a library anymore. Renamed to ESP32React
- removes own versions of asynctcp. asyncwebserver and arduinojson from lib and replace with lib_deps which are automatically pulled from github. Web code updated to support the standard way msgpack is handled in AsyncWS
- support for  ArduinoJson 7.3.0, to be compatible with Modules which has also been updated
- prepare platformio.ini for pioarduno and quickly switching between Core 2.x and 3.1/IDF 5,3 for testing
- optimized the platformio.ini and makefile


platformio.ini
```diff
@@ -120,6 +145,7 @@ board = lolin_s3
 board_build.f_cpu = 240000000L
 board_upload.flash_size = 16MB
 board_build.partitions = esp32_partition_16M.csv
+board_build.extra_flags = -DBOARD_HAS_PSRAM
```
MichaelDvP, line 148: Again this overwrites the extra_flags with USB settings, lolinS3 has psram, no need to set it here.

lib/espMqttClient/src/espMqttClient.cpp
```diff
@@ -78,37 +78,27 @@ espMqttClientSecure::espMqttClientSecure(uint8_t priority, uint8_t core)
 }
 
 espMqttClientSecure & espMqttClientSecure::setInsecure() {
-#if defined(EMC_CLIENT_SECURE)
```
MichaelDvP: dows not compile with tasmota SDK if this is removed

platformio.ini
proddy: this target just builds the binary, none of the USB settings are needed. eventually I'm going to remove the 'lolin_s3' and replace it with a generic 'esp32' for all boards.

lib/espMqttClient/src/espMqttClient.cpp
proddy: already fixed that this morning

platformio.ini
MichaelDvP: We had this before, the serial console on S3 is not working if the USB extra flags are not set.  This board profile already has PSRAM set, so this entry is bad, it does not affect PSRAM, but clears all other flags.

MichaelDvP (commented):
The flash size is increased a lot, 4M partition does not fit anymore.
```
RAM:   [==        ]  15.1% (used 49412 bytes from 327680 bytes)
*** [checkprogsize] Explicit exit, status 1
Flash: [==========]  102.9% (used 2090133 bytes from 2031616 bytes)
```
Don't know what is the cause, i guess the Cpp17.

platformio.ini
proddy: yes I remember, but I can't reproduce when testing on two different S3s. Wait until I've modified the board's and try again.

MichaelDvP: I remember you have a S3 with 2 USB, one native and one with usb/serial chip. The native does not work with serial console if you modify the extra-flags.
Settings in the board file for lolin-s3:
```
    "extra_flags": [
      "-DARDUINO_LOLIN_S3",
      "-DBOARD_HAS_PSRAM",
      "-DARDUINO_USB_MODE=1",
      "-DARDUINO_USB_CDC_ON_BOOT=1",
      "-DARDUINO_RUNNING_CORE=1",
      "-DARDUINO_EVENT_RUNNING_CORE=1"
    ],
```

proddy:
> The flash size is increased a lot, 4M partition does not fit anymore.
> 
> ```
> RAM:   [==        ]  15.1% (used 49412 bytes from 327680 bytes)
> *** [checkprogsize] Explicit exit, status 1
> Flash: [==========]  102.9% (used 2090133 bytes from 2031616 bytes)
> ```
> 
> Don't know what is the cause, i guess the Cpp17.

it was the `-fexceptions` flag. removed it.

proddy:
closing. replaced with https://github.com/emsesp/EMS-ESP32/pull/2328

PR #2331, opened 2025-01-05
web UI small changes

proddy:
- bigger icons
- use customization icon in device detail
- use same tooltips names as the functions are called
- fix the formatting of the Version page
- add the option to re-install latest version 
-

### DIFF
--- a/interface/src/app/main/Dashboard.tsx
+++ b/interface/src/app/main/Dashboard.tsx
@@ -274,7 +274,7 @@ const Dashboard = () => {
           <IconContext.Provider
             value={{
               color: 'lightblue',
-              size: '16',
+              size: '18',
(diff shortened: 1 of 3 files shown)


mattreim:
It looks somehow strange.

v3 7 2-dev 8


And when I want to install it on my `ESP32-4M` it breaks and the following warning appears in the log.


```lua
2025-01-05 20:42:37.236   ERROR 12:     [system]     Firmware upload failed - HTTP code -1
```

proddy:
yes, i reproduced the same problem with the 4MB bbqkees S32 board. I think the firmware .bin is missing. I'll fix it.

mattreim:
It's not that bad.

Issue #2354, opened 2025-01-15
Buderus AM200 / register not supported


Duff411:
SW: 3.7.2 dev.10
The following problem exists:
I have connected EMS ESP to a Buderus heating system.
The connection to the gateway is established via Modbus TCP.
The following units are installed:
GB172
RC310
MM100
MM100
AM200

I just can't access the AM200 module.
I have unit ID 17 and register block 19000.
The unit ID exists and I have read it out beforehand.
However, I always get the following error message:
“Error: Modbus exception 2: Illegal data address (register not supported by device)”

For your information:
The MM100 modules work with unit-ID 7 and register block 1000 and 2000.

Duff411:
It works 👍

PR #2356, opened 2025-01-16
prepare for modbus support AM200, #2354

MichaelDvP:
Please generate the modbus entity parameters to complete the the fix.

BTW: there are now two files in dev:
- `src/core/modbus_entity_paramters.hpp` (older)
- `src/modbus_entity_parameters.hpp` (newer)

please check.


proddy:
will do.

PR #2360, opened 2025-01-18
implement CPU temp for ESP32

proddy:


mattreim:
CPU

Great, it worked, but unfortunately it is not yet transmitted via MQTT.

EMS CPU Temp

proddy:
oops, will fix that.

PR #2361, opened 2025-01-19
include CPU temp in heartbeat MWTT topic for ESP32

proddy:


mattreim:
Unfortunately it doesn't work yet.

proddy:
fixed. will check in shortly.

PR #2370, opened 2025-01-24
add RGB Led to board profile

MichaelDvP:


proddy:
nice, thanks @MichaelDvP . I never noticed the double log entries on startup, what was shown exactly?

MichaelDvP:
With the new asyncWebServer it shows entries 1-11 and then 1-11 a second time. On second navigate to webLog all is shown ok. I first thought that this is the new queue and set it to zero. But still sometimes it shows double. 
I think there is still a message queue and after startup the messages were sent to queue and on naviagte to weblog page the index is reset and sent a second time. But not sure what happend realy.

proddy:
> With the new asyncWebServer it shows entries 1-11 and then 1-11 a second time. On second navigate to webLog all is shown ok. I first thought that this is the new queue and set it to zero. But still sometimes it shows double. I think there is still a message queue and after startup the messages were sent to queue and on naviagte to weblog page the index is reset and sent a second time. But not sure what happend realy.

I see. the Web Log does reset the index to show all the previous log messages, so it makes sense. I should have thought about that before, but good fix!

PR #2373, opened 2025-01-25
support entities api endpoint for multiple devices of same type

proddy:
when there are multiple thermostats with different hc numbers, the api/thermostat/hc*/entities would always only show hc1.
Now api/thermostat/hc2/entities works


MichaelDvP (commented):
For the entities command i think it's better always loop all thermostats, annd add the circuits to root:
```
    bool found_device = false;
    for (const auto & emsdevice : emsdevices) {
        if (emsdevice->device_type() == devicetype) {
            found_device = true;
            emsdevice->get_value_info(root, cmd, id);
        }
    }
    // if the EMS device was valid, but the cmd not found exit. it will be handled upstream.
    if (found_device) {
        return root.size() > 0;
    }
```

Issue #2321, opened 2025-01-01
Device rename does not stick


Imwithstupid007:
Hi,

I use a BBQKees gateway E32 V2 (EMS-ESP V3.7.1) to monitor my boiler (Nefit Trendline II). In EMS-ESP the boiler shows up as "CBS/GB*72, Trendline, Cerapur, Greenstar Si". 

![Screenshot 2025-01-01 192456](https://github.com/user-attachments/assets/e3fc86e0-139e-4e93-a07a-3e5fa901f721)

I renamed the device several times to "Nefit Trendline II".

![Screenshot 2025-01-01 193804](https://github.com/user-attachments/assets/50e9c430-554a-4770-84d8-120397c38aea)

However, after some time (days-weeks) it defaults back to the standard.
The rename "survives" a normal reboot indicating that it is stored properly.

Why is it defaulting back to the standard?

MichaelDvP:
The name is stored in the customization file, you can check in settings->download the file and the name. 
Have you restored an older custoization?
Can you specify what have triggert the reset to default name? A crash of ems-esp? Are other data, especially in customization are affected?
We have not seen something like that before, but for mqtt settings there is a similar report that settings are cleared after some time (with unspecified conditions).

Imwithstupid007:
I have not restored an older customization.
Only thing i have done after receiving/connecting the device is to select a few favorites and rename the device (and made a mqtt connection).
The first few times this happened I assumed it was a result of "playing around" with the device. But the last time the ESP-EMS was operational for maybe 1-2 weeks and from one day to another the device name was default again. Other changes I made seem to be unaffected (mqtt functions, favorites are accounted for).

The device is restarted yesterday and I will check daily if the name is still correct.

Imwithstupid007:
PS: I assume a device restart is effectively the same as a power cycle?

tp1de:
I recognized the same behavior while updating between different versions.

MichaelDvP:
The rename option is added a few versions ago, using an older version (and changing customizations) will erase the setting. Also the logic for reading  and display the internal name from telegram 0x1 has changed and effects the custom name.

> PS: I assume a device restart is effectively the same as a power cycle?

Mostly, power cycle is a cold start, all registers are in powerup state, restart is a soft reset, u-ram values and some registers are not reseted. But this should not affect the values saved in filesystem like the custom name.

proddy:
[...]

PR #2439, opened 2025-03-02
ESPAsyncWebServer @ 3.7.2

proddy:
plus web UI cosmetic changes


mattreim:
Hey, while testing the new versions I noticed the following.

- `Remotehum` works for me via mqtt and HA with both versions
-  With #2436 the item values ​​under devices are no longer correct

EMS-Device
-  With #2439 the `custom_name` has been reset [again](https://github.com/emsesp/EMS-ESP32/issues/2321)

And one more little thing: Should `nrgheat` and `dhw/nrg` be writable by the user via the GUI?

EMS-Energy

Maybe it will help.

MichaelDvP:
> Should nrgheat and dhw/nrg be writable by the user via the GUI?

For heatpumps these are ems values and not writeable, for burners this is calcuated by ems-esp and ***is*** writeable to set a startvalue to nvs.

PR #2443, opened 2025-03-04
fix autofresh in web

proddy:
and add upload to Version page


mattreim:
Update German translation:

```lua
  UPLOAD_DROP_TEXT: 'Legen Sie eine Firmware-Datei (.bin) ab oder klicken Sie hier', 
```

Issue #2451, opened 2025-03-07
Improvement: Support nested conditional expressions in the scheduler


philwingfield:
### DESCRIPTION

The scheduler expression language current supports a single conditional express e.g. <cond> ? <expr1> : <expr2>
This is quite limited in scope and it would be nice to support nested ones to enable more complex functionality.

### REQUESTED INFORMATION

_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [x] Searched the issue in [issues](https://github.com/emsesp/EMS-ESP32/issues)
- [x] Searched the issue in [discussions](https://github.com/emsesp/EMS-ESP32/discussions)
- [x] Searched the issue in the [docs](https://docs.emsesp.org/Troubleshooting/)
- [x] Searched the issue in the [chat](https://discord.gg/3J3GgnzpyT)
- [ ] Provide the System information in the area below, taken from `http://<IP>/api/system`

Sorry I can't paste the system info but see here for the docs where it describes the limitation
https://docs.emsesp.org/Configuring/#conditions

### TO REPRODUCE

Try to enter any nested conditional or multiple conditionals e.g. any of the following will fail

- "1<2?(3<4?5:6):7" should give 5
- "1>2?(3<4?5:6):7" should give 7
- "1<2?(3>4?5:6):7" should give 6
- "1<2?3:(4<5?6:7)" should give 3
- "1>2?3:(4<5?6:7)" should give 6
- "1>2?3:(4>5?6:7)" should give 7
- "(1>2?3:4)+(10>20?30:40)" should give 44

### EXPECTED BEHAVIOUR

The above list gives the expected results for these expressions

### SCREENSHOTS

N/A

### ADDITIONAL CONTEXT

I have a fix ready for this and will raise a PR shortly

MichaelDvP:
Nested works with some rules, multiple is not supported.
The condition is always executed first, the parser searches ? and : before calculating a formula.
`a<b ? c<d ? 1 : e<f ? 2 : 3 : 4` works
`(a<b) ? (c<d) ? 1 : (e<f) ? 2 : 3 : 4` works also
but not with `?` or `:` within the bracket

also `(a<b ? 2 : 4) + 10` does not work
you have to write `a < b ? 2 + 10 : 4 + 10`

proddy:
@philwingfield out of interest, what was your use case for adding this? I could add it as a real life example to the docs

philwingfield:
[...]

PR #2452, opened 2025-03-07
Feat: Support nested conditionals

philwingfield:
Addresses #2451

Add support for nested conditionals, as long as they are surrounded by brackets. Added a bunch of unit tests for this as well.


proddy:
I like that you extrapolated the shuntingYard tests. At the moment its embedded into the WebScheduler code: https://github.com/emsesp/EMS-ESP32/blob/1d6c2c9664999ea442c48763426b427f64a646cf/src/web/WebSchedulerService.cpp#L573

I'll make a note to move it.

MichaelDvP:
- Nice additition to have a condition inside a formula.
- For the nested conditions now you need brackets, before brackets are not allowed. I think you should also check in the `skipBrackets()` for a colon to allow the non-bracket version. Then it will work with both notations
`cond1 ? cond2 ? expr1 : expr2 : espr3` and `cond1 ? (cond2 ? expr1 : expr2) : espr3`
- Before the conditions are calculated from left to right, now from right to left. 
Now all conditions have to be calculated, before only the relevant.
example: `cond1 ? cond2 ? expr1 : expr2 : expr3`
left to right for cond1=false only calculates cond1 and skips cond2, right to left starts with calculating cond2 and then calculate cond1 in any case.

MichaelDvP:
Inspired by your `skipBracket` algorithm i try to keep the left-to-right order and also allow nested conditions without brackets.
See lines 703-723: https://github.com/MichaelDvP/EMS-ESP32/commit/72feefe709283fe2bc347997462c33ff53b5fb54
What do you think?

philwingfield:
Oh great I will take a look today

philwingfield:
Yep, your implementation is great, all of my tests pass. I added a new one for a nested conditional without brackets and mine failed. 

I recommend merging your shuntingYard.hpp and ignoring mine. Up to you if you want to keep the new unit tests, I think they are still useful!

MichaelDvP:
If you are ok with it, i'll merge your PR with the tests to my repo and make a new PR to ems-esp.

proddy:
I'll wait, then we can release 3.7.2

philwingfield:
> If you are ok with it, i'll merge your PR with the tests to my repo and make a new PR to ems-esp.

Yes absolutely, whatever is easiest for you

MichaelDvP:
I've merged the tests from this PR and made PR #2458 
We can close ths PR.

PR #2466, opened 2025-03-13
added new return code when processing commands

proddy:
- add new NO_VALUE return code which we can use later if necessary, like in shuntingyard
- tidied up error messages on failed commands
- all API tests passed

### DIFF
--- a/src/core/command.cpp
+++ b/src/core/command.cpp
@@ -184,7 +184,8 @@ uint8_t Command::process(const char * path, const bool is_admin, const JsonObjec
     }
 
     // call the command based on the type
-    uint8_t return_code = CommandRet::ERROR;
+    uint8_t return_code = CommandRet::OK;
+
     if (data.is<const char *>()) {
         return_code = Command::call(device_type, command_p, data.as<const char *>(), is_admin, id_n, output);
     } else if (data.is<int>()) {
@@ -312,11 +313,17 @@ bool Command::set_attribute(JsonObject output, const char * cmd, const char * at
         return true;
     }
 
-    // not found
     output.clear();
 
+    // attribute isn't found
+    // it could be a value command, but the value doesn't exist?
+    if (strcmp(attribute, "value") == 0) {
+        logger_.debug("%s has no value set", cmd);
+        return false; // fail
+    }
+
     char error[100];
-    snprintf(error, sizeof(error), "no %s in %s", attribute, cmd);
+    snprintf(error, sizeof(error), "no attribute '%s' in %s", attribute, cmd);
     output["message"] = error;
     return false;
 }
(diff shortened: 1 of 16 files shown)


src/core/command.cpp
```diff
@@ -312,11 +313,17 @@ bool Command::set_attribute(JsonObject output, const char * cmd, const char * at
         return true;
     }
 
-    // not found
     output.clear();
 
+    // attribute isn't found
+    // it could be a value command, but the value doesn't exist?
+    if (strcmp(attribute, "value") == 0) {
+        logger_.debug("%s has no value set", cmd);
```
MichaelDvP, line 321: better use `LOG_DEBUG` as all other loggings uses the macro.

src/core/shuntingYard.hpp
```diff
@@ -380,7 +380,13 @@ std::string commands(std::string & expr, bool quotes = true) {
...
             JsonObject   input  = doc_in.to<JsonObject>();
             std::string  cmd_s  = "api/" + std::string(cmd);
-            emsesp::Command::process(cmd_s.c_str(), true, input, output);
+
+            auto return_code = emsesp::Command::process(cmd_s.c_str(), true, input, output);
+            // check for no value (entity is valid but has no value set)
+            if (return_code == emsesp::CommandRet::NO_VALUE) {
+                return expr = ""; // just ignore for now
+            }
+
```
MichaelDvP, line 389: I think this is not needed, it returns an empy string that we can insert here as `""`, should work with a compare like `boiler/rettemp==""` (to test if a sensor is connected)

MichaelDvP (approved):
i think all is working, just a few comments.

src/core/shuntingYard.hpp
MichaelDvP: Oh, empty strings crashes the shuntingYard, i think it's the s[0] access in isnum(), maybe also others. Will check.

src/core/command.cpp
proddy: good one, I'll change.

src/core/command.cpp
```diff
@@ -184,7 +184,8 @@ uint8_t Command::process(const char * path, const bool is_admin, const JsonObjec
     }
 
     // call the command based on the type
-    uint8_t return_code = CommandRet::ERROR;
+    uint8_t return_code = CommandRet::OK;
```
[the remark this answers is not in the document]
proddy: true, it's just to stop lint complaining of having an uninitialized variable

src/core/shuntingYard.hpp
proddy: if could check that would be good. I'll also add a test and try and reproduce (using pio run -e native-test -t exec)

MichaelDvP: This change work with empty strings:
https://github.com/MichaelDvP/EMS-ESP32/commit/071e81f29bbdb21d274a79d5aaf8849c5242d233
Tested with `boiler/rettemp == ""` (returns `1`, my boiler does not have a sensor)

proddy: shall I merge into my PR? Or you overlay on after I merge?

proddy: I'll PR now

PR #2472, opened 2025-03-21
make switchtime more clear in the german translation

JokerGermany:
"Programmschaltzeit 1" don't make clear that "Eigen 1" is meant.


proddy:
can't comment on whether this is correct - @MichaelDvP ?

MichaelDvP (approved):

PR #2519, opened 2025-04-13
Update German translation

mattreim:


interface/src/i18n/de/index.ts
```diff
@@ -331,7 +331,7 @@ const de: Translation = {
   ALLVALUES: 'Alle Werte',
   SPECIAL_FUNCTIONS: 'Sonderfunktionen',
   WAIT_FIRMWARE: 'Die Firmware wird hochgeladen und installiert.',
-  INSTALL_VERSION: 'Dadurch wird die Version {1} {0}. Sind Sie sicher?',
+  INSTALL_VERSION: 'Dadurch wird die Version {1} heruntergeladen. Sind Sie sicher?',
```
proddy, line 334: INSTALL_VERSION takes 2 arguments, the first {0} is either "download" or "install"

mattreim: In this context, it should be "heruntergeladen" and not "herunterladen".

proddy: it needs two arguments, other it will break the other 10+ translations. For DE I think it's:

```
LL.INSTALL='installiert'
LL.DOWNLOAD='heruntergeladen'
LL.INSTALL_VERSION='Dadurch wird Version {1} {0}.  Sind Sie sicher?'
```

mattreim:
I had to add two new entries because it is correct on the button.

Download

proddy:
were you going to modify `Version.tsx` with INSTALL_1 and DOWNLOAD_1 or shall I?

mattreim:
Thank you, I hope that's true.

proddy:
thanks, I have manually added your changes to the next PR

PR #2528, opened 2025-04-20
fix GH build

MichaelDvP:


MichaelDvP:
I have triggert the build action manually, so we have now the dev.8 binaries.

PR #2535, opened 2025-04-25
 show console log with ISO date/time, #2533 

MichaelDvP:
and fix debug-message in console help.
Updated to dev.10


proddy:
very nice, thanks @MichaelDvP !

PR #2620, opened 2025-09-07
Nested JSON structured in Scheduler HTTP fetch

nename0:
My Goal is to have a WLED instance with a Dallas temperature sensor as remote thermostat. The problem is however that the WLED API exposes the temperature value in a nested JSON strucuture. A `GET /json/info` returns:
```
{
  "ver":"0.15.1",
  ....
  "u":{"Temperature": [23.1,"°C"]},
  ...
}
```
Currently it is not possible to fetch that value using `{"url": "http://<wled-ip>/json/info"}` in the Scheduler.

This PR adds a `keys` key for the HTTP JSON object. `keys` is a JSON array of strings (for selecting object keys) or integers (for selecting array indexes).

So for the WLED API you can specify `{"url": "http://<wled-ip>/json/info", "keys": ["u", "Temperature", 0]}` to fetch the temperature.

### DIFF
--- a/src/core/shuntingYard.hpp
+++ b/src/core/shuntingYard.hpp
@@ -697,7 +697,7 @@ std::string compute(const std::string & expr) {
         JsonDocument doc;
         if (DeserializationError::Ok == deserializeJson(doc, cmd)) {
             HTTPClient  http;
-            std::string url, header_s, value_s, method_s, key_s;
+            std::string url, header_s, value_s, method_s, key_s, keys_s;
             // search keys lower case
             for (JsonPair p : doc.as<JsonObject>()) {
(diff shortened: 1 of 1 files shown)


nename0:
> @nename0, out of interest, why did you choose WLED to fetch the Dallas temperature reading? Are you displaying it, for example, on an OLED display?

No much simpler than that: I already had a WLED setup with a LED strip in my reference room. For the upcoming winter I wanted to try the remote thermostat. So I just had to add the sensor to the ESP. WLED did give me some trouble though: https://github.com/wled/WLED/pull/4916

MichaelDvP (approved):
Looks good and tests with reading a tasmota device status works good. 
BTW: std::string is empty after  declaration, so `key_s  = ""` is not needed, but also does not hurt.

PR #2630, opened 2025-09-21
API changes

proddy:
- update libs, in web and async*
- moved shuntingyard to a separate class object
- added @MichaelDvP's changes from his fork into the ShuntingYard
- updated unit tests to cover shuntingyard
- extended message command to process logic, like:
```sh
curl -X POST \
    -H "Authorization: Bearer ${emsesp_token}" \
    -H "Content-Type: application/json" \
    -d '{"value":"system/settings/locale"}' \
    ${emsesp_url}/api/system/message
```

### DIFF
--- a/src/test/test.cpp
+++ b/src/test/test.cpp
@@ -1105,6 +1105,7 @@ void Test::run_test(uuid::console::Shell & shell, const std::string & cmd, const
         // request.url("/api/system/message");
         // EMSESP::webAPIService.webAPIService(&request, json);
 
+        // output: 70.00
         request.method(HTTP_POST);
         char data1[] = "{\"value\":\"custom/test_custom\"}";
         deserializeJson(doc, data1);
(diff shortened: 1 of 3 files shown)


proddy:
I'm hitting some crashes of the ESP32 with my latest updates, somewhere json docs are getting corrupted. I can't find the cause so need to dig out my esp-prog/JTAG debugger and go through the code. So it'll take some time, then I'll manually merge your latest PR @MichaelDvP

proddy:
@MichaelDvP I had to increase the asynctcp stack from 6KB to 8KB to prevent the ESP32 crashing. I'm not sure my change to re-route command via the shunting yard needs more memory though.

src/core/command.cpp
```diff
@@ -317,15 +322,16 @@ bool Command::set_attribute(JsonObject output, const char * cmd, const char * at
 
     // attribute isn't found
     // it could be a value command, but the value doesn't exist?
-    if (strcmp(attribute, "value") == 0) {
+    if (strncmp(attribute, "value", 5) == 0) {
```
MichaelDvP, line 325: strcmp compares to the terminating zero, this case max to 6th char in "value", with strncmp it matches any name that starts with "value".

MichaelDvP (approved):
Seems you have done a lot of tests. 
I've only tested on the esp and looked (fast) through the changes.
Only one comment.

src/core/command.cpp
proddy: good spot, I'll change it.

PR #2701, opened 2025-11-01
web optimizations

proddy:
I've optimized all the web code (100+ files) with some help from AI to clean up the code and use ReactJS Memo and Callback's for faster caching. The benefits
 - much faster building on standalone and production
 - 20% faster response times in the web browser, especially in table rendering. I'll do some real performance testing in https://pagespeed.web.dev soon
 - cleaner and more readable code
 - fixed some bugs, like the memory leak when running in standalone and the invalidated browser cache between new releases


MichaelDvP:
Testing now, looks good. 
First small finding in headline  (&nbsp;):
grafik

PR #2717, opened 2025-11-05
fix some bugs, add some text

proddy:
- call system ap/enabled was broken so I added ap and removed all the commands, only showing the device (settings, mqtt, ap etc)
- count for Tx Quality was wrong
- show languages and number of entities in log
- some optimisations for efficiency in emsesp.cpp

### DIFF
--- a/CHANGELOG_LATEST.md
+++ b/CHANGELOG_LATEST.md
@@ -49,6 +49,7 @@ For more details go to [docs.emsesp.org](https://docs.emsesp.org/).
 - ventilation bypass state from telegram 0x55C [#1197](https://github.com/emsesp/EMS-ESP32/issues/1197)
 - set selflowtemp for ems+ boilers [#2641](https://github.com/emsesp/EMS-ESP32/discussions/2641)
 - syslog timestamp [#2704](https://github.com/emsesp/EMS-ESP32/issues/2704)
+- fixed FS format command [#2720](https://github.com/emsesp/EMS-ESP32/discussions/2720)
 
 ## Changed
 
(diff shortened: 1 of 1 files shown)


src/core/system.cpp
```diff
@@ -1668,6 +1681,21 @@ bool System::command_info(const char * value, const int8_t id, JsonObject output
...
+#ifndef EMSESP_STANDALONE
+        node["channel"]    = settings.channel;
+        node["ssidHidden"] = settings.ssidHidden;
+        node["maxClients"] = settings.maxClients;
+        node["localIP"]    = settings.localIP.toString();
+        node["gatewayIP"]  = settings.gatewayIP.toString();
+        node["subnetMask"] = settings.subnetMask.toString();
+#endif
+    });
+
```
MichaelDvP: We have provisionMode, ssid and security already in the network section. Do we need all setiings for debugging (system/info)?

src/core/emsesp.cpp
```diff
@@ -846,41 +880,55 @@ std::string EMSESP::pretty_telegram(std::shared_ptr<const Telegram> telegram) {
...
         }
     }
 
-    if (type_name.empty()) {
-        // 2nd fallback, get the type name from dest
+    // Fallback for type name if not found - try src or dest
+    if (!type_found && telegram->operation != Telegram::Operation::RX_READ) {
         for (const auto & emsdevice : emsdevices) {
-            if (telegram->operation != Telegram::Operation::RX_READ && emsdevice->is_device_id(dest)) {
+            if (emsdevice->is_device_id(src) || emsdevice->is_device_id(dest)) {
```
MichaelDvP: This is not the same as before. Now we get the first entry for scr or dest, Before src have priority. 
I think the 0x2CC with different meaning/names for RC300 and boiler is a bug, But we have it in code as reported by users.

proddy: thanks, I'll remove the `emsdevice->is_device_id(src)`

proddy:
@MichaelDvP I sneaked in one last minute change. I added a new type called is_sensor to the Analog class, which we can use for internal sensors like core_voltage and supply_voltage, and preventing the user from removing them. I noticed if you delete them you can never get them back without doing a full factory reset. Are you ok with this change?

PR #2724, opened 2025-11-09
fixes to system sensors

proddy:
- fixes to last PR for system/internal analog sensors
- also add is_system to temperature sensors
- highlight system sensors in Sensors page
- remove system sensors from Dashboard (they are technical)
- change Update button icon to be consistent with other dialogs


mattreim:
@proddy With version v3.7.3-dev.26, the entries of the remote thermostats are no longer displayed and are also no longer transmitted via MQTT.

Devices:
RC100H
 MQTT(HA):
MQTT

PR #2730, opened 2025-11-10
Update German translation

mattreim:
Hello, Just one small change and two comments.
- The "no data" text is not translated on the dashboard page.
- Perhaps it's a bit cheeky, but I personally don't find a logout page very appealing; the popup menu was simpler.

### DIFF
--- a/interface/src/i18n/de/index.ts
+++ b/interface/src/i18n/de/index.ts
@@ -40,7 +40,7 @@ const de: Translation = {
   RUN_COMMAND: 'Befehl ausführen',
   CHANGE_VALUE: 'Wert ändern',
   CANCEL: 'Abbrechen',
-  REMOVE_ALL: 'Entfernen alles',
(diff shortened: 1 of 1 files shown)


proddy:
I had to remove the popup when selecting the user in the left-pane window to make the selection consistent. I understand your point about the User menu and sign-out. There are two options
 - keep as is (this is also how Home Assistant does it)
 - remove the 'user' menu option on the left and replace with an icon top right which shows a popup (this is how Unifi does it)
 
To make it more useful I can think of other details to add to the page, such as changing the password. Thoughts?

proddy:
The top-right option is what we had for the last 5 years since version 3.0 -> 3.6, and then moved it in 3.7 to be more like Home Assistant. You can see older screenshots on https://docs.emsesp.org/About/ . I don't fancy going back to the old way as I'm sure there was a reason it was moved.

PR #2744, opened 2025-11-15
webUI gpio improvements

proddy:
- list of valid GPIOs is determined by the backend now, in system()
- this removes some JS code, specially in the validations
- selecting a GPIO in the Analog is a pull-down menu only showing the valid pins to select from
- Application Settings/CUSTOM board there are no checks, assuming this is only for advanced users anyway
- Changed Analog Sensors so you can only rename a system sensor

### DIFF
--- a/interface/src/app/main/SensorsAnalogDialog.tsx
+++ b/interface/src/app/main/SensorsAnalogDialog.tsx
@@ -153,6 +153,20 @@ const SensorsAnalogDialog = ({
     [creating, LL]
   );
 
+  // Ensure the current GPIO is in the list when no creating
+  // note GPIO 99 means not set
+  const availableGPIOs = useMemo(() => {
+    const filteredList = analogGPIOList.filter((gpio) => gpio !== 99);
+    if (
+      editItem.g !== undefined &&
+      editItem.g !== 99 &&
+      !filteredList.includes(editItem.g)
+    ) {
+      return [...filteredList, editItem.g].sort((a, b) => a - b);
+    }
+    return filteredList;
+  }, [analogGPIOList, editItem.g]);
+
   return (
     <Dialog sx={dialogStyle} open={open} onClose={handleClose}>
       <DialogTitle>{dialogTitle}</DialogTitle>
@@ -174,9 +188,8 @@ const SensorsAnalogDialog = ({
               sx={{ width: '8ch' }}
               select
               onChange={updateFormValue}
-              disabled={editItem.s}
             >
-              {analogGPIOList?.map((gpio: number) => (
+              {availableGPIOs?.map((gpio: number) => (
                 <MenuItem key={gpio} value={gpio}>
                   {gpio}
                 </MenuItem>


src/core/system.cpp
```diff
@@ -2320,4 +2298,71 @@ uint8_t System::systemStatus() {
...
+#elif CONFIG_IDF_TARGET_ESP32S2
+    std::vector<uint8_t> valid_gpios = string_range_to_vector("19-20, 22-32, 40");
+#elif CONFIG_IDF_TARGET_ESP32S3
+    std::vector<uint8_t> valid_gpios = string_range_to_vector("19-20, 22-37, 39-42, 48");
+#elif CONFIG_IDF_TARGET_ESP32 || defined(EMSESP_STANDALONE)
+    std::vector<uint8_t> valid_gpios = string_range_to_vector("1, 6-11, 16-17, 20, 24, 28-31, 40");
+#else
+    std::vector<uint8_t> valid_gpios = {};
+#endif
+
```
MichaelDvP: Look like you have listed invalid pins as valid. I've checked the datasheet: Unclear, list strapping pins (esp32 0,2,5)? How to manage input only pins for esp32: 34-39
ESP32 valid: 0, 2, 4, 5,12-15, 18, 19, 23, 25-27 ,32-39
S3 valid: 2, 4-14, 17, 18, 21, (33-38), 45, 46
33-38 are used for octal SPI for 32M vchip version
S2 valid: 0-14, 19, 20, 21 ,33-38, 45,46
C3 valid: 0-10

proddy: Thanks for reviewing. I based the valid pins logic on what we already had in code, specifically in `validators.ts` and `System::is_valid_gpio()`, without making any changes.

Pin 0 is a special case because we sometimes use this to indicate disabled, as in Dallas and LED

I would block the GPIO input only pins for now. If someone needs it we can always make an exception. 

Shall I adjust based on what you researched, or commit and you make the changes?

proddy: done.

MichaelDvP:
No, i mean: 
grafik
Sensor test1 is on gpio23, but in dialog the gpio field is empty and gpio 23 can not be selected. For editing/renaming an existing sensor the actual gpio shuld be listed in the select box. For creating a new sensor a empty box is ok,

MichaelDvP:
GPIO 23 is not valid in this case, because an analog sensor already uses it: the sensor i want to edit/change name.

MichaelDvP:
There are issues with board gpios. When configuring we check gpios for pbutton, led, etc like in:
https://github.com/emsesp/EMS-ESP32/blob/60b0de79b30ab5df9f05487c920c154f0af97dbd/src/core/system.cpp#L586
but the valid_gpio vector excludes `led_gpio_`, so it is always shon as invalid. I think we should make all checks in the reload function like:
```
void System::reload_settings() {
    // set gpios to zero for valid check
    led_gpio_  = 0;
    EMSESP::webSettingsService.read([&](WebSettings & settings) {
        led_gpio_       = is_valid_gpio(settings.led_gpio) ? settings.led_gpio : 0;
```
and exclude gpio==0 for all except pbutton. 
I make achange and let you know.

PR #2750, opened 2025-11-17
adjustments to invalid sensors

proddy:
on top of https://github.com/emsesp/EMS-ESP32/pull/2748 ...

fixes being able to add a new sensor with no GPIO and geting an error message
removed 99 as a bad/disabled analog GPIO, uses type now AnalogSensor::AnalogType::NOTUSED and only allows valid gpios
optimized code, reducing number of lines in Web (e.g. validation)
made standalone test data dynamic to match C++ backend logic
GPIOs validated in associated functions, and don't use 0 for disabled

### DIFF
--- a/interface/src/app/main/Customizations.tsx
+++ b/interface/src/app/main/Customizations.tsx
@@ -285,13 +285,17 @@ const Customizations = () => {
     return value as string;
   }
 
+  const isCommand = useCallback((de: DeviceEntity) => {
+    return de.n && de.n[0] === '!';
+  }, []);
+
   const formatName = useCallback(
     (de: DeviceEntity, withShortname: boolean) => {
       let name: string;
-      if (de.n && de.n[0] === '!') {
+      if (isCommand(de)) {
         name = de.t
-          ? `${LL.COMMAND(1)}: ${de.t} ${de.n.slice(1)}`
-          : `${LL.COMMAND(1)}: ${de.n.slice(1)}`;
+          ? `${LL.COMMAND(1)}: ${de.t} ${de.n?.slice(1)}`
+          : `${LL.COMMAND(1)}: ${de.n?.slice(1)}`;
       } else if (de.cn && de.cn !== '') {
         name = de.t ? `${de.t} ${de.cn}` : de.cn;
       } else {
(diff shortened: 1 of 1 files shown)


proddy:
as soon as GH push is working again, I'll upload the latest changes and then if you could test agan @MichaelDvP

src/core/analogsensor.cpp
```diff
@@ -157,12 +157,13 @@ void AnalogSensor::reload(bool get_nvs) {
             if (!found) {
                 // check if the GPIO is valid before registering. If not, force set the sensor to disabled, but don't remove it
                 // should only trigger if uploading a customization file with invalid gpios.
+                AnalogType type = static_cast<AnalogType>(sensor.type);
                 if (!EMSESP::system_.is_valid_gpio(sensor.gpio)) {
```
MichaelDvP: This check is obsolete, you now check in WebcusomizationService and remove the sensor.

MichaelDvP (approved):
Works now for analog gpios. but not checking used gpios for analog and board is imo a step back. The goal for less code is not reached it's much more now.

For S3 we need to add the gpio 0, 1 to the vector here https://github.com/proddy/EMS-ESP32/blob/c9bddba446f16e3eb33ba6c39e27b7b551247611/src/core/system.cpp#L2375
Also esp32 should contain gpio 1 as valid, E32 uses it for ETH.power.

I have not tested latest change to board profile. The automatic board-check works not if ETH is already started, So it's critical to set board_profile to default when a gpio is not valid. My E32V2_2 sets S32 when changing LED to gpio 1.

MichaelDvP:
Nicer, but in web sensor type is shifted by one.

MichaelDvP:
Now it works (Maybe i have missed to rebuild the web after last merge, now i rebuild all).

proddy:
> Now it works (Maybe i have missed to rebuild the web after last merge, now i rebuild all).

ok. I remember I did a -1 in the web Menu to compensate. I'm still testing the new changes for board_proflle and gpio checking. I'll commit later today when I'm back home

proddy:
I committed my changes, but it's still not 100% perfect when changing a board profile in the webUI. I'll spend some time tomorrow to take a look and hopefully finish this.

PR #2793, opened 2025-12-04
PSram memory

MichaelDvP:
and some other changes. Please check, work is still in progress,
Normally psram should be used for all larger objects, there are some settings in sdkconfig to control the usage.
But changing the settings 
`CONFIG_SPIRAM_MALLOC_ALWAYSINTERNAL=4
CONFIG_SPIRAM_MALLOC_RESERVE_INTERNAL=196000`
have no effect.tested with a local `sdkconfig.deaults` and in the mail sdkconfig in the framework tools folder.. So i've looked for other possibilities and found https://github.com/pschatzmann/esp32-psram
Some of the fles does not compile with actual arduino, but i only use `AllocatorPSRAM.h` to move vectors, lists, etc. to psram if possible. 

Please check. I'm not sure it's the right approach, but on E32v2 i get much more free heap. But i have not checked all timings.

BTW: fixed the syslog start/stop without reboot (mentioned on discord). I'm not sure what was wrong, but not it works for me.

### DIFF
--- a/interface/package.json
+++ b/interface/package.json
@@ -41,7 +41,7 @@
     "react": "^19.2.1",
     "react-dom": "^19.2.1",
     "react-icons": "^5.5.0",
-    "react-router": "^7.10.0",
+    "react-router": "^7.10.1",
     "react-toastify": "^11.0.5",
     "typesafe-i18n": "^5.26.2",
     "typescript": "^5.9.3"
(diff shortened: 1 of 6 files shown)


proddy:
nice, I'll check next mid-week when I'm back home (currently travelling)

proddy:
Did you make a stub of AllocatorPSRAM.h for EMSESP_STANDALONE so we can remove all those surplus #ifdefs?

proddy:
ok done.

I must say @MichaelDvP , your work on PSRAM is really impressive. I'm seeing a +12KB gain in available heap. Amazing stuff!

PR #2805, opened 2025-12-12
Fix system/metrics endpoint duplicate labels & add enum support for metrics

gr3enk:
This PR fixes an issue where the /api/system/metrics endpoint printed duplicate labels for some Prometheus metrics:

```
// old
emsesp_ntp_info{ntpstatus="disconnected", server="pool.ntp.org", tzlabel="Europe/London", ntpstatus="disconnected"} 1

// new
emsesp_ntp_info{ntpstatus="disconnected", server="pool.ntp.org", tzlabel="Europe/London"} 1
```

Also added enum support for prometheus metrics alongside numbers and booleans.

### DIFF
--- a/src/core/system.cpp
+++ b/src/core/system.cpp
@@ -1635,7 +1635,7 @@ std::string System::get_metrics_prometheus() {
     };
 
     // helper function to process a JSON object recursively
-    std::function<void(const JsonObject &, const std::string &)> process_object = [&](const JsonObject & obj, const std::string & prefix) {
+    std::function<void(const JsonObject, const std::string &)> process_object = [&](const JsonObject obj, const std::string & prefix) {
         std::vector<std::pair<std::string, std::string>> local_info_labels;
         bool                                             has_nested_objects = false;
 


gr3enk:
Yes, I was checking as well right now. 

- 11601 bytes for system
- 6840 bytes for thermostat (71 entitites)
- 19412 bytes for boiler (174 entities)

i think 16k fixed for system and 160 for each entity is good.

src/core/emsdevice.cpp
```diff
@@ -1699,14 +1699,33 @@ std::string EMSdevice::get_metrics_prometheus(const int8_t tag) {
...
+    };
+
+    // Dynamically reserve memory for the result
+    size_t entity_count = 0;
+    for (const auto & dv : devicevalues_) {
+        if (tag >= 0 && tag != dv.tag) {
+            continue;
+        }
+        // only count supported types
+        if (is_supported_type(dv.type)) {
```
MichaelDvP: Better `if (dv.hasValue() && is_supported_type(dv.type)) {`
some boiler types have a lot of entities without values. and you generate the metrics also only for entities with values.

PR #2829, opened 2025-12-19
HA fixes part 3

proddy:
rollback changes in scheduler (force)
some minor clean-up
added v to "sw"

### DIFF
--- a/src/web/WebCustomEntityService.cpp
+++ b/src/web/WebCustomEntityService.cpp
@@ -383,12 +383,6 @@ void WebCustomEntityService::publish() {
         return;
     }
 
-    if (Mqtt::publish_single()) {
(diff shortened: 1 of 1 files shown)


MichaelDvP:
Yes and no, We have mqttclient set to always queue messages
https://github.com/emsesp/EMS-ESP32/blob/c8b4a38bb687a9a2c4c3ce62b68a595b07b9d4a4/lib/espMqttClient/src/Config.h#L40
but check connected before queing the messages here:
https://github.com/emsesp/EMS-ESP32/blob/c8b4a38bb687a9a2c4c3ce62b68a595b07b9d4a4/src/core/mqtt.cpp#L593

Afair there were users with bad mqtt config and queue was running full until web-ui was not responsive anymore.

Issue #2837, opened 2025-12-21
allow to switch between firmware partitions (to make previous upload active)


proddy:
### Discussed in https://github.com/emsesp/EMS-ESP32/discussions/2831

<div type='discussions-op-text'>

<sup>Originally posted by **VlastiBroucek** December 20, 2025</sup>
Recently, I have noticed few people asking how they could get to previous DEV release and the answer was that the only way is to do their own build.

I think not everyone can do that - I for one did not attempt that as yet and I consider myself to be IT veteran, but not in ESP32.

But I have another idea. I realised that it is possible to restart to previous DEV from telnet console by using `restart [partitionname]` and either `app0` or `app1` in `su` mode. The partition other than the one currently used seems to be generally holding the previous install. I used it myself few times recently.

So, my idea(s) are:

- to have new command in telnet to tell us version installed in each partition so we can verify before using `restart` command,
- to have similar information somewhere in GUI, and
- to have RESTART button in GUI with option to restart to previous version.

As I said, my ESP32 knowledge is fairly limited, so not sure if this is possibly doable, but it might be a nice improvement and can be beneficial when there is a need to quickly compare behaviour of two versions.</div>

proddy:
This is now implemented in 3.7.3-dev.40

- The `show` command lists the partitions and like last time you can switch using the `restart <partition>` command
- The WebUI will show all stored firmware versions in any `app` partitions (app0,app1,boot,factory) with an Info button where you can pull up details of the build and install date. 

Note this only works after dev-40 is installed.

Issue #2825, opened 2025-12-18
Analog Sensors>CounterX entity can't be added to the HA energy dashboard


ESP4Ever:
Hello!
Currently it is not straight-forward of adding a counter entity to the HA energy dashboard due to lack of "device_class" definition. 
Could you please add a "device_class" selector to the Analog Sensors>CounterX window with "none", "energy", "power", "gas" and "water" values?
"none" = not needed, don't add

proddy:
we'll add this to the next major version. I don't want to introduce any more breaking changes or big code additions right now as we're trying to get the next stable version released before Santa arrives

PR #2840, opened 2025-12-22
a selection of minor fixes and some new features

proddy:
Some fixes and features

- fixes GPIOs not checked when board profile is adjusted #2841
- changes to state_class and device_class for HA (mainly analog sensors)
- some changes for Analog Sensors>CounterX entity can't be added to the HA energy dashboard #2825
- implements rollbacking to previous versions from OTA partitions #2837
- use LED as visual indicator when using the button #2848
- add back ability to set TxMode to off #2850

### DIFF
--- a/interface/pnpm-lock.yaml
+++ b/interface/pnpm-lock.yaml
@@ -2884,8 +2884,8 @@ packages:
     resolution: {integrity: sha512-pkonvlKk8/ZuR0D5tLW8ljt5I8kmxp2XKymhepUeOdCEfKpZaktSArkLHZt76OB1ZvO9bssUsDty4SWhLvZpLg==}
     engines: {node: '>=0.10.0'}
 
-  ts-api-utils@2.2.0:
-    resolution: {integrity: sha512-L6f5oQRAoLU1RwXz0Ab9mxsE7LtxeVB6AIR1lpkZMsOyg/JXeaxBaXa/FVCBZyNr9S9I4wkHrlZTklX+im+WMw==}
+  ts-api-utils@2.3.0:
+    resolution: {integrity: sha512-6eg3Y9SF7SsAvGzRHQvvc1skDAhwI4YQ32ui1scxD1Ccr0G5qIIbUBT3pFTKX8kmWIQClHobtUdNuaBgwdfdWg==}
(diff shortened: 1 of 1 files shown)


mattreim:
I would also like to take this opportunity to wish you all happy holidays.

MichaelDvP (commented):
would be good to clear the snapshot gpios after they are not used anymore. In non psram systems these vectors are on the heap.

src/core/system.cpp
```diff
@@ -291,33 +304,138 @@ void System::store_nvs_values() {
...
+            if (EMSESP::nvs_.isKey(part->label)) {
+                p_info.version = EMSESP::nvs_.getString(part->label).c_str();
+            } else {
+                p_info.version = "";
+            }
+
+            // get install date from NTP if active and add
+            if (ntp_connected_) {
+                char   time_string[25];
+                time_t now = time(nullptr) - uuid::get_uptime_sec();
```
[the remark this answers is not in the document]
proddy: oops, fixed

proddy:
@MichaelDvP I had the last two days off work so some time to work on some small changes. I haven't tested on all boards yet but I think its mostly ok.

I may decide to change the new system command `txenabled` to `tx/enabled` to be consistent with the other commands.

src/core/emsesp.cpp
```diff
@@ -1810,10 +1811,14 @@ void EMSESP::shell_prompt() {
 
 // main loop calling all services
 void EMSESP::loop() {
-    uuid::loop();         // store system uptime
-    esp32React.loop();    // web services
-    system_.loop();       // does LED and checks system health, and syslog service
-    webLogService.loop(); // log in Web UI
```
MichaelDvP: I did not find this weblog call in the new version.

proddy: oops again! that is why I love pair-programming with you!

proddy:
@MichaelDvP I found another issue that needs fixing. When setting the install_date it happens before NTP is active, so never works. I'll make a fix soon.

proddy:
oh, I implemented it differently. Just checked the code in . When NTP is connected it will update the internal EMS-ESP partition list with the install date, but only once.

PR #2847, opened 2025-12-24
Added some missing Czech translations

VlastiBroucek:
I have added some missing Czech translations. It's my native language but I've been living in AU since 1992, so I have added only those I am positive about. The language is changing terribly, plus some technical terms are "questionable".


proddy:
nice, thanks!

Issue #2901, opened 2026-01-06
determine valid GPIOs based on EMS-ESP board profile


proddy:
We determine the available GPIOs today by the model/chipset (ESP32-C3, ESP32-S2, ESP32S3 and the basic ESP32). 

We should change the logic to base this on the EMS-ESP board type (S32, E32, E32V2, E32V2_2, MH-ET, NODEMCU, LOLIN, OLIMEX, OLIMEXPOE, C3MINI, S2MINI, S3MINI, S32S3 and CUSTOM).

This will ensure BBQKees CE-certified boards stay certified and not misused or bricked by accident. For these BBQKees gateway boards use of arbitary GPIOs in the Analog Sensor page will be restricted.

If the board profile is CUSTOM we can apply the chipset logic as it stands today.

PR #2904, opened 2026-01-07
 GPIOs depends on board profile, #2901 

MichaelDvP:
Checked this with my E32V2_2 board, gpio33 (board NTC) not included for E32V2_2, but gpio2 (singleLED) can be configured as analog.
Boardprofile E32V2_2
 GPIOs:
  in use: 4(UART Rx) 5(UART Tx) 32(LED) 14(Dallas) 34(Button) 39(core_voltage) 36(supply_voltage) [total 7]
  available: 2 [total 1]

Boardprofile CUSTOM:
 GPIOs:
  in use: 4(UART Rx) 5(UART Tx) 32(LED) 14(Dallas) 34(Button) 39(core_voltage) 36(supply_voltage) 33(board) 35(in) [total 9]
  available: 2 19 21 22 25 26 27 37 38 [total 9]

For boards coded by fuse there is only this board and custom selectable, older BBQKees boards are restricted, but can by changed to any other board profile. 

@bbqkees please also check if this the way you think of.

### DIFF
--- a/test/test_api/test_api.h
+++ b/test/test_api/test_api.h
@@ -278,8 +278,7 @@ void test_25() {
     auto expected_response =
         "[{\"api_data\":\"# HELP emsesp_system_uptimesec uptimeSec\\n# TYPE emsesp_system_uptimesec gauge\\nemsesp_system_uptimesec 0\\n# HELP "
         "emsesp_system_txpause txpause\\n# TYPE emsesp_system_txpause gauge\\nemsesp_system_txpause 0\\n# HELP emsesp_system_info info\\n# TYPE "
-        "emsesp_system_info gauge\\nemsesp_system_info{version=\\\"dev\\\", resetreason=\\\"Unknown / Unknown\\\", gpios_in_use=\\\"23, 5, 2, 18, 0\\\", "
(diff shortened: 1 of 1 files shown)


src/web/WebStatusService.cpp
```diff
@@ -49,8 +49,6 @@ void WebStatusService::systemStatus(AsyncWebServerRequest * request) {
     auto *     response = new AsyncJsonResponse(false);
     JsonObject root     = response->getRoot();
 
-    root["emsesp_version"] = EMSESP_APP_VERSION;
```
proddy: this is critical for Status and Version WebUI pages to work.

MichaelDvP:
> GPIO02 is still used for the single color LED

Should we add this as system sensor `led`, then it can be used from scheduler or command and the board shows no free gpios.

I've added a check to show CUSTOM board selection only in developer mode.

proddy:
> > GPIO02 is still used for the single color LED
> 
> Should we add this as system sensor `led`, then it can be used from scheduler or command and the board shows no free gpios.

Good idea.

Issue #2908, opened 2026-01-08
3.8.1dev5 Custom board no longer supported


ESP4Ever:
Hi
Just updated from dev3 to dev5 and immediately got a total config failure. 
After factory reset found no custom board in the list

proddy:
living on the edge @ESP4Ever 😄 

@MichaelDvP `system::check_upgrade()` needs to be adjusted to automatically set `developer mode` if the board profile is CUSTOM

ESP4Ever:
> living on the edge [@ESP4Ever](https://github.com/ESP4Ever) 😄

More like a blade runner...

> enable developer mode in settings to switch to custom board.

It was enabled but it didn't save from a config crash(

MichaelDvP:
> @MichaelDvP system::check_upgrade() needs to be adjusted to automatically set developer mode if the board profile is CUSTOM

If custom is set, it shows in web interface also if developer mode is off:
https://github.com/emsesp/EMS-ESP32/blob/577964befefd6cc3a32f6bfb89475f857bb53cd9/interface/src/app/settings/ApplicationSettings.tsx#L506-L509

MichaelDvP:
Try dev.6

ESP4Ever:
RTL8201 PHY uses these pins
<img width="345" height="156" alt="Image" src="https://github.com/user-attachments/assets/c61541fc-2169-4ddc-8264-f82d87502075" />

but they are available in the analog GPIO list
<img width="301" height="568" alt="Image" src="https://github.com/user-attachments/assets/f521f703-7c3c-4025-b53c-5dc7d184cb4a" />

ESP4Ever:
dev8 - no GPIO selection problems anymore, thank you so much for a quick fix!
The issue can be closed

proddy:
can you close it please

PR #2909, opened 2026-01-09
HA number mode selection #2900, board specific gpio settings

MichaelDvP:
I set to all predefied esp32 boards the specific GPIO, also exclude uart0 on the board that have a chip connected (some eth boards does not have uart/usb chip). The MH-ET datasheet lists pads `"can be used always"` and `"if no other function"`, i have include only the `always pins`
BBQKess boards are already reduced to no free gpio. I think the E32V2 is always sold without relais (i have them on my prototype).
`Custom` allow now also the strapping pins, that are often used for ETH connection, e.g. #2908, WT32-ETH01, Lilygo eth-lite, other olimex like ESP32-POE2, etc.
Fixed the comments and seperate BBQKees specific and other boards, btw: ADC2 is used for WiFi, no conflict with 36, 39 on E32V2_2.
No changes for C3, S2, S3 boards.


proddy:
nice work. I'll merge now, then test, and if all good we'll make this 3.8.1 and release as stable

PR #2912, opened 2026-01-09
update testdata (no `.0`-decimales), fix c3 compile issue

MichaelDvP:


src/core/emsdevicevalue.h
```diff
@@ -76,7 +76,7 @@ class DeviceValue {
         MBAR,        // 24 - mbar - atmospheric pressure
         LH,          // 25 - l/h - volume flow rate
         CTKWH,       // 26 - ct/kWh - monetary
-        HZ,          // 27 - Hz - frequency
+        HERTZ,       // 27 - Hz - frequency
```
proddy: that's a bit drastic. I would also mean changing ha_classes as HA expects HZ as UOM. An easier, safer alternative is to add this to `emsdevicevalue.h`

```
#ifdef HZ
#undef HZ
#endif
```

proddy: ah, yes, ok!

PR #2916, opened 2026-01-10
webUI changes and comments

proddy:


proddy:
> Im asking me why we discuss board profile here and in discord before the change and now you remove the custom option from BBQKees boards without any discussion? Trump style?

@MichaelDvP the last PR was just fixing reactJS warnings, no logic was changed. If so, it wasn't intentional. What is different now? Again, instead of being German and direct it would be more helpful if you actual describe what you're seeing

MichaelDvP:
> What is different now? 

With activated developer mode you can't switch from BBQKees board to CUSTOM anymore.

PR #2934, opened 2026-01-24
a collection of small changes

proddy:
** DRAFT **

- adjusted web cache, makes it feel snappier
- replace ordererd_map with map to save on heap
- add PSRAM to UUID libs (log etc)
- optimizations to reduce string concat (just the big ones)
- optimizations, adding volatile and const where needed (mutex will be next)
- emsfactory refactor


proddy:
@MichaelDvP  anything interesting worth while in this PR to merge?

MichaelDvP:
Looks good to me. Tested and no issues found. For my system with only ein hc and one dhw circuit memory changes only very little, but i think for some systems it makes a difference.

MichaelDvP (approved):

proddy:
> Looks good to me. Tested and no issues found. For my system with only ein hc and one dhw circuit memory changes only very little, but i think for some systems it makes a difference.

Thanks for taking a look. I ran some simulation by running with -DEMSESP_TEST and then manually adding thermostats and heating circuits (with the add command).

Issue #2936, opened 2026-01-28
v3.8.0 / v3.8.1 / v3.8.2dev: Software reset CPU after 2-3 minutes when MQTT-AutoDiscovery enabled (due to single room control sensors?)


ruegr:
### DESCRIPTION

I'm using a BBQKees EMS Gateway E32 V2 to monitor and control my heat pump and ventilation. The gateway constantly reboots after 2-3 minutes with this info message:
`Last system reset reason Core0: Software reset CPU, Core1: Software reset CPU`

Noteworthy: I have 15 thermostats (thiw230, SRC single room control) connected to the ConnectKey K40RF.

The problem occurs with firmware 3.8.0+ (tested: 3.8.0, 3.8.1, 3.8.2-dev0) as soon as MQTT with Discovery is enabled. When just MQTT ist enabled (without Discovery) the gateway runs (at least) for hours without any problem.

With an older firmware version (iE 3.7.2) the gateway doesn't reboot with MQTT and Discovery enabled.

### REQUESTED INFORMATION

_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [x] Searched the issue in [issues](https://github.com/emsesp/EMS-ESP32/issues)
- [x] Searched the issue in [discussions](https://github.com/emsesp/EMS-ESP32/discussions)
- [x] Searched the issue in the [docs](https://emsesp.org/Troubleshooting/)
- [x] Searched the issue in the [chat](https://discord.gg/3J3GgnzpyT)
- [x] Provide the System information in the area below, taken from `http://<IP>/api/system`

System information with v3.7.2
```json
  {
  "system": {
    "version": "3.7.2",
    "uptime": "000+01:23:15.100",
    "uptimeSec": 4995,
    "platform": "ESP32",
    "cpuType": "ESP32-D0WD-V3",
    "arduino": "ESP32 Arduino v2.0.17",
    "sdk": "v4.4.7-dirty",
    "freeMem": 145,
    "maxAlloc": 99,
    "freeCaps": 4159,
    "usedApp": 2243,
    "freeApp": 2429,
    "partition": "app0",
    "flash_chip_size": 16384,
    "resetReason": "Software reset CPU / Software reset CPU",
    "psram": true,
    "psramSize": 4093,
    "freePsram": 4047,
    "model": ""
  },
  "network": {
    "network": "Ethernet",
    "hostname": "ems-esp",
    "TxPowerSetting": 0,
[...]

Issue #2935, opened 2026-01-27
New entities "comfort point" / "comfort point" offset for heating curve


camelpode:
-->

### DESCRIPTION
I have found two new entities for the heating curve.
For Buderus WSW196i, with BC400

1. Comfort point Offset
2. Comfort point Temperature

These are used to offset the flow temperature at a certain outside temperature, when the hc control mode is set to "outside basepoint"

### REQUESTED INFORMATION

_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [X ] Searched the issue in [issues](https://github.com/emsesp/EMS-ESP32/issues)
- [ X] Searched the issue in [discussions](https://github.com/emsesp/EMS-ESP32/discussions)
- [ X] Searched the issue in the [docs](https://emsesp.org/Troubleshooting/)
- [ ?] Searched the issue in the [chat](https://discord.gg/3J3GgnzpyT)
- [ ?] Provide the System information in the area below, taken from `http://<IP>/api/system`

```
Paste System information here....

Buderus WSW196i, with BC400

Circuit HC1: Rx: 10 0B FF 00 01 AF 00 00 00 00 2D 23 00 00 16 14 00 05 01 00 D2
Circuit HC2: Rx: 10 0B FF 00 01 B0 00 00 00 00 3C 24 00 00 16 0F 00 03 02 00 B8
1. Comfort point Offset (Position 11)
2. Comfort point Temperature (Position 12) 

EMSGateway:$ read 0x10 0x02AF
2026-01-27T22:04:36.818 N 24: [emsesp] thermostat(0x10) W me(0x0B), RC300Summer(0x02AF), data: 00 00 00 00 2D 23 00 00 16 14 00 05 01 00


EMSGateway:$ read 0x10 0x02B0
2026-01-27T22:04:59.990 N 25: [emsesp] thermostat(0x10) W me(0x0B), RC300Summer(0x02B0), data: 00 00 00 00 3C 25 00 00 16 0F 00 03 02 00



Example:
Circuit HC1: Rx: 10 0B FF 00 01 AF 00 00 00 00 2D 23 00 00 16 14 00 05 01 00 D2
At temperature 1°C, the flowtemperature is offset +5K

Circuit HC2: Rx: 10 0B FF 00 01 B0 00 00 00 00 3C 24 00 00 16 0F 00 03 02 00 B8
At temperature 2°C, the flowtemperature is offset +3K

[...]

Issue #2933, opened 2026-01-22
Reset heating error is not working


hafo-135:
I'm running 3.8.2-dev0 on Plattform: ESP32S3-16MB+  PSRAM with an Bosch SupraEco.
My heating is showing an error message on the display. I tried to send command Reset -> Error and Reset -> Message but it is not working. Error is still shown. How to reset an error by EMS-ESP? 
In general sending commands to boiler is working  e.g. change temp setting

According to the log everything seems to be fine.
2026-01-22T21:55:04.261 T 1244: [telegram] Rx: 90 48 F7 06 2C
2026-01-22T21:55:04.274 I 1245: [command] Called command boiler/reset (Reset) with value Fehler
2026-01-22T21:55:04.283 T 1246: [telegram] Rx: C8 90 FF 05 02 01 F6 A8
2026-01-22T21:55:04.310 T 1247: [telegram] Rx: 90 48 FF 05 01 F6 08

[...]

Issue #2931, opened 2026-01-20
Timing issue on own customized entities


tp1de:
I am using latest dev version 3.8.2-dev.0.

I added an own customized entity type bool for telegram 2CC offset 3 (hc1 dhw priority) for my RC310 thermostat.
This works so far as desired. Just the timing of changes is not nice:

Within the RC310 display the value change is updated within 1-2 seconds.

When changing the value within ems-esp web-ui it takes up to 60 seconds until the value change is seen on the  dashboard. Same when using home assistant (mqtt). I believe that while sending the command to change, the old values is still persisitant, until next telegrams are read periodically.

Using standard entities this is different. Any changes are displayed within seconds (mqtt too).
This should be implemented while using own customized entities as well.

proddy:
we could change the `send_write_request` calls in `WebCustomEntityService::command_setvalue()` with the same `type_id` in the `validate_typid` which will do a post-fetch. The only side effect is that some entities use different telegrams for reading the value. For example, setting a boiler's flow temp, we write to 0x1A (or 0x2E0), and to validate the change, we read from 0x18 (or 0xE4). But in most cases it's harmless.

proddy:
what do you think @MichaelDvP  ?

MichaelDvP:
> what do you think @MichaelDvP ?

Yes, we can add the validation. For custom it's always write and read on the same type/offset, so no problem. EMS+ mostly broadcast on change, EMS not. The mentioned 2CC telegram is a bit special, It's only send from thermostat to boiler, so it's a boiler telegram (and implemented in boiler), but also stored in thermostat. The series is 0x2CD, 2CE, ..  as mixer telegrams for hc1.., and boiler 0x2CC for hc1 unmixed circuits, shorter telegram, but manly same offsets. 
There was a lot of confusion about this telegram in the past, i hope i understand it right now.

proddy:
in 3.8.2-dev.2

PR #2939, opened 2026-01-30
fixes and additions: #2918, #2931, #2933, #2935, #2936

MichaelDvP:


interface/src/app/status/SystemLog.tsx
```diff
@@ -308,6 +308,8 @@ const SystemLog = () => {
                 <MenuItem value={50}>50</MenuItem>
                 <MenuItem value={75}>75</MenuItem>
                 <MenuItem value={100}>100</MenuItem>
+                <MenuItem value={500}>500</MenuItem>
+                <MenuItem value={1000}>1000</MenuItem>
```
proddy, line 306: would be better only show 500 and 1000 if `psram` is true

proddy: Also the logic around `MAX_LOG_ENTRIES` in this file could also be removed

proddy:
I see the unit tests failed, I can re-generate that if you'd like me to

interface/src/app/status/SystemLog.tsx
MichaelDvP: The selectbox is only shown for psram true, without psram buffer is fixed to 25.

agree, `MAX_LOG_ENTRIES` only saves browser memory, a few hundred kB should not matter today, that's why i set it to very high. We can also remove it.

MichaelDvP:
test 2 fails because of  burnminpower now missing, the expected testdata should be adapted.

The others are missing `haclimate` if mode is not set. The expected value have mode-type set, but no mode.  I think we never have a thermostat with mode not set at all, so the test-telegrams are not generated as in real life. But i will re-check. 
I've added this mode-check to make sure the climate entity is only created after we have received telegrams for temperature and mode.

Checked, the Junkers have mode only in Set-telegram (0x165, 0x179), and the test sends only the monitor (0x16F). 
How to proceed? add a set telegram in test, than we have more data in the response. Or remove the `haclimate`  from expected response? Test 24/25 needs to be adapted in expected number of entities from thermostat.

MichaelDvP:
I've adapted the expected test reslts.

interface/src/app/status/SystemLog.tsx
proddy: ah yes, I missed that change. thanks!

Issue #2784, opened 2025-12-01
Customised EMS device model and manufacturer


proddy:
From discord:

"I can customize the name of bus devices, but this seems to be the "Model" only. What about allowing to separately customize "Model" or "Manufacturer"?
Background is that for my mixer and solar devices both model and manufacturer are properly populated by the manufacturer, but for my boiler and thermostat the manufacturer name is missing completely (and the model has just a dummy name, but this I can customize already now). This looks ugly in tables where both properties are shown in distinct columns"

Feature to add a custom name for an EMS device model and manufacturer.

proddy:
in 3.8.2-dev.4

PR #2953, opened 2026-02-12
 customize device brand #2784 

MichaelDvP:
also fix a possible memory leak when a device is redetected.


proddy (approved):
looks good. Another great feature added!

mattreim:
Hi everyone, during the first test I was able to change the manufacturer name and it is also displayed in EMS-ESP, but in HA it looks like this.

Boiler:
Kessel

Thermostat:
Thermostat

MichaelDvP:
@mattreim  Thanks for reporting, fixed in dev5

PR #2954, opened 2026-02-12
fix brand in HA

MichaelDvP:


proddy:
i tested in HA and it works correctly now.

mattreim:
It works, thank you very much.

PR #2986, opened 2026-03-17
update Core3, change TLS library

MichaelDvP:
updated to latest core 2 changes.

Since i have no luck connecting to mqtt://test.mosquitto.org:8883 with encryption, i've changed to a different TLS library. This works for me. But if you get encrypted mqtt running, we can also stay on the other lib.

Strangly i only cn connect mqtt encrypted after reboot, not on a paramter change. But same behaviour with core2 release builds.

I've also added a sendmail command. Settings in App-settings with some defaults to send just a body with `call system/sendmail bodystring` but you can also use a json as bodystring with recipient, topic, body, etc. overwriting the defaults.


proddy:
wow, great work @MichaelDvP . I love the sendmail feature and good find on the ESP_SSLClient library, Thank you for keeping this Core3 branch alive and up to date with the dev branch. I really want to make the switch on the next major version and thinking off an offline migration tool to get around the LittleFS problem

MichaelDvP:
I like the core3 with tasmota core. Running fine and have more free memory. No issues except the first installation wth bootloader and littleFS. My working (heating connected) ems-esp is on this build since weeks.

proddy:
> I like the core3 with tasmota core. Running fine and have more free memory. No issues except the first installation wth bootloader and littleFS. My working (heating connected) ems-esp is on this build since weeks.

same, I've been running it too the last weeks. I'll make some small changes and then we can discuss in the GitHub issue the next steps.

Issue #2991, opened 2026-03-19
Fehlende Intitäten


Martl72:
Ich betreibe zwei Buderus Frischwasserstationen bei einer werden alle daten angezeigt bei der zweiten nur wenig. Ich 
füge Bilder hinzu und bin für jede Hilfe dankbar. Bin technisch nicht so versiert.

<img width="1700" height="277" alt="Image" src="https://github.com/user-attachments/assets/5fbc9b2f-5255-49f3-9802-a22e952a12d2" />

<img width="1514" height="441" alt="Image" src="https://github.com/user-attachments/assets/893abf50-ed07-4ddf-88d1-a701c522599d" />

<img width="1499" height="782" alt="Image" src="https://github.com/user-attachments/assets/389f5f15-d2da-4a0d-ab72-edf198060c40" />

[log.txt](https://github.com/user-attachments/files/26122741/log.txt)

Martl72:
Beide stationen sind miteinander verbunden und im display SC300 werden alle daten angezeigt 

<img width="1137" height="532" alt="Image" src="https://github.com/user-attachments/assets/09a4771d-1b0e-4a1e-b672-44723d88c7f0" />

MichaelDvP:
In english please.

Acually ems-esp does not support dhw4 circuit, we simply don't know what telegrams are used for it (for dhw3 see #1387). The short log helps a bit, i can identify the temperature monitor on 0x7D7 and Status on 0x7E1. These telegram-ids follow the dhw3 telegrams (dhw3: 0x7D6, dhw4: 0x7D7), but for other telegrams this is not possible, eg. dhw3 uses 0x7A5 and 0x7A6. 

Please add the system info (see help page of ems-esp) and add a longer log of >5 minutes. Please use normal logging that is easier to read than hex logging. Maybe change some values on the thermostat while logging to see the command telegrams

Seems you have a lot of errors on the bus, rx and tx. Check the connections, try a different connection point.
Is there a reason you have changed the default bus-address to terminal (0x0A)?

Martl72:
Hello
First of all, thank you for your help. Unfortunately, I'm not technically savvy enough to know what the numbers mean. I've logged the data again when I turn on the water and tried to create different scenarios. I'm only really interested in the flow rate (l/min) for both the flow and the water supply. I only changed the bus address for testing purposes; there's no particular reason for it. If you need anything else, I'll try to provide it. Thank you very much.

[...]

PR #2994, opened 2026-03-21
dev.12, add dhw4, 5,.. circuits, #2991

MichaelDvP:


proddy:
@MichaelDvP could you also add to Core3 so we keep both branches in sync, I know its double work. until we've worked out a method to migrate people to v3.9.

Issue #3002, opened 2026-03-29
Full system backup and restore


proddy:
Add a System Backup button to the webUI Download/Upload page which exports the whole system (all settings files, NVS stored values) to a single JSON file so a system can be restored.

This is needed when we move to 3.9 and switch to Arduino Core v3 which uses a different filesystem

proddy:
If you want to pre-load via pio, copy the backup json file to the  `/data/pre_load.json` file and run ` pio run -t buildfs -t uploadfs`

PR #3006, opened 2026-03-30
sync with core3 features

proddy:
- show firmware size instead of partition size in Version page
- removed flto in build, upgraded to latest libraries in platformio.ini
- optimized for vite v8
- renamed common.h to emsesp_common.h to avoid build conflicts with IDF
- full system backup and restore (https://github.com/emsesp/EMS-ESP32/issues/3002)
- some minor changes in some web screens (dont show failed EMS bus on Dashboard etc)
- Modules have been disabled


MichaelDvP (approved):
Looks ok. Some C6 checks (e.g. uart) are useless, as core 2 does not support C6. But does not hurt.
Testing the backup i see that the warning about settings files with passwords are gone. Intended?

proddy:
> Looks ok. Some C6 checks (e.g. uart) are useless, as core 2 does not support C6. But does not hurt. Testing the backup i see that the warning about settings files with passwords are gone. Intended?

I can't remember why, but I probably had a good reason!

I'm thinking of removing the 'Application Settings', 'Customizations', 'Custom Entities' buttons entirely and just have one export for the full system backup (including Scheduler). What do you think?

MichaelDvP:
Agree, only system backup is better.

proddy:
I'm making some updates - this PR is still work in progress

proddy:
Think it kinda works. Can you take a look @MichaelDvP when you have time?

This is the warning for people upgrading from 3.8 to 3.9 (switching to Tasmota core)

PR #3009, opened 2026-04-04
https client for scheduler/shuntingYard

MichaelDvP:
- move http/https clients to heap(psram) to reduce stack usage.
- https client uses ESP_SSLClient, actually without certificate-check.
- Tested POST with pushover .
- i have not tested GET


proddy:
I've been away the last week, so will check this weekend and merge with the other PR

proddy:
can you add to core3 too?

MichaelDvP:
> can you add to core3 too?

This is core3 only, core2 with psram has https in http class.

proddy:
> > can you add to core3 too?
> 
> This is core3 only, core2 with psram has https in http class.

yes, sorry - I missed that

src/web/WebSchedulerService.cpp
```diff
@@ -363,39 +363,139 @@ bool WebSchedulerService::command(const char * name, const std::string & command
...
+                            http->addHeader(asyncsrv::T_Content_Type, asyncsrv::T_application_json, false); // auto-set to JSON
+                        } else {
+                            http->addHeader(asyncsrv::T_Content_Type, asyncsrv::T_text_plain, false); // auto-set to JSON
+                        }
+                    }
+                    httpResult = http->POST(value.c_str());
+                } else {
+                    httpResult = http->GET(); // normal GET
+                    if (httpResult > 0) {
+                        result = http->getString().c_str();
```
proddy, line 486: result is never used - is this intentional? @MichaelDvP

MichaelDvP: Actually: yes. We can skip or insert a LOG_DEBUG to use it. 
The http in SchedulerService is for http commands POST and sometimes GET like shelly(get): http://<shellyIP>/relais/0?turn=on
We don't need the result text. Maybe later...

PR #3020, opened 2026-04-15
sync with dev

proddy:


MichaelDvP:
Is it intended that you roll back a lot of core 3 changes? Also add back the core3/core2 #ifdef.

proddy:
> Is it intended that you roll back a lot of core 3 changes? Also add back the core3/core2 #ifdef.

no, mistake. I just cut & paste a few .cpp files without checking. I'll fix that

PR #3025, opened 2026-04-17
Core3

MichaelDvP:
I've merged the latest core2 changes to core3 and fixed the merge errors. Now the branches are in sync and we can automatic merge core2 changes direct again. 
Also added back the email-settings and formatted WebSettingsService.


proddy:
that's amazing! And quite some manual work, thanks Michael. It was hard keeping both branches in sync

PR #3028, opened 2026-04-18
heap memory optimizations

proddy:
Various changes to improve memory utilization. This also fixes https://github.com/emsesp/EMS-ESP32/discussions/3021, where an 11KB drop was noticed between dev-12 and dev-13. With these changes that issue has been addressed, and the bonus is an additional ~10KB free'd up without the loss of performance.

- added back LTO, to optimize flash usage, saving 5KB.
- Improve web page chunking in Vite, so instead of having 58 handlers, there's one. In vite 8 the rolldown-native codeSplitting groups worked differently to v7. This is the biggest addition, saving close to 13KB. Also, move all etags to flash. See vite reported bug in https://github.com/vitejs/vite/issues/ 22007
- replace SemVer with our own lightweight version
- When looking at partitions, only call esp_image_verify() and store the entry for partitions that actually have content


proddy:
this is quite some changes @MichaelDvP , it works on my system - can you check on yours?

when merged, I'll apply the same fixes to core3

MichaelDvP:
> this is quite some changes @MichaelDvP , it works on my system - can you check on yours?

Yes, works with much better ram.
```
    "version": "3.8.2-dev.12",
    "freeMem": 142,
    "maxAlloc": 103,
    "freeCaps": 4190,
    "usedApp": 2382,
    "freeApp": 2290,

    "version": "3.8.2-dev.13",
    "freeMem": 132,
    "maxAlloc": 95,
    "freeCaps": 4180,
    "usedApp": 2600,
    "freeApp": 2072,

    "version": "3.8.2-dev.16",
    "freeMem": 152,
    "maxAlloc": 107,
    "freeCaps": 4200,
    "usedApp": 2395,
    "freeApp": 2277,
```
> when merged, I'll apply the same fixes to core3

Should work with merge dev into core3.

proddy:
> Should work with merge dev into core3

there are still some nuances in core3, like the new loop() for NTP checking, so I'm still doing things manually to be sure.

PR #3029, opened 2026-04-18
sync memory optimizations from dev-16 to core3

proddy:
These are all the changes from https://github.com/emsesp/EMS-ESP32/pull/3028 to optimize the heap memory usage.

It also has a fix for the DNS name that was dominated by 'tasmota' and not using EMS-ESP's hostname, for Eth and WiFi.


proddy:
I noticed NTP is not working with core3, so will look into that

MichaelDvP:
When checking i've found differences:
in system.cpp
https://github.com/emsesp/EMS-ESP32/blob/d6d3a034ad7f0d4ea06dbc10ae22b78577c349ea/src/core/system.cpp#L1770
and
https://github.com/emsesp/EMS-ESP32/blob/c40d8287494b02f7ffaa5030754c356e9214e473/src/core/system.cpp#L1854
i think the core3 variant with nvs1 checking should also be in dev.

And in platformio.ini core3 uses `core_build_flags = -std=gnu++20`, dev uses `core_build_flags = -std=gnu++2a`
Makefile in both branches set gnu++20
Any reason for this?

proddy:
> And in platformio.ini core3 uses core_build_flags = -std=gnu++20, dev uses core_build_flags = -std=gnu++2a
> Makefile in both branches set gnu++20
> Any reason for this?

see https://github.com/emsesp/EMS-ESP32/pull/3028/changes/77b8b21aea43973e38c9706c2f29acb68e95ebb9

proddy:
> i think the core3 variant with nvs1 checking should also be in dev.

yes, left over bug. I'll fix. thanks for spotting

proddy:
@MichaelDvP I was playing around with OLED (see discussion) so now have time to fix some of these issues. Let me know if which one you want me to pick up as I see you've started fixing a few (like the version check).

Issue #2866, opened 2025-12-31
Feature: Holiday mode


TheMagnetar:
### DESCRIPTION

As suggested in #2467. Implement writing and reading start and end day vacation. There is already a rough working proof-of-concept using custom entities:

```
name : device_id : type_id : offset : writable : size : scale : unit
Holiday_Start_Year : 0x10 ; 0x269 : 0 : yes : uint8 : 1 : -
Holiday_Start_Month : 0x10 ; 0x269 : 1 : yes : uint8 : 1 : -
Holiday_Start_Day : 0x10 ; 0x269 : 2 : yes : uint8 : 1 : -
Holiday_End_Year : 0x10 ; 0x269 : 3 : yes : uint8 : 1 : -
Holiday_End_Month : 0x10 ; 0x269 : 4 : yes : uint8 : 1 : -
Holiday_End_Day : 0x10 ; 0x269 : 5 : yes : uint8 : 1 : -
Holiday_unknown : 0x10 ; 0x269 : 6 : yes : uint8 : 1 : -
Holiday_Temperature : 0x10 ; 0x269 : 7 : yes : uint8 : 1 : °C
```

### REQUESTED INFORMATION

_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [x] Searched the issue in [issues](https://github.com/emsesp/EMS-ESP32/issues)
- [x] Searched the issue in [discussions](https://github.com/emsesp/EMS-ESP32/discussions)
- [x] Searched the issue in the [docs](https://emsesp.org/Troubleshooting/)
- [x] Searched the issue in the [chat](https://discord.gg/3J3GgnzpyT)
- [x] Provide the System information in the area below, taken from `http://<IP>/api/system`

### TO REPRODUCE

None. New feature

### EXPECTED BEHAVIOUR

_A clear and concise description of what you expected to happen._

### SCREENSHOTS

_If applicable, add screenshots to help explain your issue._

### ADDITIONAL CONTEXT

_Add any other context about the issue here._

**(Please remember to close the issue when it has been addressed)**

tp1de:
I personally think that it does not make a lot of sense to implement the holidaModes and switchPrograms within ems-esp.
It is much easier to use ha automations rather than EMS+ thermostat logic.

With VPN access to their home it is easy to control while using HA (or ioBroker, Loxone ...) or by ems-esp dashboard.

VlastiBroucek:
[...]

PR #3033, opened 2026-04-21
Fix RC300Holiday telegram length for inclusion of HC/DHW mode & assignment

bimbalasim:
**Problem:** The `0x269` holiday telegram is registered with fetch size `6`, covering only the date bytes (offsets 0–5). Custom entities targeting HC/DHW assignment bytes (offsets 9–13, documented in #2866) cannot be read because the custom entity `fetch()` logic defers to the device driver's fetch when `is_fetch(0x269)` is true — meaning only 6 bytes are ever polled. The DHW byte at offset 13 is never received, see #2467 (@drheck).

**Fix:** Increase fetch size from `6` to `14` so the full HC/DHW assignment block is included in the polled response. The `process_RC300Holiday` handler is unaffected — it still only processes the first 6 bytes.

### DIFF
--- a/src/devices/thermostat.cpp
+++ b/src/devices/thermostat.cpp
@@ -202,7 +202,7 @@ Thermostat::Thermostat(uint8_t device_type, uint8_t device_id, uint8_t product_i
         if (model == EMSdevice::EMS_DEVICE_FLAG_RC100) {
             register_telegram_type(0x43F, "CRHolidays", true, MAKE_PF_CB(process_RC300Holiday), 6);
         } else {
-            register_telegram_type(0x269, "RC300Holiday", true, MAKE_PF_CB(process_RC300Holiday), 6);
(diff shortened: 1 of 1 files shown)


MichaelDvP (approved):
This is ok, but for future we need a better solution.
When setting the fetch length i 've mostly used the length of the telegram if i found an example in the comments, If not, i used the highest read value. But the bosch devices are changing the different rc300, bc400,, UI800, R3000 all use the same telegram, but with different length. I hink one variant have up to 19 bytes.

I'll make a check to custom entities and compare the fetch.length to make a an extra fetch if out of range.

MichaelDvP:
@bimbalasim I've added a PR to check the fetched length for custom entities and fetch the value if out of length. Also set the fetch length for this telegram to 18, so it's only one fetch. Are you ok with closing this PR and use my modification?

bimbalasim:
@MichaelDvP Yes, of course -- thanks a lot for taking care of this issue!

MichaelDvP:
Should work in dev.17

PR #3036, opened 2026-04-22
small fixes

MichaelDvP:
- include change from pr #3033 and add length check for custom entities (PR #3033 is obsolet now)
- revert #3021, it does not change anything,same code size,sae memory
- fix version check to show the warning messages, tested


proddy:
the revert is the only the JWT code, you got me worried their for a minute!

MichaelDvP:
Sorry for confusion. Should be revert to Commit c54da18

PR #3037, opened 2026-04-22
Core3 update

MichaelDvP:
I first added in system.cpp line 428 update the otadata partiton to core3 format. This works after update to core3. But if someone switches app0/app1 partitons by terminal or install saved version, it will not work. As the littleFS always fails on switching to core 3, i now moved the otadata update to factory settings. 
I don't want to do this on every boot. I'm not sure if writing is executed also if there is no change and would wear out the otadata partition with too much writes.
The 4M tasmota updates without any issue, littleFS and otadata works.

I've merged dev into core 3 again to have both branches in sync. Adds a lot of commits, but no code, as you have already added manual all changes. The only new are the github-action updates from dev now merged.


proddy:
thanks for keeping core3 in sync.

> if someone switches app0/app1 partitons by terminal or install saved version, it will not work. As the littleFS always fails on switching to core 3, i now moved the otadata update to factory settings.

We could check the upgrade in `System::check_upgrade()` when going from 3.x to 3.9 and then force remove all the other partitions? This will prevent people from switching.

I don't want to do this on every boot. I'm not sure if writing is executed also if there is no change and would wear out the otadata partition with too much writes.
The 4M tasmota updates without any issue, littleFS and otadata works.

> I've merged dev into core 3 again to have both branches in sync. Adds a lot of commits, but no code, as you have already added manual all changes.

I'm still manually looking at every change in dev and then manually adding to core3, because I'm paranoid. But as you've seen I still make mistakes. Are you auto-merging via GitHub? If so, how so I can learn that trick!

We also need to add the latest DE translations from dev to core3 (https://github.com/emsesp/EMS-ESP32/pull/3035/changes). Let me know if you want me to do that.

MichaelDvP:
> Are you auto-merging via GitHub? If so, how so I can learn that trick!

Yes, i localy use git-GUI to merge and it shows all merge conflicts. These conflicts can be solved in VSC and before commit i can see in the git-GUI the all changes. 
Normally auto-merge works good, core3 is ahead of dev and these ahead commits stay, only the new dev changes are merged and mostly does not conflict. 
When adding changes to core3 not exactly the same as in dev, i have to look for double code after merge.

proddy:
> > Are you auto-merging via GitHub? If so, how so I can learn that trick!
> 
> Yes, i localy use git-GUI to merge and it shows all merge conflicts. These conflicts can be solved in VSC and before commit i can see in the git-GUI the all changes. Normally auto-merge works good, core3 is ahead of dev and these ahead commits stay, only the new dev changes are merged and mostly does not conflict. When adding changes to core3 not exactly the same as in dev, i have to look for double code after merge.

thanks Michael, that was really helpful.

Issue #3047, opened 2026-04-27
updated version check


proddy:
I made some changes to how versions are handled
- versions for dev and stable are stored in Cloudflare cache under emsesp.org
- when a new build is generated the version details are updated (has the version name and release date)
- dates are in UTC format
- cache is 1 minute, to keep within Cloudflare query limits
- to request the Cloudflare KV, it's proxied via an endpoint https://emsesp.org/versions.json
- the C++ code does periodic checks to fetch the version details, on boot and then every hour afterwards
- a `upgradeable` has been added to the system heartbeat MQTT topic. Users can poll/scan this.
- the emsesp.org docs uses the same endpoint, so it will always show the latest versions, reducing the need to republish the site

to-do:
- highlight more prominently that there is an update available in the WebUI (now it's only on the Version page)

_Originally posted by @proddy in https://github.com/emsesp/EMS-ESP32/discussions/3044#discussioncomment-16719137_

proddy:
Cloudflare adjusted to accept http:: for that single endpoint, so it works on ESP32's without https/ssl. Confirmed working on a 16MB S32 gateway board. Also tested with `curl -I http://emsesp.org/versions.json`

proddy:
Highlighting the changes (in red boxes), showing the new yellow badge

<img width="884" height="622" alt="Image" src="https://github.com/user-attachments/assets/4f5935ca-0da7-4b49-a27a-9eccd2519b0a" />

<img width="884" height="622" alt="Image" src="https://github.com/user-attachments/assets/09a1727a-6b26-4535-8dfa-213ab0450720" />

PR #3048, opened 2026-04-27
update auto version checking and optimized web

proddy:
- updated how automatic versioning is handled, see https://github.com/emsesp/EMS-ESP32/issues/3047
- removed unnecessary useMemo and useCallback from web code - speeds things up a little


proddy:
@MichaelDvP there's a lot of changes and hard to review, but if you have time I would only look at the changes in `WebStatusService.*`. It has some new functions and a loop() that is called from the emsesp main loop.

ignore the WebUI changes, it's just some cleaning.

it's going to be nightmare to merge with core3, so I'll think I'll do it manually.

MichaelDvP:
Puh, really a lot of changes. I'll focus on the version functions.
> it's going to be nightmare to merge with core3, so I'll think I'll do it manually.

As far as i see there is no change in core3 modified functions, so it auto-merges with only very little conflicts: version number and package update. Try it. (I made a new branch before testing automerge,
grafik

src/web/WebStatusService.cpp
```diff
@@ -311,46 +310,164 @@ uint8_t WebStatusService::upgradeImportantMessages(std::string & version) {
...
+        return false;
     }
 
-    root["emsesp_version"] = current_version_s; // always send back current version
+    JsonDocument         doc;
+    DeserializationError err = deserializeJson(doc, http.getStream());
+    http.end();
+    if (err) {
+#if defined(EMSESP_DEBUG)
+        EMSESP::logger().debug("versions.json: parse error (%s)", err.c_str());
```
MichaelDvP: DeserializationError is an enum, not a string

proddy: thanks, fixed

proddy:
> Puh, really a lot of changes. I'll focus on the version functions.
> 
> > it's going to be nightmare to merge with core3, so I'll think I'll do it manually.
> 
> As far as i see there is no change in core3 modified functions, so it auto-merges with only very little conflicts: version number and package update. Try it. (I made a new branch before testing automerge, grafik

yes, sorry, a lot of changes. The web is just some tweaking of useMem/useContext so no point checking that. The main changes are done in the `webStatusService.cpp` which you looked at, and also the github actions. There's also some code for Cloudflare workers and pages which you don't get to see since it's hosted on my private account for emsesp.org ;-)

when you've tested and happy with the PR, I'll merge and then do your trick to apply to core3.

src/ESP32React/NetworkSettingsService.cpp
```diff
@@ -321,6 +321,7 @@ void NetworkSettingsService::WiFiEvent(WiFiEvent_t event, WiFiEventInfo_t info)
                                       WiFi.localIP().toString().c_str(),
                                       WiFi.getHostname(),
                                       emsesp::Helpers::render_value(result, ((double)(WiFi.getTxPower()) / 4), 1));
+        emsesp::EMSESP::webStatusService.schedule_versions_refresh(); // run the version fetch as soon as the main loop picks it up
```
MichaelDvP, line 324: These calls have to be move to the loop inn core3. 
But it seems a bit overcomplicated to call this on connect. The webStatusLoop check the connect status and gets the versions onlyif connected, so you can set the values as dfault and wait for connect.

proddy: good point, I had it there before because it would take ~10 seconds to get the version data, so when you're in thr WebUI looking at the Version page it would tell you "you need an internet connection". It was an old race condition that probably is fixed now, I'll double check

MichaelDvP: Misunderstanding: It works in dev. 

I only wonder why you trigger the first check in the networkservice callback on connect. You already check the connect state in webStatusLoop, so it is double. Imho you can remove the "schedule_versions_refresh()" function and the versions_next_fetch_ms_ == 0 check and it will work also.

For core3 we have no working network-callback and pull the connection states in the netowrk loop. So the schedule_version_refresh call has to be moved too (or do all in the webStatusLoop). I forgot this when merging to core3 and always get the "you need an internet connection" message.

proddy: I suggest then to make it compatible with both dev and core3, move all calls to `refresh_versions_cache()` to only `WebStatusService::loop()` and have that loop check if the Eth/WiFi is re-connected and force the refresh pulling the fetch from http://emsesp.org ?

Maybe also change the Web message "Internet connection required for automatic version checking and upgrading" to something nicer, like "Unable to fetch latest versions online", or maybe just remove it.

MichaelDvP: Tested a bit and now understood the intension of this construction. The connect state is set before we get a IP and can contact the server. 
For all in `WebStatusService::loop()` we have to change the check to `if (!ETH.hasIP() && !WiFi.localIP() && !ETH.hasGlobalIPv6() && !WiFi.globalIPv6()) {` and maybe reset the `versions_next_fetch_ms_` if we want to fetch direct after reconnect, but i think once a hour is good, also if connection is intermediate lost.

proddy: ok, I'll implement your suggestion, move the check to the main loop, and will try it out.

proddy: @MichaelDvP I've stripped out all the network-related code and centralised this under a new src/core/network.cpp file. It didn't like it all scattered around the ESP32React folder in NetworkService/APService files. It makes the code tidier and all the network handling is done in one place. It is however a big ccde change so sorry again! I'm currently running the final tests in core3. If it works I'll port it to 3.8.2/dev

proddy:
merged into core3, we'll keep dev clean for the upcoming new stable release

PR #3052, opened 2026-05-01
refactor networking into a single class

proddy:
- new cleaner network code, removed all logic from ESP32React() Settings and Status classes
- merged with the improved version update check from dev, using cloudflare KV instead of GitHub API to fetch versions
- removed useMemo and useCallback, not needed for single user access and reduces code size and increases performance on the WebUI
- some other changes I've been making. PR reviews are slow which is why these are getting bigger. I need to find a betterr strategy for rolling up my updates

tested with different logic
- no wifi or ethernet, should start an AP
- has an ethernet port, but no cable or picks up no IP from router, fallback to wifi
- on wifi and the connection breaks (no IP, bad password etc), switch to AP
- on AP and Ethernet is working, use that
- on AP, no ethernet, but WiFi is back up, so switch back


MichaelDvP:
When using a plain new ems-esp the log is flooded (every3 sec) with `Network connected via AP (IP: 192.168.4.1)`
Connecting to AP works(without any log message), Wifi-Scan and select SSID also. But after entering all network settings the network is not connected.
Also after disconnecting the AP client it does not change. I have to call restart from serial console. After restart the network is ok.
But before imo it was: Entering the wifi settings connects immediate and we can see the connect state in status, after disconnecting the AP client the AP stops and only the wifi network stays on.

proddy:
Thanks for testing. Looks like I missed some cases, I'll fix those. One of the designs was to remove the combination of having both softAP and Sta active as I didn't see it necessary. The captive portal was only to get first users onto a WiFi network.

MichaelDvP:
> One of the designs was to remove the combination of having both softAP and Sta active as I didn't see it necessary.

Ok, but after setting the station params we need a reboot, this should be triggert in NetworkSettiingsSetrvice if ssid changes from empty to set..

Before we checks for WiFi.connect() and with a typo in password it stays in reconnecting to station and never fall back to AP-mode, that is fixed now with check for station-IP.

proddy:
yes, just fixed that and now testing. It shouldn;t need a reboot, just a network reconnect call. will do a PR shortly. 

Another question: in the old design on E32s it would switch to Ethernet if the WiFi SSID is blank. I'm thinking of removing that specific check so it will always connect to the best network Eth -> WiFi -> AP -> Error. What do you think?

MichaelDvP:
> I'm thinking of removing that specific check so it will always connect to the best network Eth -> WiFi -> AP -> Error. What do you think?

This was mainly to skip the eth initialization and save memory if a SSID is set. Check the memory of 4M versions when doing this.

proddy:
> > I'm thinking of removing that specific check so it will always connect to the best network Eth -> WiFi -> AP -> Error. What do you think?
> 
> This was mainly to skip the eth initialization and save memory if a SSID is set. Check the memory of 4M versions when doing this.

oh, good point. I forgot about that. I'll leave it as is.

PR #3059, opened 2026-05-04
fix warning "esp32-hal-gpio.c:181] __digitalWrite(): IO 2 is not set …

proddy:
When EMS-ESP boots on a E32V one of the first visible errors in the log is

`esp32-hal-gpio.c:181] __digitalWrite(): IO 2 is not set as GPIO. Execute digitalMode(2, OUTPUT) first`

### DIFF
--- a/src/core/system.cpp
+++ b/src/core/system.cpp
@@ -840,11 +840,6 @@ void System::button_init() {
 
 // set the LED to on or off when in normal operating mode
 void System::led_init() {
-    // RGB: drive strip low before init. Standard GPIO LED: pinMode must precede digitalWrite (ESP32 Arduino hal).
-    if (led_type_) {
-        EMSESP_RGB_WRITE(led_gpio_, 0, 0, 0);
-    }
-
     if (!led_gpio_) { // 0 means disabled
         LOG_INFO("LED disabled");
         return;
--- a/src/core/system.cpp
+++ b/src/core/system.cpp
@@ -840,19 +840,21 @@ void System::button_init() {
 
 // set the LED to on or off when in normal operating mode
 void System::led_init() {
-    // disabled old led port before setting new one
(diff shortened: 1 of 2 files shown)


src/core/system.cpp
```diff
@@ -840,19 +840,21 @@ void System::button_init() {
 
 // set the LED to on or off when in normal operating mode
 void System::led_init() {
-    // disabled old led port before setting new one
-    led_type_ ? EMSESP_RGB_WRITE(led_gpio_, 0, 0, 0) : digitalWrite(led_gpio_, !LED_ON);
+    // RGB: drive strip low before init. Standard GPIO LED: pinMode must precede digitalWrite (ESP32 Arduino hal).
+    if (led_type_) {
+        EMSESP_RGB_WRITE(led_gpio_, 0, 0, 0);
+    }
```
MichaelDvP: This is useless since the new gpio checks. 

Originally it disables the led, then reads settings from settingService and sets System::led_port_and initialize new led port.
Now the led_port is set in EMSESP::system_.store_settings(settings); and here we onlyhave the new one.

Intend was when changing settings from gpio2/blueLed to gpio32/rgbLed  make sure the old led is off.

proddy: I see the error everytime the E32V2 boots, do you see it too?

MichaelDvP: I have not seen it, but can imaging that digtalWrite without pinmode set before can throw this error. 
But we can skip this resetting here complete, it is useless to clear the new gpio and then initialize it to off (lines 845 and 854 do exactly the same). To clear the old led gpio we have to add a function to switch off before setting the new gpio.

proddy: that's ok, I'll revert the code as the warning is harmless. Note I get this on every boot (after a flash erase) only on a E32V2.1 not the latest E32 gateways

MichaelDvP: > I'll revert the code as the warning is harmless. 

You can merge as is, it's better than before. 
Just the lines 843-846 can also be removed, they always do nothing.

PR #3065, opened 2026-05-07
sendmail name rollback, added bin

proddy:
@MichaelDvP I've added a new bin folder. I've seen people asking for the esptool method to upload the firmware, so thought I'd just include it here. Let me know what you think,


proddy:
Actually thinking about it, it's not good practice to include the .bin files in the source control system. I'll remove this and add it to the docs (emsesp.org) as static files, and then remove the link to the https://github.com/emsesp/EMS-ESP-Flasher-CLI which isn't very handy

MichaelDvP:
Yes, i also think it's better in the docs. 
Maybe add a link where o find partition,bootloader, boot_app0 bin files and a link to https://docs.espressif.com/projects/esptool/en/latest/esp32/esptool/flashing-firmware.html

PR #3069, opened 2026-05-09
enable OTA uploads of filesystem (pre_load.json)

proddy:
add OTA upload for filesystem. So when there is a `data/pre_load.json` file (which is essentially a system backup renamed) you can

```
 pio run -t buildfs 
 pio run -t uploadfsota
```

`pio run -t uploadfs` still works over a COM port.

### DIFF
--- a/src/ESP32React/UploadFileService.cpp
+++ b/src/ESP32React/UploadFileService.cpp
@@ -4,6 +4,7 @@
 
 #include <esp_app_format.h>
 #include <esp_ota_ops.h>
+// #include <esp_partition.h>
 
 static String getFilenameExtension(const String & filename) {
     const auto pos = filename.lastIndexOf('.');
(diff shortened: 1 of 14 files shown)


src/ESP32React/UploadFileService.cpp
```diff
@@ -96,28 +104,55 @@ void UploadFileService::handleUpload(AsyncWebServerRequest * request, const Stri
...
+                return;
+            }
+            if (final) {
+                if (!Update.end(true)) {
+                    emsesp::EMSESP::logger().err("Update.end failed: %s", Update.errorString());
+                    handleError(request, 500);
+                    return;
+                }
+            }
+        } else {
```
MichaelDvP: this `else` belongs to `if (_is_firmware...`

proddy: thanks, fixed.

PR #3080, opened 2026-05-14
upgrade bus id from x0B to x49

proddy:
upgrade bus id from x0B to x49, unless set

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,4 @@ interface/.tsbuildinfo
 test/test_api/package-lock.json
 .clangd
 mklittlefs
+.cursor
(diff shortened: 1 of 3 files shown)


MichaelDvP:
Sadly this will work only for the 4M Tasmota systems. For S3, E32V2 the FS gets formated and set the default ID 0x49. But the user will upload the backup after this and this sets back to 0x0B again. Don't know how to solve this. Overwriting each backup is not a good solution.

proddy:
> Sadly this will work only for the 4M Tasmota systems. For S3, E32V2 the FS gets formated and set the default ID 0x49. But the user will upload the backup after this and this sets back to 0x0B again. Don't know how to solve this. Overwriting each backup is not a good solution.

ah, good point. It's a bit of a hack but we can add to "System::check_restore()" so when a backup file is being processed, we override the ems_bus_id (if set to 0B)

MichaelDvP:
Yes, i see that you stored the original version in backup file. So we can check and change the ID only if backup file is pre 3.9.0 with ID 0x0B.

proddy:
closing this, as it's merged with https://github.com/emsesp/EMS-ESP32/pull/3081

Issue #3063, opened 2026-05-06
Feature: Make RGB LED (preset colors) accessible via API/etc #3039


proddy:
(from the discussion in _Originally posted by @proddy in https://github.com/emsesp/EMS-ESP32/discussions/3039#discussioncomment-16688430_)

For for those who automate installation locally or remotely it could be a good indication if they can verify something, and then send a blink command to let the end-user know the configuration was successful.

proposal is to implement an `/api//system/led` command that takes an argument in the form `[color]:[style]` where style could be `blink_once', `blink_2`, `blink_3`, `solid`. After 3 seconds, the LED reverts back to its previous state.

For older non-RGB models, the colour would default to just being on.

Then it can be called via an REST HTTP command for remote access, or via the Scheduler.

proddy:
@MichaelDvP I started on this and refactored all the LED code into a new led.cpp LED Class. PR coming shortly

proddy:
Implemented

- patterns: `blink`, `blink2`, `blink3`, `rgb`, `cpc`
- colors: `red`, `green`, `blue`, `orange`, `yellow`, `cyan`, `pink`, `white` and `on` , `off`

Can be called via API like

```sh
curl -X POST \
    -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VybmFtZSI6ImFkbWluIiwiYWRtaW4iOnRydWV9.2bHpWya2C7Q12WjNUBD6_7N3RCD7CMl-EGhyQVzFdDg" \
    -H "Content-Type: application/json" \
    -d '{"data":"yellow:blink2"}' \
    http://ems-esp.local/api/system/led 
```

PR #3081, opened 2026-05-15
Feature led api #3063

proddy:
Implements https://github.com/emsesp/EMS-ESP32/issues/3063

plus some minor changes like
 - use uuid:getuptime instead of millis()
 - rename Wifi hostname to Network hostname
 - includes EMS bus ID 0x0B->0x49 (https://github.com/emsesp/EMS-ESP32/pull/3080)
 - upgrade packages


src/core/led.cpp
```diff
@@ -0,0 +1,287 @@
...
+    if (led_flash_step_
+        && (uint32_t)(current_time - led_short_timer_) >= (is_user_led_blink ? HEALTHCHECK_LED_FLASH_FAST_DURATION : HEALTHCHECK_LED_FLASH_DURATION)) {
+        led_long_timer_  = 0; // stop the long timer
+        led_short_timer_ = current_time;
+
+        if (++led_flash_step_ == 8) {
+            // finished first iteration, reset the whole sequence
+            led_long_timer_ = uuid::get_uptime();
+            led_flash_step_ = 0;
+            reset_led(true); // LED back to what is was before
```
MichaelDvP: should be `reset_led(healthcheck == 0)` otherwise in error state it starts with green and continues with red/blue blinks.

MichaelDvP:
Nice, works. You can also remove the led_gpio, hide_led, led_type in system, they are not used anymore.
The automatic change to bus_id 49 works also as expected on restore backup from older version.

proddy:
> Nice, works. You can also remove the led_gpio, hide_led, led_type in system, they are not used anymore. The automatic change to bus_id 49 works also as expected on restore backup from older version.

thanks for taking a look so quickly. I'll tidy up the old variables and also see if I can make it flash when there is a firmware uploading

src/core/led.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+    // if button is pressed, show LED (yellow on RGB LED, on/off on standard LED) and exit
+    if (last_button_busy_ != button_busy) {
+        last_button_busy_ = button_busy;
+        set_led(button_busy ? Color::OFF : Color::YELLOW); // Yellow
+    }
+
+    // we only need to run the LED sequence_led if there are errors, or if a button has been pressed or a user-requested LED blink is active
+    if ((!healthcheck || button_busy) && !is_user_led_blink) {
+        return; // nothing to show
```
MichaelDvP: You should do a `reset_led();` here to activate the green led again if healthcheck is zero. 
You can skip the ` reset_led()` call in system.cpp then, that is only called on change on health.

proddy:
thanks for the suggestions, I've implemented them and made some more minor improvements.

PR #3083, opened 2026-05-17
various optimizations

proddy:
I let Claude Code go through the code base and recommend some changes. It makes the code cleaner but doesn't improve on the heap memory, only suppresses some of the dips when processing large loads by using the stack more efficiently.

This was the first batch:
- replace `std::function` with lambda (in `emsdevice.h` and `emsesp.h`), uses less stack and less CPU, apparently
- add const everywhere where it should be, also enforces passing by reference instead of copying
- wrap all `LOG_*` functions with a condition to present processing the command and then deciding if we want to log it
- new `psram_json_allocator.h` so ArduinoJson can also use PSRAM, based on Michael's code
- new function `brand_to_cstr()` to return a fixed char *, preventing std::string allocation
- re-enabled the two functions `reserve_device_values` and `reserve_telegram_functions` to reserve memory
- moved all URI handlers to a single function (from 25 to 1) saving 2.5-3KB heap
- use PSRAM in AsyncJsonResponse()


src/core/emsdevice.cpp
```diff
@@ -105,8 +105,12 @@ const char * EMSdevice::uom_to_string(uint8_t uom) {
 }
 
 std::string EMSdevice::brand_to_char() {
+    return std::string{brand_to_cstr()};
+}
+
+const char * EMSdevice::brand_to_cstr() const {
     if (!custom_brand().empty()) {
-        return custom_brand();
+        return custom_brand().c_str();
```
MichaelDvP, line 113: I have not tested these changes yet, but iirc we have const char*  before adding the custom brand. returning custum_brand.c_str() and put this in a json does not work. json expects a const char * object as fixed address and does not make a copy. But the c_str() is deleted before the json is sent.
We get garbage in custom brand on the web page.

proddy: Ah good point! ArduinoJson doesn't copy char * strings. I'll revert...

proddy:
@MichaelDvP some more changes incoming, which will make you go crazy. Sorry. Best thing is to keep trying it out and see where it breaks. The main change is now in the SecurityManager.cpp which handles a single URI endpoint capturing all /rest and /api calls. My tests all run, but no doubt something else could be broken down stream.

proddy:
I'm merging this. I have been testing the last week and it seems stable enough.

PR #3093, opened 2026-05-24
network fallback, fixes #3090

### LINKED ISSUE #3090: [core3] Network reconnect issue
I had a router update last day, wifi was off for ~ 5minutes and my ems-esp does not reconnect. It fell back to AP before the router was up again. I have another AP with same SSID (mesh) but ems-esp does not connect to it, BSSID was not set.
Made a few tests with serial terminal to view what happen: 
- With more retries ems-esp connects to same AP if it is online again
- Calling `wifi reconnect `  while one AP is offline connects to the second AP.
- Just waiting it never connects to second AP.

IMHO the fallback should not happen when  STA was connected once. Ten it's clear this SSID/passoword works and the AP is down. Only on restart and STA is not reachable the fallback should happen. When connection is lost the stored STA should be deleted to connect to a second AP with same SSID. Same logic as on `wifi reconnect`, e.g. call begin().

I'll make some changes, test and make a PR to discuss the different logic.
BTW: setting ALL_CHANNEL_SCAN seems to work with core3 now and picks the strongest AP in a mesh.

MichaelDvP:
This is working for me. Fallback to AP is only if after start no connection is possible. Once connected it does not switch to AP and retries until ETH/WiFi is up again, or the user make a restart.


proddy:
This is much cleaner, and is also how other manufacturers handle network disconnects, and is also in line with the recent change where we removed AP_ON_DISCONNECT. thanks! I'll do some quick testing my end....

MichaelDvP:
I have removed the limitation for 4M systems not connecting to eth. 
BBQKees gateway E32 has 4M, no PSRAM and ethernet and should work with core3.

proddy:
@MichaelDvP should we change this:

image

perhaps rename to "When network connection not found"

PR #3098, opened 2026-05-25
Core3 updates

proddy:
- Add TLS support for all boards, including 4MB. HTTP and HTTPS are supported everywhere. Removed NO_TLS_SUPPORT
- Scheduler execute immediate calls an action
- Some cosmetic changes on Scheduler page
- exclude uptimes from HA Discovery, replace with timestamp
- replace HTTPClient with ESP_SSLClient
- fix HA Discovery bug
- Renamed uptime to bootTime


MichaelDvP:
I'm surprised that it fits on a 4M ESP32 chip, but flash is critical low:
`Flash: [==========]  99.3% (used 2017207 bytes from 2031616 bytes)`
and for C6 it's too large:
`Flash: [==========]  101.7% (used 2065410 bytes from 2031616 bytes)`

If we stay on this, we have to think about saving flash or create a loader again. (or drop 4M support).

empty scheduler name was intended for internal scheduls that hide from dashboard/MQTT. E.g. timer scheduls. For the immediate schedule it really makes no sense.

proddy:
We could add -DEMSESP_EN_ONLY for those boards with flash. It saves 160KB.
In my latest PR I've removed `HTTPclient` which saves some flash. s_4M is now 98.9% (8.4KB less)

The mandatory name in the Scheduler is so the execute command works. There is no unique ID in the SchedulerItem struct. It was easier to look up via the name than refactor the code.

MichaelDvP:
> We could add -DEMSESP_EN_ONLY for those boards with flash. It saves 160KB.

Yes, good idea.

PR #3101, opened 2026-05-28
extra handling for 0x470. #3096, fix scheduler onChange name check

MichaelDvP:


proddy:
great changes. Could you also apply this PR to the core3 branch so we stay in sync?

PR #3102, opened 2026-05-29
merge dev changes to core3

MichaelDvP:


proddy:
thanks for doing this. I noticed I introduced a big in the http code that causes a crash. When it's doing a GET to fetch a value (http or https). e.g. create an Scheduler item with type immediate and value `{"url":"https://emsesp.org/versions.json"}`. I'm back home on Monday so have limited time now to debug it.

MichaelDvP:
Looks like missing /r/n at end of header. But i can't test now.

proddy:
> Looks like missing /r/n at end of header. But i can't test now.

yes, that was it! thanks

PR #3109, opened 2026-06-06
Core3 updates #3

proddy:
- cancel firmware upload when doing an auto-update via HTTP/S
- Scheduler of type Immediate can be called directly from the Dashboard
- add missing volatile to values updated async/ISR
- Add new application setting to customise the name of EMS-ESP which is shown in the Web UI, also system info and MQTT info
- formatting & package update


src/web/WebSchedulerService.cpp
```diff
@@ -333,10 +333,12 @@ void WebSchedulerService::publish(const bool force) {
 }
 
 // count number of entries, default: only named items
-uint8_t WebSchedulerService::count_entities(bool cmd_only) {
+// if exclude_immediate is true, include those that are of type SCHEDULEFLAG_SCHEDULE_IMMEDIATE
```
MichaelDvP: This comment is misleading. 
Why do you want to exclude the immediate scheduls? Imo these type of scheduls are intended to make a complex command and can execute it by command or dashbord.

proddy: the comment is incorrect.

Immediate items are excluded in the count() because it's used in WebDataService::dashboard_data() for the Dashboard Data. We have a similar approach for analogsensors and temperaturesensors. But I agree is's confusing so will add back the Scheduler Immediate to the dashboard so they can be executed directly. These items should not be part of the scheduler since there technically isn't a 'schedule', but that's for another day.

MichaelDvP:
You don't need to call execute from web page, just set schedule to active (as any other schedule) and the scheduler executes it immediate (as before).

proddy:
> You don't need to call execute from web page, just set schedule to active (as any other schedule) and the scheduler executes it immediate (as before).

I prefer it this way. There is no active on/off for immediate scheduler items. I think this way is more intuitive. Have you tried it?

MichaelDvP:
> Have you tried it?

A few minutes ago. Dashboard is ok, but call it via terminal loops forever.

proddy:
> > Have you tried it?
> 
> A few minutes ago. Dashboard is ok, but call it via terminal loops forever.

oh, not good, I see that too. I'll make some changes. Using the active bit flag on/off to execute a command is a bit awkward. I'll make a new command.

Issue #3107, opened 2026-06-03
In Cooling mode changes in Thermostat Parameters are not sent back correctly to the physical Thermostats (and the bosch app)


MarkusMas:
### DESCRIPTION

Hello, first off, I’m new to the ems-esp project and would like to thank everyone involved.

When my heatpump switches to cooling mode, none of the parameters of the room thermostats (from ems-esp UI/ or HA UI) are applied/ send back correctly to the heatpumps room thermostats.

On the other hand all changes made on the physical room thermostats and in the bosch app are correctly reflected in the ems-esp data.

### REQUESTED INFORMATION

<details>
  <summary>systembackup/ config</summary>
  
```json
{
   "system":{
      "version":"3.8.2",
      "uptime":"000+00:15:15.610",
      "uptimeSec":915,
      "resetReason":"Power on reset / APP CPU reset by PRO CPU",
      "platform":"ESP32",
      "cpuType":"ESP32-D0WD-V3",
      "arduino":"ESP32 Arduino v2.0.17",
      "sdk":"v4.4.7-dirty",
      "freeMem":171,
      "maxAlloc":107,
      "freeCaps":4160,
      "usedApp":2378,
      "freeApp":2294,
      "partition":"app0",
      "flash_chip_size":16384,
      "psram":true,
      "psramSize":4093,
      "freePsram":4022,
      "model":"BBQKees Electronics E32V2.2 rev.2.2/2026012",
      "txpause":false,
      "gpios_allowed":"2, 4, 5, 14, 32, 34, 36, 39",
      "gpios_in_use":"2, 4, 5, 14, 32, 34, 36, 39",
      "gpios_available":""
   },
   "network":{
      "network":"Ethernet",
      "hostname":"ems-esp",
      "TxPowerSetting":0,
      "staticIP":false,
      "lowBandwidth":false,
      "disableSleep":true,
      "enableMDNS":true,
      "enableCORS":false
   },
   "ntp":{
      "NTPstatus":"connected",
      "enabled":true,
      "server":"time.google.com",
      "tzLabel":"Europe/Berlin",
      "timestamp":1780499322,
      "NTPStatus":"connected"
   },
   "ap":{
      "provisionMode":"disconnected",
      "ssid":"ems-esp",
      "security":"wpa2",
      "channel":1,
      "ssidHidden":false,
      "maxClients":4,
      "localIP":"192.168.4.1",
      "gatewayIP":"192.168.4.1",
[...]

Issue #1781, opened 2024-05-30
change “hc1 selected room temperature” in cooling mode not possible


kwant64:
### PROBLEM DESCRIPTION

_A clear and concise description of what the problem is._

### REQUESTED INFORMATION

_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [ ] Searched the problem in [issues](https://github.com/emsesp/EMS-ESP32/issues)
- [ ] Searched the problem in [discussions](https://github.com/emsesp/EMS-ESP32/discussions)
- [ ] Searched the problem in the [docs](https://emsesp.github.io/docs/Troubleshooting/)
- [ ] Searched the problem in the [chat](https://discord.gg/3J3GgnzpyT)
- [ ] Provide the output of http://ems-esp.local/api/system :

```lua
  System information output here:

We make use of a bbq-kees EPS gateway.
We use it for a NEFIT BOSCH Heatpump.

Wen we select Cooling it is not possible to change “hc1 selected room temperature”
When we change this value, an other field is changing. ( “hc1 manual temperature”).
On the device we kan change that value but not true the gateway.
Is this a bug or is ment to be in this way.
When we change to heating we kan change the value off the manual temperature and de selected room temperature is changing also…
We like to have the posseblillty to change de selected room temperature also in cooling mode.

Is there a possibility to fixed this?
```

### TO REPRODUCE

_Steps to reproduce the behavior:_

JUst place the heatpunp in cooling mode en try to change  “hc1 selected room temperature”

### EXPECTED BEHAVIOUR

_A clear and concise description of what you expected to happen._

We expected a change of this value

### SCREENSHOTS

_If applicable, add screenshots to help explain your problem._

### ADDITIONAL CONTEXT

_Add any other context about the problem here._

**(Please, remember to close the issue when the problem has been addressed)**

MichaelDvP:
Please log all while changing the value on the device, see https://emsesp.github.io/docs/Troubleshooting/#im-missing-certain-data-from-an-ems-device

kwant64:
[log(3).txt](https://github.com/emsesp/EMS-ESP32/files/15497735/log.3.txt)


Here is the log file.

MichaelDvP:
This is a very short log and it does not contain any change. It also looks like from an older ems-esp version. 
Send the system info and a log of > 2 min for the change of the value (working change on the device in cooling mode)

MichaelDvP:
[...]

PR #3111, opened 2026-06-07
reset system status

MichaelDvP:
this resets the status after 5 minutes, or via status page.
Also include the old fix for #1781

I've reverted all tests for #3107, seems that i'll never get a feedback to these tests and will never get a log of changing the modes. So the cooling modes stay unclear and better to leave it on the working heating modes.


proddy:
thanks. We should sync with core3 and then make core3 the next dev. After my recent PRs are checked are you ok if I do the merge? Otherwise, we'll never get onto the new Espressif core v3

PR #3113, opened 2026-06-07
merge dev changes into Core3

MichaelDvP:


proddy:
Thanks for fixing the console loop. Now I remember how it was implemented. Two questions

- does Scheduler Immediate need to be in the loop as it it's called on demand now?
- I'm thinking of introducing Commands as a new Module. Every Command has a Command, a Value and a unique Name, Then the Scheduler becomes a lot simplier. You just schedule a command (day schedule, timer, on-condition, on-change). This way commands can be re-used. And then there is no need for Immediate anymore. What do you think?

proddy:
> Thanks for fixing the console loop. Now I remember how it was implemented. Two questions
> 
> * does Scheduler Immediate need to be in the loop as it it's called on demand now?
> * I'm thinking of introducing Commands as a new Module. Every Command has a Command, a Value and a unique Name, Then the Scheduler becomes a lot simplier. You just schedule a command (day schedule, timer, on-condition, on-change). This way commands can be re-used. And then there is no need for Immediate anymore. What do you think?

** edit ** I made a start seperating out commands from the scheduler in a new fork, I'll see if I can I get it finished this week.

MichaelDvP:
> does Scheduler Immediate need to be in the loop as it it's called on demand now?

the direct call is very bad. It's called from web in the context of tcpip task and can blow this stack. Better to set active true and execute it in the loop. This is context scheduler task and stack is large enough.

proddy:
> > does Scheduler Immediate need to be in the loop as it it's called on demand now?
> 
> the direct call is very bad. It's called from web in the context of tcpip task and can blow this stack. Better to set active true and execute it in the loop. This is context scheduler task and stack is large enough.

as yes, now I understand. I'll implement the loop and use the active true

PR #3114, opened 2026-06-08
Commands

proddy:
Implemented Commands

- Scheduler picks a Command
- Scheduler Immediate is gone, it's just a command now
- Commands can be executed. It doesn't run in the emsesp loop, so I'm not sure of the stack usage. I haven't monitored memory
- Commands can be run via Console with `call commands <name> [<value>]`
- You can use Commands in Commands, e.g. a Command "command" can be "command/<name>"
- Upgrade to <3.9.0-dev-11 is handled automatically. Settings files are converted
- Made some changes to the WebUI to make it more user-friendly when editing commands and scheduler items
- Tested on S3 and 4MB non-PSRAM chips, and S3 and E32's

Also
- Fixed bug when comparing dev versions


MichaelDvP:
A lot of work, just to replace the schedule_immediate, that nobody needs (you can do this better in the home-automation).
The structure is clear, but you have to do one step (and a lot of clicks) more to add a schedule. Don't know what is better.
You have skipped the shunting yard processing, all my commands does not work. ( in line 146 of webCommandService use ` input["data"] = compute(data);`)
If you want to stay with this approach, i think it's better to do the scheduler in the main loop and run the command class async in it's own task. Prevents stack overflow for tcpip stack and long blocking for http/https processing.

src/core/network.cpp
```diff
@@ -82,7 +82,7 @@ void Network::begin() {
     WiFi.persistent(false);
     WiFi.setAutoReconnect(false);
     WiFi.mode(WIFI_STA);
-    WiFi.disconnect(true, true); // wipe old settings in NVS
+    WiFi.disconnect(true); // turn STA off; don't eraseap here (STA netif not started yet -> log_e, and persistent(false) means nothing is in NVS anyway)
```
MichaelDvP: we need to wipe the STA list. begin is called also on reconnect and if the old AP is off, it never reconnects to a different station with same ssid.

proddy:
> A lot of work, just to replace the schedule_immediate, that nobody needs (you can do this better in the home-automation).

You're such a delight to work with!

proddy:
> You have skipped the shunting yard processing, all my commands does not work. ( in line 146 of webCommandService use  input["data"] = compute(data);)

Could you give some examples? I can't reproduce, but may be missing the point. In the Command the Value is parsed via the shuntingYard

MichaelDvP:
> Could you give some examples? 

grafik


> I can't reproduce, but may be missing the point. In the Command the Value is parsed via the shuntingYard

I can't find where the shunting yard is called and get a `heatcnt + 1 not found in custom` error, after setting the `input["data"] = compute(data);` it works for me.

proddy:
thanks, I'll look into that. The test I used was:

image

proddy:
I'm moving the Scheduler to sync in the emsesp.loop() and making the Command async with xTaskCreate. When that works I'll fix the compute() issue

PR #3129, opened 2026-06-21
Core3 WebUI optimizations

proddy:
some optimizations to reduce bundle size by 13kb
- make compatible with react-router v8
- own version of toast for messages
- own version of async-validator
- move all router routes to one file, wrapped with isAdmin
- optionally build with specific languages instead of all 13


proddy:
@MichaelDvP lots of changes I don't expect anyone can review all of them. I'll merge this and if there are mistakes they will show. Things to pay attention to is
- is the toast messaging, whether it looks ok
- the field validation is still working

PR #3134, opened 2026-06-25
fix: negative dew point calculation (emulated RC100H)

s-schnitzer:
### Summary
If the dew point of an emulated RC100H would be negative, an integer overflow results in a wrong value reported to the heat pump.

### Steps to reproduce:
1. Go to "Devices" and select the BC400-HP 
2. Send values for `hc1 room temperature from remote` and `hc1 room humidity from remote` 
3. Go to "Devices" and select RC100H. Check whether "hc1 dew point temperature" is correct.

Example with correct dew point (greater than 0 °C): 
dew_point_ok

Example with wrong dew point (slightly below 0 °C):
dew_point_faulty

### Impact
If the dew point goes below 0 °C, a high dew point is sent to the HP, which (typically) will let the HP detect a temperature below the dew point and immediately stops cooling.

### DIFF
--- a/src/core/roomcontrol.cpp
+++ b/src/core/roomcontrol.cpp
@@ -377,19 +377,23 @@ void Roomctrl::temperature(uint8_t addr, uint8_t dst, uint8_t hc) {
 
 // send telegram 0x047B only for RC100H
 void Roomctrl::humidity(uint8_t addr, uint8_t dst, uint8_t hc) {
+    int16_t dew  = calc_dew(remotetemp_[hc], remotehum_[hc]);
+    int8_t  dew8 = EMS_VALUE_INT8_NOTSET;
+    if (dew != EMS_VALUE_INT16_NOTSET) {
+        dew8 = static_cast<int8_t>((dew >= 0 ? dew + 5 : dew - 5) / 10);
+    }
     uint8_t data[11];
-    data[0]      = addr | EMSbus::ems_mask();
-    data[1]      = dst & 0x7F;
-    uint16_t dew = calc_dew(remotetemp_[hc], remotehum_[hc]);
-    data[2]      = 0xFF;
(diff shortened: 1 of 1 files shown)


proddy:
looks like a good fix. @MichaelDvP ok?

MichaelDvP:
Ok.

proddy:
thanks @s-schnitzer for your contribution!

MichaelDvP:
But 20% rel. Humidity seems very dry. Maybe absolut humidity 20g/qm is more realistic today in Europe. This equals around 70% rel. Humidity at 25°C.

proddy:
@MichaelDvP @s-schnitzer do we need to fix the relative humidity? I do agree anything less than 40% is unusual

MichaelDvP:
@proddy Nothing to fix for humidity, this was a value @s-schnitzer send to the remote emulation, measured from somewhere else. The fix was the dewpoint was reported wrong for such values.
20% rel hum was reached here yesterday outdoors at 39°C, or it could be reached indoor in winter with very low outdoor temperatures and high heating demand. But not indoor with cooling demand.

s-schnitzer:
@MichaelDvP I agree that my values were unrealistic for Europe. I found this bug while playing around with these numbers to better understand how my heat pump operates. I want to use a manually calculated dew point and activate cooling if the outside temperature exceeds a threshold. Since my heat pump supports neither of these, I tried to provide the outside temperature as RC100H room temperature and "1" as room humidity. My manually calculated dew point can be passed as lower temperature. I fully understand that this not a common use-case at all. But since the existing calculation was obviously buggy, I considered it best to just fix it :)

MichaelDvP:
@s-schnitzer Thanks for explaining this. And thanks for doing these tests pointing you to the calculation bug.

Issue #3135, opened 2026-06-26
EasyControl - incorrect dew point


HansRemmerswaal:
The hc1 dew point temperature as shown in the EMS-ESP for the EasyControl can't be correct.
With a room temperature of 22.5 and a relative air humidity of 75% the dew point is about 17.9 and not 13 as shown in the screenshot.

I used this [link ](https://www.aquatec-vochtbestrijding.be/dauwpunt-berekenen/) to calculate the dew point. And that value is similar to what I see on weather forecast sites for my area.

On the EasyControl I can't see the dew point and on the menu of the heat pump I also can't see the dew point.

<img width="972" height="413" alt="Image" src="https://github.com/user-attachments/assets/f5f5e31d-7823-4572-9bb1-b2357a5effeb" />

<img width="540" height="308" alt="Image" src="https://github.com/user-attachments/assets/a401e698-a169-494d-9f61-b9c2896443e4" />

[emsesp_system_info.json](https://github.com/user-attachments/files/29375597/emsesp_system_info.json)

proddy:
Perhaps related to https://github.com/emsesp/EMS-ESP32/issues/1491#issuecomment-1849933866

Could you provide the telegram when reading 0x047B?

MichaelDvP:
Please check telegram 0x47B. Seems the value in offset 0 is something other than dewtemperature. 
In #1491it was strange that there is a dewtemperature of 3 degrees without humidity sent. 
Do you have any idea what this value represents?

For this thermostat we can remove  dewtemperature or calculate it by ems-esp if it is not send at offset 2 of 0x47B. 
What do you think?

Sbried:
0x047B Offset 0 is the low resolution due point of the remote controller in °C. Range 0-254, invalid = 255
Offset 1 is the humidity sensor value of the remote controller in %. Range 0-254, invalid = 255
Offset 2-3 is the high resolution dew point temperature in 0,1 °C. Range 0-65534, invalid = 65535.

0x047C ... 0x0482 is the same for the heating circuits 2...8

HansRemmerswaal:
ems-esp:$ watch on 47B
Watching incoming telegrams, displayed in decoded format
Filtering only telegrams that match a telegram type of 0x47B
2026-06-27T13:07:05.916 N 1: [emsesp] thermostat(0x38) W thermostat(0x10), RemoteHumidity(0x047B), data: 0E 4B
[...]

PR #3136, opened 2026-06-27
fixes for #3127 and #3135

MichaelDvP:


proddy:
thanks for getting these fixes in so quick. I'll cherry pick the ones for core3

PR #3158, opened 2026-07-14
updates from dev

MichaelDvP:
adds a lot of commits because i've always merged dev branch. As you ported manual there is mostly no change,only my changes come through, and that is not so much.
Includes changes for #3107, but without feedback i don't know if it works.
Also changes for #3148, i think the switch for syncing thermostat time gives the user more control what is written, the extra timezone option is only for one user, but does not hurt and normal users will not activate by accident.
Some missing translations, etc.


src/web/WebCommandService.cpp
```diff
@@ -374,13 +374,12 @@ std::string WebCommandService::get_metrics_prometheus() {
 void WebCommandService::get_value_json(JsonObject output, const CommandItem & ci) {
     output["name"] = (const char *)ci.name;
     // output["fullname"]  = (const char *)ci.name;
-    // output["type"]     = "command";
+    output["type"]     = "command";
```
proddy: why are these added - was it for the prometheus metrics? I commented them out before as it looked strange to show these unused parameters

MichaelDvP: I have not thought about. attributes were not working, i've enabled them and uncomment the attributes. I think the right type should be "string" to be in line with other values.

MichaelDvP: I've changed to string and added the other attributes. readable/writable are used from ioBroker to create the entities, don't know why they are commented out.

proddy: > I've changed to string and added the other attributes. readable/writable are used from ioBroker to create the entities, don't know why they are commented out.

it;'s because if you query other entities, like commands, you get data that is pointless. For example with commands. I have a command:

image

and 

`http://ems-esp.local/api/commands/do_nothing` shows me:

```json
{
  "name": "do_nothing",
  "fullname": "do_nothing",
  "type": "string",
  "command": "bad",
  "value": "0",
  "readable": true,
  "writeable": true,
  "visible": true
}
```

value, readable, writeable, visibile mean nothing

MichaelDvP: Yes, the complete commands class does not fit to other entities. 
api/commands/values shows the commands,, the attribute values shows the data for the command. 

Normal entities can be readable, eg. boiler/flowtemp, they do not have a command, readable/writable are boiler/selflowtemp, we can read the value and set it. Commands like boiler/reset does no have a value, they are writeable but not readable.

Here we have commands but strangely they give values back, and different vales depending an the way you call it. It's not logical. (Also commands that call commands are not very logical).

Here a nice crash:
grafik
With a url in data field it is queued, does not crash, but runs forever.

proddy: I can't check at the moment but guessing if you do the same call on a scheduler entry it will also show you value, readable, writeable, visibile.

Any ideas on how to clean this all up so it's consistent?

MichaelDvP: Scheduler read/write the value "ON"/"OFF". So it's a normal entities readable and writable and we can get a value back.

The commands show in web with the arrow and no value, execute only. If we adapt this in api its writable but not readable and api/commands/value gives enpty strings back. api/commands/values should only list names and emptyvalue.

But you can call a command with value and overwrite the default value, so we can alternative set it readable/writable, list the default values in api/commands/values and also give a attribute value back for api/commands/name/value. I think this is the better option, in the edit dialog it's also called value.

proddy: ok! that sounds like a good way forward

PR #3162, opened 2026-07-17
api attributs

MichaelDvP:
Now all readable entities have a `value` field and `api/device/values`  shows this field, also `api/device/shortname/value` gives this value back.
For api/commands/info` it shows name: a string that contains `"command:value"`
Scheduler in core2 have command/value fiellds and the setable was called `active`. Now this is in line and `value` is the on/off stable/readable state.

I think now fits all in one scheme.

One thing left: Because the commands execute on call the command.cpp checks the attribute first and execute if not found. 
so for an unknow attribute it give authentification failed instead of attribute not found back. Not critical but confusion. Maybe we need an extra check for the attribute separator.

### DIFF
--- a/src/web/WebCommandService.cpp
+++ b/src/web/WebCommandService.cpp
@@ -318,10 +318,19 @@ bool WebCommandService::get_value_info(JsonObject output, const char * cmd) {
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


proddy:
I'm back from vacation now, so will try this out

proddy:
works well. thanks.

PR #3185, opened 2026-08-06
minor fixes

proddy:
- DNS shows ipv4 by scanning all slots (not just 0 which will show the ipv6)
- fix LED flashing on boot
- added MAC Address fix to use same address as 3.8.x (feature is disabled for now)

### DIFF
--- a/src/core/led.cpp
+++ b/src/core/led.cpp
@@ -16,317 +16,317 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
- #include "led.h"
- #include "emsesp.h"
- 
- namespace emsesp {
- 
- uuid::log::Logger LED::logger_{F_(led), uuid::log::Facility::KERN};
- 
- // initialise the LED, fetching the settings from the WebSettingsService
- // set the LED to on or off when in normal operating mode
- void LED::init() {
-     // copy the application settings
-     EMSESP::webSettingsService.read([&](WebSettings & settings) {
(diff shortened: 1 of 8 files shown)


MichaelDvP:
Why does led.cpp shows all lines as changed? Is there a formatting issue?

proddy:
> Why does led.cpp shows all lines as changed? Is there a formatting issue?

oh, strange. I'll revert and re-submit

src/core/led.cpp
```diff
@@ -59,11 +59,16 @@ bool LED::loop(uint8_t healthcheck, bool button_busy) {
...
+        // while the button is held show the LED (yellow on RGB LED, on/off on standard LED),
+        // and on release go back to the default state and re-check the health on the next cycle
         if (last_button_busy_ != button_busy) {
             last_button_busy_ = button_busy;
-            set_led(button_busy ? Color::OFF : Color::YELLOW); // Yellow
+            if (button_busy) {
+                set_led(Color::YELLOW); // Yellow
+            } else {
+                reset_led();
+                previous_healthcheck_ = System::HEALTHCHECK_RESET;
```
MichaelDvP, line 70: If you reset `previous_healthcheck_` in the 'reset_led()` function, it ensures that health is blinking correct on next check and also if hide_led is changed.

MichaelDvP (approved):

src/core/led.cpp
proddy: I've fixed in the next PR

Issue #3089, opened 2026-05-20
Remove uptime and uptime (sec) as MQTT Discovery topics


proddy:
uptime and uptime (sec) have their own MQTT discovery topics, meaning a state change (and if persisted a DB update) every 60 seconds. It's better to remove these discovery topics, but still leave them in the heartbeat MQTT topic, and instead use the `bootTime` from the `info` topic and have HA use this along with `"device_class": "timestamp"` to figure out the uptime itself.

In the meantime, to remove the DB spamming in HA you can use

```yaml
recorder:
  exclude:
    entities:
      - sensor.system_uptime
```

this was originally reported by user `Amejonah` in Discord: https://discord.com/channels/816637840644505620/816638587508490250/1505605727429853395

proddy:
fixed in core3, which will be the next dev

PR #3186, opened 2026-08-07
more minor fixes, Eth fixes and other improvements

proddy:
- cleaned up LED as suggested by Michael
- fix edge when hostname/mDNS revrets to tasmota by bouncing the DHCP client when the name was actually wrong and DHCP had already started
- fix Uptime/Bootime warning in HA
- HA entities are in lower-case
- Changes to Dialog boxes so it's consistent between the command types. Removed > from the Command table
- Replaced Execute with Run in webUI
- Fix crash when changing AP provision mode
- Fix crash on Eth when there is no network and it continues to reconnect and load the PHY driver
- Fix going to AP even when AP provison mode is set to 'Never'


src/core/mqtt.cpp
```diff
@@ -1067,12 +1067,12 @@ bool Mqtt::publish_ha_sensor_config(uint8_t               type,        // EMSdev
...
 
         // Override - there are exceptions for DeviceType::SYSTEM, which uses the heartbeat topic
-        // and when fetching the version and bootTime we want to take this from the info topic instead
+        // and when fetching the version and uptime we want to take this from the info topic instead
         if (device_type == EMSdevice::DeviceType::SYSTEM) {
             // handle the exceptions
             if (strncmp(entity, "version", 7) == 0) {
                 snprintf(stat_t, sizeof(stat_t), "~/%s", F_(info));
-            } else if (strncmp(entity, "bootTime", 8) == 0) {
+            } else if (strncmp(entity, "bootTime", 6) == 0) {
```
MichaelDvP: Why only compare 6 chars?

src/core/mqtt.cpp
```diff
@@ -587,9 +587,8 @@ void Mqtt::ha_status() {
 
     // These come from the info MQTT topic - and handled in the publish_ha_sensor_config function
     publish_system_ha_sensor_config(DeviceValueType::STRING, "Version", "version", DeviceValueUOM::NONE);
-    publish_system_ha_sensor_config(DeviceValueType::STRING, "Boot time", "bootTime", DeviceValueUOM::UPTIME);
+    publish_system_ha_sensor_config(DeviceValueType::STRING, "Uptime", "bootTime", DeviceValueUOM::UPTIME);
```
MichaelDvP: A bit confusing, is it uptime or boottime?

src/core/mqtt.cpp
proddy: it should be 8

src/core/mqtt.cpp
proddy: boot time (time & date) is calculated from NTP and added as a bootTime element to the MQTT info topic. The HA sensor is called `uptime` and it derives the elapsed time value from the bootTime. It is confusing. I changed it because in HA was showing the uptime and the label was bootime.

MichaelDvP: I'll never get the HA logic (if it can be called logic). We shortly replaced uptime by boottime, #3089, now we replace boottime by uptime and mix name and entity.

proddy: > I'll never get the HA logic (if it can be called logic). We shortly replaced uptime by boottime, #3089, now we replace boottime by uptime and mix name and entity.

not really. bootTime was added to the info topic as a timestamp and sent only once on EMS-ESPboot. `uptime` is the duration (in a sort of ISO 8601 format) and `uptime_sec` are already in the heartbeat topic. HA can calculate the system uptime accurately from the bootTime when using the new UPTIME device class, replacing the need to write code in EMS-ESP to calculate it like we had before.

Issue #3184, opened 2026-08-05
EMS-ESP 3.8.3  crashed in less than 24h due to "no free mem"


ralfbergs:
### DESCRIPTION

Home Assistant reported that EMS-ESP had restarted. When I investigated it, I found out that it was due to "Free memory" going to less than 84 K in less than 24 hours. I had to downgrade from 3.8.3 to 3.8.2 again therefore.

### REQUESTED INFORMATION

_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [X] Searched the issue in [issues](https://github.com/emsesp/EMS-ESP32/issues)
- [X] Searched the issue in [discussions](https://github.com/emsesp/EMS-ESP32/discussions)
- [X] Searched the issue in the [docs](https://emsesp.org/Troubleshooting/)
- [X] Searched the issue in the [chat](https://discord.gg/GP9DPSgeJq)
- [X] Provide the System information in the area below, taken from `http://<IP>/api/system`

```json
[...]

PR #3187, opened 2026-08-09
workaround for memory leak, #3184

MichaelDvP:
confirmed working.
What about 3.8.4 main? Leave it? make 3.8.5? I have a 3.8.4-patch.1 as source in my main branch, but no binaries.


proddy:
I would wait on patching 3.8.4. It's not something many users will encounter. We should re-sync the patch branch at some point (https://github.com/emsesp/EMS-ESP32/tree/patch) and add the fix there. And when we have more minor fixes on 3.8.4, we can do a 3.8.5.

I'll be out the next 2 weeks so can't make the change myself

PR #3190, opened 2026-08-16
Main

fengshutao:
Add XIAO ESP32-C3 device


proddy:
@fengshutao do you want to add this to the dev branch?

Issue #3213, opened 2026-09-01
Reduce EMS-ESP energy consumption


proddy:
This issue is to look at overall CPU and power consumption on the ESP32, to reduce heat and power draw. The EMS-ESP runs on 2 cores, one of which is at 100% with no idle time.

One user mentioned a 2-degree increase between 3.8 and 3.9.x. And there is another discussion on https://github.com/emsesp/EMS-ESP32/discussions/3208

The BBQKees E32-V22 board has an onboard dallas sensor (`gateway_temperature`) which is sent in the temperaturesensor_data MQTT payload. The S2/S3/C3 uses the internal ESP32 sensor and this is published in the MQTT heartbeat as just `temperature`.

Options to look into are:

- switching off WiFi radio if ETH is used
- using 10BASE-T half duplex with autonegotiation off on ETH boards
- adding a yield to the main Core0 loop so it's idle task can run
- looking into other examples, like https://tasmota.github.io/docs/Energy-Saving/

Whopper500:
From my experience with several different ESP32 applications and energy consumption measurements:
1. Switch from Wifi to Eth
2. Use sleep modes

I already designed (hardware) und programmed (firmware) several battery powered applications. CPU underclocking is a must-have, but the effect is limited. UART was suprisingly energy intensive (if I remember correctly: +8 mA just the initialization). Use of a high quality voltage regulator with 3,3V or even 3V input to ESP32 had a strong effect, but hardware changes are not part of the discussion.

From my experience, use of sleep modes - especially deep sleep - is a game changer.

You could easily check the Reset reason and if it is wake up from deep sleep, the ESP32 could ignore major parts of the code, since no reconfiguration or other setup overhead is necessary.

Important: Before deep sleep, MQTT needs to be terminated correctly. gpio settings can be fixed during sleep (see "gpio_hold").

Whopper500:
Addition: W5500 Ethernet Boards are significantly more efficient compared to e.g. LAN8720A.
W5500 operates outside the CPU, has its own sleep and power down controller and keeps connected even within Deep Sleep periods.

We speaking about up to 50mA difference between these architectures.

proddy:
I've implemented the changes I proposed in dev-7 and will monitor the temperature readings. The sleep I'll look into next,

Whopper500:
[...]

PR #3215, opened 2026-09-02
Network optimizations

proddy:
Some small changes in the network code following further code inspection, triggered by the CPU consumption discussion on https://github.com/emsesp/EMS-ESP32/issues/3213

- added yield to main loop, so core 1’s idle task is run
- power down WiFi when Ethernet is active, freeing up 20kb heap
- WiFi nosleep takes effect on next reconnect
- added eth_10mbit, which forces 10BASE-T half-duplex & autonegotiation off, halving the PHY’s power draw. It’s off by default as it will cause issues on managed switches with a port forced to 100Mb

### DIFF
--- a/CHANGELOG_LATEST.md
+++ b/CHANGELOG_LATEST.md
@@ -37,8 +37,5 @@ This release is based on the latest Espressif/Arduino core version 3. It brings 
 - changes to the Dialogs in Dashboard page, removed the run icon in the Commands page. Renamed Execute to Run.
 - set `None` for undefined values in HA `val_tpl`
 - changes in WebUI - remove multiple calls systemInfo endpoint, Vewrsion pages checks for internet connection [#3195](https://github.com/emsesp/EMS-ESP32/pull/3195)
-- added yield to main loop, so core 1’s idle task is run [#3213](https://github.com/emsesp/EMS-ESP32/pull/3213)
-- power down WiFi when Ethernet is active, releasing DMA buffer pool of around 20KB [#3213](https://github.com/emsesp/EMS-ESP32/pull/3213)
-- WiFi nosleep takes effect on next reconnect [#3213](https://github.com/emsesp/EMS-ESP32/pull/3213)
-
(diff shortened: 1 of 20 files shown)


MichaelDvP:
Is it intended to have the eth_10mbit option only for custom boards? 
I have expected the option in network settings in an ethernet section.

proddy:
> Is it intended to have the eth_10mbit option only for custom boards? I have expected the option in network settings in an ethernet section.

Yes, it's still quite experimental and needs more data to see if its actually worth adding (to reduce heat). I could have put it in the Network section and wrapped developer_mode around it. Shall I move it?

MichaelDvP:
Yes, for me it is equivalent to the wifi no_sleep option. Could save energy but also could make trouble.

proddy:
> Yes, for me it is equivalent to the wifi no_sleep option. Could save energy but also could make trouble.

okay, I'll move it to the Network Settings page. One reason was that under custom it would only flag if PHY was not None. ESP32-S3/C3 boards have no Ethernet PHY so I need to make some changes to fetch PHY or board data in the network pages. s

proddy:
eth_10mbit moved.

MichaelDvP (approved):
Looks good. ETH connection with a lot of more free mem.
No issues when using eth_10mbit connected to eth port of a wifi repeater

proddy:
> Looks good. ETH connection with a lot of more free mem. No issues when using eth_10mbit connected to eth port of a wifi repeater

Thanks Michael. I know some of my PRs are quite impactful and sometimes too rushed, so I do appreciate the time you take to review and test, and provide really good feedback without getting too annoyed!

proddy:
And for the record I did test:

- pulling the Ethernet cable and watching fallback to WiFi, and to AP when no SSID is configured
- toggle `phy_type` to none and checking WiFi still comes up normally
- enabling the 10Mbit option and  confirming the link comes up and the web UI is still kinda responsive
- `nosleep` it now takes effect without a reboot when toggled

Issue #3216, opened 2026-09-03
Improve MD5 and firmware uploads


proddy:
### Discussed in https://github.com/emsesp/EMS-ESP32/discussions/3203

<div type='discussions-op-text'>

<sup>Originally posted by **proddy** August 30, 2026</sup>
From ralfbergs on Discord:

Hi, I would like to mention that the handling of firmware uploads that include the MD5 digest is not really very intuitive. You don't even know whether it's considering the digest.
If it's not too much effort, I'd like to suggest trying to improve the UX in this scenario, e.g. as follows:
After uploading the MD5, you could mention something like "Ok, digest received, now upload BIN". And then, when the BIN has been received, and the MD5 matches, you could issue "Great, MD5 matches, proceeding to flash now..."
Just an idea... And only if the effort is very, very minimal...
(Currently it shows MD5: <md5>, but the MD5 shown is the MD5 hash computed over the uploaded MD5 file, not the hash included in the file... This seems pretty useless...)</div>

PR #3217, opened 2026-09-03
Cleaner MD5 and Firmware upload messages in WebUI

proddy:
Implemented #3216

- When uploading a MD5 the dialogue changes, showing validation and next steps
- Changes to standalone to simulate a slow file upload
- eFuse writing removed (this is now down in the new BBQKees installer)


MichaelDvP (approved):

MichaelDvP (commented):
Isn't @bbqkees using the `system/fuse/mfg`and mfgadd  commands?

proddy:
> Isn't @bbqkees using the `system/fuse/mfg`and mfgadd commands?

for 3.8.x yes, but efuse burning will be shortly replaced with a new flash tool I'm building and I'd rather not expose in the code where we keep our secrets

PR #3220, opened 2026-09-04
Add OLIMEXPOEW board profile; guard against PSRAM/Ethernet clock GPIO conflict

pepe82sh:
## Problem

On classic ESP32 (D0WD), external/off-package PSRAM (used by WROVER modules) is wired to GPIO16 (CS) and GPIO17 (CLK). These are the same two pins `ETH.begin()` can use to drive the RMII clock output (`ETH_CLOCK_GPIO16_OUT` / `ETH_CLOCK_GPIO17_OUT`). Calling `ETH.begin()` with one of those clock modes while PSRAM is enabled doesn't fail cleanly - it corrupts heap metadata and crash-loops the device (`Guru Meditation Error: StoreProhibited`) on every boot, before the console or network ever come up. Recovery requires a full flash erase.

This is exactly what happens with the existing `OLIMEXPOE` board profile (Olimex ESP32-POE, `eth_clock_mode=3`/`ETH_CLOCK_GPIO17_OUT`) on the WROVER/PSRAM variant of that board (e.g. ESP32-POE-ISO-WROVER). That clock mode is correct for the WROOM (no PSRAM) variant, but crashes the WROVER variant.

Confirmed directly with Olimex support ([forum thread](https://www.olimex.com/forum/index.php?topic=9374.0)):

> When WROVER module (instead of the default WROOM) is used then GPIO16 and GPIO17 are unavailable (both are used for the PSRAM inside the module).
>
> For boards using WROVER, the Ethernet clock moved from GPIO17 to GPIO0

Also matches known arduino-esp32 issues: espressif/arduino-esp32#3979, espressif/arduino-esp32#6417.

## Fix

1. **New `OLIMEXPOEW` board profile** - same GPIOs as `OLIMEXPOE` but `eth_clock_mode=1` (`ETH_CLOCK_GPIO0_OUT`), matching Olimex's own pin change for the WROVER variant.
2. **Guard in `Network::initialPhase()` / `Network::startEthernet()`** - detects `PSRAM enabled && eth_clock_mode in {GPIO16_OUT, GPIO17_OUT}` and skips Ethernet (falls back to WiFi/AP) instead of calling `ETH.begin()` into the crash. This isn't just for the new profile - anyone who picks the existing, ambiguous `OLIMEXPOE` name on WROVER hardware, or configures `CUSTOM` with that combination, now gets a clean fallback instead of a crash-loop that requires a full flash erase to recover from.

## Testing

Both changes verified on real Olimex ESP32-POE-ISO-WROVER hardware (ESP32-D0WD-V3) over serial console, against both `3.8.4` (release) and current `dev` (3.9.0-dev, Arduino Core 3.3.8 / IDF 5.5.4):

- `OLIMEXPOE` (the conflicting combination) previously crash-looped every boot (`Guru Meditation Error: StoreProhibited`, `EXCVADDR: 0x0000000b`, traced to `heap_tlsf.c` via `uart_driver_install` <- `System::uart_init`). With this fix it boots clean, logs the new error, and comes up in AP mode. PSRAM confirmed still available.
- `OLIMEXPOEW` boots clean with PSRAM available, and the Ethernet PHY link actually comes up - confirmed via a directly connected host's NIC reporting `100baseTX <full-duplex>` link state.

### DIFF
--- a/interface/src/app/main/types.ts
+++ b/interface/src/app/main/types.ts
@@ -307,6 +307,7 @@ export const BOARD_PROFILES = {
   LOLIN: 'Lolin D32',
   OLIMEX: 'Olimex ESP32-EVB',
   OLIMEXPOE: 'Olimex ESP32-POE',
+  OLIMEXPOEW: 'Olimex ESP32-POE-ISO WROVER (PSRAM)',
   C3MINI: 'Wemos C3 Mini',
   S2MINI: 'Wemos S2 Mini',
   S3MINI: 'Liligo S3'
--- a/src/core/console.cpp
+++ b/src/core/console.cpp
@@ -305,7 +305,7 @@ static void setup_commands(std::shared_ptr<Commands> const & commands) {
             std::string         board_profile = Helpers::toUpper(arguments.front());
             if (!EMSESP::system_.load_board_profile(data, board_profile)) {
                 shell.println(
-                    "Invalid board profile (S32, E32, E32V2, E32V2_2, MH-ET, NODEMCU, LOLIN, OLIMEX, OLIMEXPOE, C3MINI, S2MINI, S3MINI, S32S3, CUSTOM)");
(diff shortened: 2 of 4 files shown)


proddy:
looks ok, @MichaelDvP ?

MichaelDvP (approved):

Issue #3201, opened 2026-08-30
EMS-ESP 3.9.0-dev.5 - dev.8 restarts unexpectedly.


VlastiBroucek:
### DESCRIPTION
After running for several days and being restarted by scheduler earlier today, there have been too unexpected restarts in the afternoon.

<img width="475" height="318" alt="Image" src="https://github.com/user-attachments/assets/3c2540cb-86da-4571-9c66-6c0e43f00b99" />

I have not noticed them immediately, so did not have time to set up console logging or anything else.

This is, however, what is in log after the second restart:

```
2026-08-30T16:54:10.000 I 1: [emsesp] EMS-ESP version 3.9.0-dev.5
2026-08-30T16:54:10.000 I 2: [emsesp] Last system reset reason Core0: Timer group1 watch dog reset, Core1: APP CPU reset by PRO CPU
2026-08-30T16:54:10.000 I 3: [emsesp] Loaded board profile E32V2 (PSRAM: 4096 KB)
2026-08-30T16:54:10.000 I 4: [emsesp] Starting Commands service
2026-08-30T16:54:10.000 I 5: [emsesp] Starting Custom Entity service
2026-08-30T16:54:10.000 I 6: [emsesp] Starting Scheduler service
2026-08-30T16:54:10.000 I 7: [emsesp] Library loaded: 129 EMS devices, 569 device entities, 11 languages (en,de,nl,sv,pl,no,fr,tr,it,sk,cs)
2026-08-30T16:54:10.000 I 8: [temperaturesensor] Starting Temperature Sensor service
2026-08-30T16:54:10.000 I 9: [emsesp] Starting Web Server
2026-08-30T16:54:10.000 I 10: [emsesp] Starting Modules service
2026-08-30T16:54:14.057 I 11: [emsesp] Starting MQTT service
2026-08-30T16:54:19.057 I 12: [network] Network connected via Ethernet (IP: 192.168.50.132)
2026-08-30T16:54:19.057 I 13: [emsesp] Starting mDNS Responder service for ems-esp.local
2026-08-30T16:54:19.057 I 14: [emsesp] Starting NTP service
2026-08-30T16:54:20.486 I 15: [mqtt] MQTT connected
2026-08-30T16:54:22.558 I 16: [system] NTP connected
2026-08-30T16:54:27.342 I 17: [emsesp] Detected EMS device: boiler (0x08)
[...]

PR #3240, opened 2026-09-14
more memory debug, optimized http, better Eth error handling

proddy:
A number of changes

- added `ethernet_status()` and moved up the gpio16/17 check so it's called once in the begin() 
- Possible fix for #3201 - a race condition between reading strings for the version data outside AsyncTCP
- Optimisations to HTTPS handling. -D ENABLE_PSRAM moves the ~17 KB of per-request TLS buffers to PSRAM on boards that have it
- Use Node26, removing need for corepack
- `show system` shows more memory stats plus whether Ethernet is active or not. This is needed for the new flash installer to detect a healthy gateway board.
- ensure internal sensors are created on a fresh install (not factory). like gateway_temperature, core_voltage, supply_voltage, led
- bumped ESP32Async/ESPAsyncWebServer @ 3.12.1
- added option to auto-check for firmware versions

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -29,7 +29,7 @@ BreakBeforeBinaryOperators: NonAssignment
 BreakConstructorInitializersBeforeComma: true # Always break constructor initializers before commas and align the commas with the colon.
 ExperimentalAutoDetectBinPacking: false
 KeepEmptyLinesAtTheStartOfBlocks: false
-MaxEmptyLinesToKeep: 4
+MaxEmptyLinesToKeep: 1
 PenaltyBreakBeforeFirstCallParameter: 200
 PenaltyExcessCharacter: 10
 PointerAlignment: Middle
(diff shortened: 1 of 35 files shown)


MichaelDvP:
I have not checked all code ( a lot of changes), but test gives some strange result:
With update disabled. the settings->firmware shows ` Checking online for available versions…`,what is obvious wrong, since it is disabled. But the stored versions are also not shown, why?.
When changing anything in application settings the serial terminal freeze. Web still works. Seems main task is blocked.
**Edit: Found: i was using custom board profile and `reset_unused_gpios()` sets the serial tx to input.This is not caused by this PR changes.**

proddy:
> With update disabled. the settings->firmware shows  Checking online for available versions…,what is obvious wrong, since it is disabled. But the stored versions are also not shown, why?.

I think I fixed this and also made some modifications.

benbrummer:
Maybe this fixes my issue, which I faced with the last two dev updates, wäre my device was not accessible. Removing power brought it back.

proddy:
this PR has been open  for a few days now so I'll merge and fix any issues going forward

PR #3242, opened 2026-09-16
fix: add 12 kW and 15 kW steps to enum_maxHeat

Jaano:
## Problem

The heat limit registers on heat pumps using `EMS_DEVICE_FLAG_HEATPUMP` (product
172/173) are bytes 2/3/4 of telegram `0x0492`, read as a direct index into
`enum_maxHeat`. That table only had six entries, so any unit whose auxiliary
electric heater goes above 9 kW reports an index the table can't resolve.

Nothing bounds-checks the raw byte on the way in, and every render path guards
on `value < options_size`, so an out-of-range index is dropped silently: no
value is published to MQTT, the API omits the field, Home Assistant shows
`unknown`, and nothing appears in the log. `Helpers::value2enum` stops at the
same boundary, so the value can't be written back from EMS-ESP either, leaving
the entity permanently stuck.

## Root cause / verification

Measured on a Bosch Compress 7000i AW (CS7000iAW 13 OR-T, product 172) with a
15 kW auxiliary heater, by setting each limit from the unit's own display and
reading telegram `0x0492` back:

```
display 12 / 9 / 6 kW  ->  0x0492 data: 03 00 06 05 04

raw  kW   note
  0   0   agrees with the existing table
  1   2   not offered by this heater
  2   3   not offered by this heater
  3   4   not offered by this heater
  4   6   agrees with the existing table
  5   9   agrees with the existing table
  6  12   new
  7  15   new
```

Indices 0, 4 and 5 match what EMS-ESP already decodes, confirming the existing
entries and their ordering — the table was correct, just short. The 2 kW and
4 kW entries belong to the 4-stufig (2/4/6/9 kW) heater variant described in
the HPC 410 service manual section 4.2.2, so a 3 kW-stepped heater simply skips
them and walks 0, 3, 6, 9, 12, 15 kW, i.e. raw 0, 2, 4, 5, 6, 7. The same
sparse pattern is already encoded for CS6800 devices in
`Boiler::process_HpHeaterConfig`, which maps raw `{0, 2, 4, 5}` onto
`{0, 3, 6, 9}` kW.

## Fix

Add the missing `12kW`/`15kW` locale strings and extend `enum_maxHeat` with
the two new steps in `src/core/locale_common.h`. `CHANGELOG_LATEST.md` updated.

Only devices carrying `EMS_DEVICE_FLAG_HEATPUMP` are affected; CS6800 devices
use `enum_maxHeat1`/`enum_maxHeat2` and are unchanged.

## Testing

Standalone build: `test entity_dump` output differs from the committed
`docs/dump_entities.csv` in exactly the six `maxheat` rows for products 172
and 173, and nowhere else.

Flashed to the hardware the bug was found on (Bosch Compress 7000i AW,
CS7000iAW 13 OR-T, product 172, 15 kW auxiliary heater) and confirmed live via
the device API. All three previously-`unknown` entities now resolve correctly

## Checklist (CONTRIBUTING.md)

- [x] PR is against `dev`
- [x] Single fix, no unrelated files touched
- [x] Formatted per `.clang-format`
- [x] `CHANGELOG_LATEST.md` updated
- [x] I agree to the CLA in `CONTRIBUTING.md`


proddy:
nice work, thanks!

PR #3248, opened 2026-09-20
Littlefs migrate

proddy:
Auto migrate older LittleFS file system used in Arduino Core 2 (v3.8.x) to the Core 3 used in 3.9.0


proddy:
@MichaelDvP I found a way to check for older 3.8.x based LittleFS formats and migrate to the 3.9.x format. I've tested in E32s board in various formats and it seems to work. Can you take a look and do some tests if you have time?

This will remove the need to export settings->factory reset->import settings when upgrading from stable 3.8.4 to 3.9.0

I will need to make a new 3.8.5 version with the updated messages afterwards

proddy:
oops, I realise this PR was automatically merged with the previous one. Apologies for that

